PR #1, opened 2022-09-06
Fix linting, move tsconfig.json to root

kenoir:
## What does this change?

This branch fixes linting and moves to use the root tsconfig.json for the cdk package.


jfsoul (approved):
🙏

PR #2, opened 2022-09-06
Add skeleton CDK infra, build & deploy

kenoir:
## What does this change?

This change adds a:
- `scripts/ci.sh` to build and package assets
- `.github/workflows/ci.yml` to configure GHA & run the build script
- `riff-raff.yml` to work with the GHA riff-raff action to enable deployment of the CF stack & lambda

It is intended to provide a skeleton deployment mechanism that can be extended for other lambdas in this repo.

**Note:** The `repo-fetcher` lambda will not yet function as it lacks the necessary configuration (it will be deployable & runnable however).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,10 +39,10 @@ jobs:
         run: ./script/ci.sh
 
       - name: Upload to riff-raff
-        uses: guardian/actions-riff-raff@v1
+        uses: guardian/actions-riff-raff@v1.1.3
         with:
           app: github-lens
-          buildNumber: 'dev'
+          buildNumber: ${{ env.GITHUB_RUN_NUMBER }}
           projectName: playground::github-lens
           configPath: riff-raff.yml
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         with:
           app: github-lens
           buildNumber: ${{ env.GITHUB_RUN_NUMBER }}
-          projectName: playground::github-lens
+          projectName: deploy::github-lens
           configPath: riff-raff.yml
     env:
       NODE_OPTIONS: '--max_old_space_size=4096'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+# Find full documentation here https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
+name: CI
+
+on:
+  pull_request:
+
+  # Manual invocation.
+  workflow_dispatch:
+
+  push:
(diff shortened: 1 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Run script/ci
+        run: ./scripts/ci.sh
+
+      - name: Upload to riff-raff
+        uses: guardian/actions-riff-raff@v1.1.3
+        with:
+          app: github-lens
+          buildNumber: ${{ env.GITHUB_RUN_NUMBER }}
+          projectName: playground::github-lens
```
jfsoul, line 46: nitpicking, but to avoid more than one package name floating around in Riffraff - can we just call it `security::github-lens` (or the deploy tools variant) from the start?

packages/cdk/bin/cdk.ts
```diff
@@ -4,4 +4,4 @@ import { GithubLens } from '../lib/github-lens';
 
 const app = new App();
 
-new GithubLens(app, 'GithubLens-CODE', { stack: 'deploy', stage: 'CODE' });
+new GithubLens(app, 'GithubLens-CODE', { stack: 'playground', stage: 'CODE' });
```
NovemberTang, line 7: why have we moved this to dev playground?

packages/common/config.ts
```diff
@@ -9,8 +9,14 @@ export type Config = {
 		appInstallationId: string;
 	};
 	dataBucketName: string | undefined;
+	dataKeyPrefix: string | undefined;
 };
 
+export const decrypt = <T>(configValue: T, secret: boolean): T => {
```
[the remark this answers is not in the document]
jfsoul: (say this partly because it's currently a dummy implementation too)

jfsoul (approved):
👍

packages/cdk/bin/cdk.ts
kenoir: It wasn't deployed anywhere up till now (though it should be eventually). It's in `playground` to avoid messing around in `deploy` until we've got the structure of the app sorted. I'm happy to shift it to over though 🤷

riff-raff.yml
```diff
@@ -0,0 +1,33 @@
+regions: [eu-west-1]
+stacks: [playground]
+deployments:
+  lambda-to-s3:
+    type: aws-s3
+    app: repo-fetcher
+    sources:
+      - packages/repo-fetcher/dist/repo-fetcher.zip
+    parameters:
+      bucket: developer-playground-dist
```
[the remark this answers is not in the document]
kenoir: The `aws-s3` deployment type doesn't support this I think? I should move this into the `aws-lambda` step.

PR #3, opened 2022-09-07
Add teams that own repos to Repo object

NovemberTang:
## What does this change?

Apart from linting, this change introduces a a way to include repository team owners as part of the Repo object 

## How to test

`npm run test -w packages/repo-fetcher`

Verify that all tests pass

## Have we considered potential risks?

We have skirted round any potential API limits but requesting owners per team, rather than by repo. There are ~100 teams and ~3000 repos at time of writing, so even paginated, this is a 30x reduction in pages requested.

### DIFF
--- a/packages/common/github/github.ts
+++ b/packages/common/github/github.ts
@@ -96,9 +96,10 @@ export const listTeams = async (config: Config): Promise<TeamsResponse> => {
 
 export const getReposForTeam = async (config: Config, teamName: string): Promise<TeamRepoResponse> => {
     const octokit = await getOctokit(config);
-    const request =  await octokit.request("GET /orgs/guardian/teams/{team}/repos", {
-        team: teamName,
+    const request =  await octokit.paginate(octokit.teams.listReposInOrg, {
+        org: "guardian",
+        team_slug: teamName,
         per_page: defaultPageSize,
-      });
-    return request.data;
+      }, response => response.data);
+    return request;
 };
--- a/packages/common/github/github.ts
+++ b/packages/common/github/github.ts
@@ -21,6 +21,12 @@ const octokit = new Octokit();
 export type RepositoriesResponse = GetResponseDataTypeFromEndpointMethod<
     typeof octokit.repos.listForOrg
     >;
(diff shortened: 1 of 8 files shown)


kenoir (approved):
👍

packages/common/github/github.ts
```diff
@@ -82,3 +88,32 @@ export const listRepositories = async (
 		(response) => response.data,
 	);
 };
+
+export const listTeams = async (config: Config): Promise<TeamsResponse> => {
+	const octokit = getOctokit(config);
+	return await octokit.paginate(
+		octokit.teams.list,
+		{
+			org: 'guardian',
```
jfsoul, line 97: maybe we should move `'guardian'` into a const somewhere now it's in a few methods here

packages/repo-fetcher/src/transformations.test.ts
```diff
@@ -0,0 +1,164 @@
...
+
+		const ownerArray = findOwnersOfRepo('someRepo', [owner1, owner2, owner3]);
+
+		expect(ownerArray).toStrictEqual(['team1', 'team2']);
+	});
+});
+
+describe('repository objects', function () {
+	it('should combine a RepositoryResponse with a list of owners', function () {
+		const repo = {
```
jfsoul, line 25: as this is a bit of an unwieldy object, we could store it in a separate file as a json test resource that we load into this test?

packages/repo-fetcher/src/transformations.ts
```diff
@@ -0,0 +1,97 @@
...
+	return repos
+		.filter((repo) => repo.role_name === 'admin')
+		.map((repo) => repo.name);
+};
+
+export const createOwnerObjects = async (
+	config: Config,
+	teamSlug: string,
+): Promise<RepoAndOwner[]> => {
+	const allRepos: TeamRepoResponse = await getReposForTeam(config, teamSlug);
```
jfsoul, line 85: For more separation of concerns, we could refactor this so that all the github requests are made higher up, and `transformations.ts` is just data manipulation (that's what it sounds like to me from the name, without reading the code)

PR #4, opened 2022-09-09
Add data about valid github teams.

NovemberTang:
## What does this change?

We are now able to use information about which teams are in P+E. This information will be used by both tracker and repocop.


packages/repo-fetcher/src/validGithubTeams.ts
```diff
@@ -0,0 +1,56 @@
+export const validGithubTeams = [
```
jfsoul, line 1: might be worth calling these _valid Github admin/owner teams_ and adding a comment explaining that these are the subset of teams that map to an actual product team (or another arrangement i.e. https://github.com/guardian/recommendations/blob/main/ownership.md#orphaned-projects), i.e. we are confident have the ability to maintain a repo.

packages/repo-fetcher/src/handler.ts
```diff
@@ -10,19 +10,19 @@ import {
...
 	getAdminReposFromResponse,
 	RepoAndOwner,
 	transformRepo,
 } from '../src/transformations';
+import { validGithubTeams } from './validGithubTeams';
 
-const save = (repos: Repository[]): Promise<void> => {
-	const prefix = config.dataKeyPrefix;
-	const key = `${prefix}/github/repos.json`;
+const save = (JsonString: string, path: string): Promise<void> => {
```
jfsoul, line 21: nitpick `json: string` would be idiomatic/consistent naming and capitalisation I think?

packages/repo-fetcher/src/validGithubTeams.ts
```diff
@@ -0,0 +1,56 @@
...
+	{ name: 'infosec', engineering: false },
+	{ name: 'infosec-admin', engineering: false },
+	{ name: 'data-and-insight', engineering: false },
+	{ name: 'guardian-design-team', engineering: false },
+	{ name: 'guardian-us-design-team', engineering: false },
+	{ name: 'esd', engineering: false },
+	{ name: 'esd-admin', engineering: false },
+	{ name: 'enterprise-infrastructure', engineering: false },
+	{ name: 'it-australia', engineering: false },
+	{ name: 'glabs-au', engineering: false },
```
jfsoul, line 55: is any of this information private? I don't think so, but I thought I'd ask!

packages/repo-fetcher/src/handler.ts
```diff
@@ -36,7 +36,19 @@ const createOwnerObjects = async (
 
 export const main = async (): Promise<void> => {
 	console.log('[INFO] starting repo-fetcher');
+
+	await save(JSON.stringify(validGithubTeams), 'github/teamValidity.json');
```
jfsoul, line 40: I think if we don't need this data for `repo-fetcher` itself, then the data should simply live in the `api` project and be served from there - bypassing a datastore altogether. The "fetchers" are more for grabbing data from an external service.

OTOH if we need the data here then it could just live in `common` (not sure we ever will need this)

packages/repo-fetcher/src/validGithubTeams.ts
NovemberTang: good question. I don't think any of this information is particularly sensitive, we're not exposing any personal information. But I don't think this information was publicly accessible before. Is it worth closing the repo over? (tracker has the same info and is private)

jfsoul: I'm on the fence, on the one hand it could be used for social engineering, on the other we probably get more value from this being an open repo than the risk of that. Ask for a third opinion?

NovemberTang: if other org members are granted access, i'm not sure what the value of keeping this open is. Could you explain your thinking on

> we probably get more value from this being an open repo

@kenoir do you wanna weigh in?

jfsoul: I think just the usual [reasons](https://www.theguardian.com/info/developer-blog/2014/nov/28/developing-in-the-open) for working in the open, so nothing critical but also think risk is low on the other side

kenoir: Open by default 👍 If we feel weird about any of the configuration, maybe it shouldn't be hard-coded in the app anyway.

packages/repo-fetcher/src/handler.ts
```diff
@@ -10,19 +10,18 @@ import {
...
 	findOwnersOfRepo,
 	getAdminReposFromResponse,
 	RepoAndOwner,
 	transformRepo,
 } from '../src/transformations';
 
-const save = (repos: Repository[]): Promise<void> => {
-	const prefix = config.dataKeyPrefix;
-	const key = `${prefix}/github/repos.json`;
+const save = (json: string, path: string): Promise<void> => {
```
kenoir, line 20: Can you more closely type this so it's not string and `JSON.stringify` stays in this function? Perhaps by using a union type for all the saveable things?

packages/repo-fetcher/src/handler.ts
```diff
@@ -36,7 +35,10 @@ const createOwnerObjects = async (
 
 export const main = async (): Promise<void> => {
 	console.log('[INFO] starting repo-fetcher');
+
```
kenoir, line 38: NIT: This makes `main` a tiny bit crowded, is it reasonable to split some things out into other other functions for readability?

packages/api/src/validGithubTeams.ts
```diff
@@ -0,0 +1,56 @@
+export const validGithubTeams = [
+	{ name: 'android-admins', engineering: true },
```
kenoir, line 2: Perhaps we should move this config into S3 or out of the app eventually?

kenoir (approved):
A few suggestions, nothing blocking for merge if you want to iterate.

packages/api/src/validGithubTeams.ts
jfsoul: On the other hand, source code is more visible and editable by anyone with access! Anecdotally, we find maintaining S3 config difficult at the moment (e.g. Anghammarad, google chat bots etc)

NovemberTang: yeah this was my main concern. Given this data is going to be hand edited for at least the next while, having it in an audit trail and somewhere easy to fine seemed easier than having to authenticate, find the file, download it, edit it, and reupload it with no hard and fast rules about versioning.

kenoir: Agreed, this is the right place, right now. I have a suspicion this is actually part of a mapping from real business teams to GitHub teams and as such should perhaps be shared config from some other system.

PR #12, opened 2022-10-07
Add skeleton CDK Infrastructure for API

kenoir:
## What does this change?

This change adds a skeleton API app and Api Gateway lambda implementation.

In future PRs we should: 

- Add authorisation method (using pan-domain-auth or API key)
- Implement minimum passthrough response after retrieval from S3

## How to test

Tests pass, builds & deploys.

## How can we measure success?

API endpoint returns 200 OK from lambda.


package.json
```diff
@@ -6,13 +6,14 @@
...
-    "build": "npm run build --workspace=repo-fetcher",
+    "build": "npm run build --workspace=repo-fetcher --workspace=api",
     "lint": "eslint packages/** --ext .ts --no-error-on-unmatched-pattern"
   },
   "author": "Guardian Developers",
   "workspaces": [
     "packages/repo-fetcher",
-    "packages/cdk"
+    "packages/cdk",
+    "packages/api"
```
akash1810, line 16: We could make this `packages/*` for ease.

packages/api/package.json
```diff
@@ -3,7 +3,14 @@
   "version": "0.0.1",
   "description": "API that exposes a snapshot of your GitHub organisation",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects repo-fetcher",
+    "build": "esbuild src/*.ts --bundle --outdir=dist --platform=node --target=node16 --external:aws-sdk",
+    "esbuild": "esbuild",
+    "dev": "../../node_modules/.bin/ts-node ./src/handler",
```
akash1810, line 9: Don't think we need the path to node_modules, `ts-node ./src/hander` should be ok.

packages/cdk/package.json
```diff
@@ -11,6 +11,8 @@
     "aws-cdk": "2.39.1",
     "aws-cdk-lib": "2.39.1",
     "constructs": "10.1.50",
-    "source-map-support": "^0.5.21"
+    "source-map-support": "^0.5.21",
+    "@types/js-yaml": "^4.0.5",
+    "js-yaml": "^4.1.0"
```
akash1810, line 16: Don't think we're using these deps, shall we remove them?

riff-raff.yml
```diff
@@ -13,6 +13,18 @@ deployments:
       bucketSsmLookup: true
       fileName: repo-fetcher.zip
       lookupByTags: true
+  api:
```
akash1810, line 16: Would recommend waiting for https://github.com/guardian/github-lens/pull/16, and rebasing, as it simplifies this by auto-generating the `riff-raff.yaml` file.

packages/cdk/lib/github-lens.ts
```diff
@@ -65,7 +67,20 @@ export class GithubLens extends GuStack {
...
 
-		// TODO: Make DATA_KEY_PREFIX configurable
+		const noMonitoring: NoMonitoring = { noMonitoring: true };
+
+		new GuApiLambda(this, `${apiApp}-lambda`, {
+			fileName: `${apiApp}.zip`,
+			handler: 'handler.main',
+			runtime: Runtime.NODEJS_16_X,
+			monitoringConfiguration: noMonitoring,
+			app,
```
akash1810, line 77: Should this be `app: apiApp`?

packages/cdk/lib/github-lens.ts
```diff
@@ -38,6 +39,7 @@ export class GithubLens extends GuStack {
 
 		const paramPathBase = `/${this.stage}/${this.stack}/${app}`;
 		const repoFetcherApp = 'repo-fetcher';
+		const apiApp = 'api';
```
akash1810, line 42: Could we make this `github-lens-api`, to make the SSA tags more distinctive?

`Stack = deploy, Stage = INFRA, App = api` vs `Stack = deploy, Stage = INFRA, App = github-lens-api`

packages/cdk/lib/github-lens.ts
akash1810: FYI: If we do this, the `contentDirectories` in `.github/workflows/ci.yml` would also need changing.

packages/api/package-lock.json
```diff
@@ -0,0 +1,578 @@
+{
```
akash1810, line 1: A little surprised we have a dedicated lock file here, IIUC when using workspaces, there should only be one lockfile, sitting next to the main `package.json`.

packages/cdk/lib/github-lens.ts
akash1810: I wonder if `Stack = github-lens, Stage = INFRA, App = api` is better? 🤔 (creating new stacks require a little change in Riff-Raff config though)

kenoir: I think calling the app `github-lens-api` will result in slightly more human readable names in AWS, so i'm for it.

packages/cdk/lib/github-lens.ts
```diff
@@ -10,7 +13,7 @@ import { Key } from 'aws-cdk-lib/aws-kms';
 import { Runtime } from 'aws-cdk-lib/aws-lambda';
 
 export class GithubLens extends GuStack {
-	constructor(scope: App, id: string, props: GuStackProps) {
+	constructor(scope: App, id: string, props: GuStackProps, domainName: string) {
```
akash1810, line 16: minor, not blocking: something like this is a little more explicit, as the field is named, rather than positional:

```ts
interface GithubLensProps extends GuStackProps {
  domainName: string
}

class GithubLens extends GuStack {
  constructor(scope: App, id: string, props: GithubLensProps) {
    const { domainName } = props;

    // ... rest of the things
  }
}

new GithubLens(app, "GithubLens-INFRA", {
  stack: "deploy",
  stage: "INFRA",
  env: { region: "eu-west-1" },
  domainName: "github-lens.gutools.co.uk"
})
```

packages/github-lens-api/src/app.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { Express } from 'express';
+import request from 'supertest';
+import { buildApp } from './app';
+
+describe('github-lens api lambda', () => {
+	let app: Express;
+
+	beforeEach(() => {
+		app = buildApp();
+	});
```
akash1810, line 11: minor: with only one test, I don't think the `beforeEach` will have much effect?

akash1810 (approved):

PR #13, opened 2022-10-10
Some small bugfixes

kenoir:
## What does this change?

Some bug fixes for issues noticed while pairing with @akash1810.
- Update the way cron schedule is set in CDK to be a bit more readable
- Build the upload path to S3 using `path.join`
- Fix a JSON decoding issue when uploading `repos.json to S3`

## How to test

```sh
# Tests pass
npm run test

# Local run behaves as expected
 npm -w packages/repo-fetcher run dev
```


akash1810 (approved):
🚢

PR #15, opened 2022-10-13
Update ./script/setup.sh to setup node environment

kenoir:
## What does this change?

This change adds a `setup_node_env` script which tries to use `fnm` or `nvm` to configure your node environment and run `npm install`, in order to more comprehensively set up the project in a first time run.

## How to test

Run the script!

## How can we measure success?

Does this feel like an easier developer experience?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ The project consists of the following subprojects:
 
 ## Local Development
 
-If this your first time developing github-lens, you should run
+If this your first time developing github-lens, you should run:
+
 ```
 ./scripts/setup.sh
 ```
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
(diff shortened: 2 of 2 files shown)


akash1810 (approved):
[Not a _huge_ fan of this pattern](https://github.com/guardian/csnx/pull/64#issuecomment-1233255452), but happy to go with it.

It might be worth abstracting this sort of thing into a CLI tool that can be re-used to avoid having the same/similar code spread across various repositories.

kenoir:
> [Not a _huge_ fan of this pattern](https://github.com/guardian/csnx/pull/64#issuecomment-1233255452), but happy to go with it.
> 
> It might be worth abstracting this sort of thing into a CLI tool that can be re-used to avoid having the same/similar code spread across various repositories.

Not sure either. I'd say this doesn't make a choice for you if you've already got `nvm` installed and encodes a real preference we have.

PR #16, opened 2022-10-13
move from CODE to INFRA

JuliaBrigitte:
Since there is no need to have CODE and PROD because this is a low risk infrastructure application to which we can directly deploy, we moved from CODE to INFRA.

We also autogenerate riff-raff.yaml from now on.


.github/workflows/ci.yml
```diff
@@ -39,11 +39,17 @@ jobs:
         run: ./scripts/ci.sh
 
       - name: Upload to riff-raff
-        uses: guardian/actions-riff-raff@v1.1.3
+        uses: guardian/actions-riff-raff@v2
         with:
```
kenoir, line 43: I think the indentation here needs fixing.

kenoir: Ignore this, reviewing on my phone made this look weird.

kenoir (approved):

packages/cdk/bin/cdk.ts
```diff
@@ -1,7 +1,12 @@
...
-const app = new App();
+const app = new GuRootExperimental();
 
-new GithubLens(app, 'GithubLens-CODE', { stack: 'deploy', stage: 'CODE' });
+new GithubLens(app, 'GithubLens-INFRA', {
+	stack: 'deploy',
+	stage: 'INFRA',
+	env: { region: 'eu-west-1' },
+});
+//new GithubLens(app, "my-stack-PROD", {stack: 'deploy', stage: 'PROD'});
```
akash1810, line 12: Don't think we need this line?

JuliaBrigitte: removed the line, but can not merge on my own because not 100% sure what we need to delete after merging.

PR #17, opened 2022-10-13
Hand github client dependency to github functions

kenoir:
## What does this change?

This is slightly neater as the GitHub client dependency is handed in making it mockable and reducing the number of instances of client needed.

## How to test

Tests pass!

### DIFF
--- a/packages/common/aws/s3.ts
+++ b/packages/common/aws/s3.ts
@@ -18,10 +18,11 @@ export const putItem = async <T>(
 			ContentType: 'application/json; charset=utf-8',
 			ACL: 'private',
 		});
-
 		try {
 			await s3Client.send(command);
-			console.log('[INFO] Item uploaded to s3 successfully');
+			console.log(
+				`[INFO] Item uploaded to s3 successfully to: s3://${dataBucketName}/${key}`,
+			);
 		} catch (e) {
 			console.log(`[ERROR] ${(e as Error).message}`);
 		}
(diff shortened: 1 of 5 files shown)


packages/common/aws/s3.ts
```diff
@@ -18,10 +19,10 @@ export const putItem = async <T>(
 			ContentType: 'application/json; charset=utf-8',
 			ACL: 'private',
 		});
-
+		const s3Path = path.join('s3://', dataBucketName, key);
```
kenoir, line 22: Oop, I think I need to remove the path.join here.

packages/repo-fetcher/src/handler.ts
```diff
@@ -1,23 +1,24 @@
...
 	RepositoriesResponse,
 	TeamRepoResponse,
 } from '../../common/github/github';
 import {
+	getOctokit,
 	getReposForTeam,
 	listRepositories,
 	listTeams,
 } from '../../common/github/github';
-import type { Repository } from '../src/transformations';
```
akash1810, line 14: Surprised CI didn't catch this previously 🤔 .

akash1810 (approved):

packages/repo-fetcher/src/handler.ts
kenoir: I don't think it lints properly in CI, i'll add a step to ensure that happens.

kenoir: Ah, this is actually because our [`eslint` rules](https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-config-typescript/index.js) don't include [no-useless-path-segments](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-useless-path-segments.md).

PR #18, opened 2022-10-17
Improve logging

nicl:
## What does this change?

Adds support for log levels via a `LOG_LEVEL` env var. Largely based off: https://github.com/guardian/tracker/blob/cb2695a12ee0c446df6e4ffcc9969d5650b59dc1/packages/common/src/logger.ts.

## How to test

Try out different levels locally using e.g. 

    LOG_LEVEL=trace npm -w packages/repo-fetcher run dev

(Note you may want to add some trace log lines first.)

### DIFF
--- a/packages/common/log/log.ts
+++ b/packages/common/log/log.ts
@@ -1,0 +1,44 @@
+// These follow the non-aliased methods available on Node's console object:
+// https://nodejs.org/api/console.html#consoletracemessage-args.
+export const LOG_LEVELS = ['error', 'log', 'debug', 'warn', 'trace'] as const;
+
+export type LogLevel = typeof LOG_LEVELS[number];
+
+export function getLogLevel(logArg?: string): LogLevel {
+	const defaultLogLevel: LogLevel = 'log';
+
+	if (!logArg) {
+		return defaultLogLevel;
+	}
+
+	return LOG_LEVELS.includes(logArg as LogLevel)
+		? (logArg as LogLevel)
+		: defaultLogLevel;
+}
+
+// Overrides the global console.* functions so call early.
+export function configureLogging(logLevel: LogLevel) {
+	const shouldLog = (level: LogLevel) => {
+		return LOG_LEVELS.indexOf(level) <= LOG_LEVELS.indexOf(logLevel);
+	};
+
+	const _console = console;
+	global.console = {
+		...global.console,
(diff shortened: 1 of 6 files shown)


kenoir (approved):
👌

packages/repo-fetcher/src/handler.ts
```diff
@@ -40,13 +41,15 @@ const createOwnerObjects = async (
 };
 
 export const main = async (): Promise<void> => {
-	console.log('[INFO] starting repo-fetcher');
+	configureLogging(getLogLevel(process.env['LOG_LEVEL']));
```
akash1810, line 44: minor: could we push the reading of env vars into [`getConfig`](https://github.com/guardian/github-lens/blob/ebc769af0213fff421ceab1c497650101f20139b/packages/common/config.ts#L48), to keep configuration things co-located?

packages/common/log/log.ts
```diff
@@ -0,0 +1,44 @@
+// These follow the non-aliased methods available on Node's console object:
+// https://nodejs.org/api/console.html#consoletracemessage-args.
+export const LOG_LEVELS = ['error', 'log', 'debug', 'warn', 'trace'] as const;
```
akash1810, line 3: IIUC from the definition of `shouldLog`, the ordering of `LOG_LEVELS` is important. This ordering feels a little off - shouldn't the severity be increasing?

packages/common/log/log.ts
```diff
@@ -0,0 +1,44 @@
+// These follow the non-aliased methods available on Node's console object:
```
akash1810, line 1: minor: Worth noting, this implementation differs from [tracker's](https://github.com/guardian/tracker/blob/main/packages/common/src/logger.ts). Should we align them? (Either variant works).

Also, having this file in `common/log/log.ts` creates quite the import, as there aren't any other files in the `log` directory, could we simplify with simply `common/log.ts`?

akash1810 (commented):
Nice! QQ on `LOG_LEVELS`, and a couple of minor comments that can be addressed as follow-ups.

packages/common/log/log.ts
nicl: Order is definitely important here! `warn` is the one I wasn't sure of here to be honest but what were you thinking? (broadly it's L-R less logs to more).

PR #19, opened 2022-10-21
Jb/facilitate development

JuliaBrigitte:
## What does this change?

As a starting point the api now serves some endpoint on dev after running fetch.
For new the output is stored in the local directory test.
Kenny suggested moving this to a local S3, but as far as I know the current discussion point is to move the whole repo fetcher into tracker

## How to test

To fetch the data you need to run:

```
# Fetches data from GitHub
npm -w packages/repo-fetcher run dev

# Start the API locally
npm -w packages/github-lens-api run dev
```
This will start the api on localhost:3232
Running localhost will show the current endpoints
(both commands are provided in package.json as fetch and localapi)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "A lens into your GitHub organisation",
   "scripts": {
     "test": "jest --detectOpenHandles",
-    "start": "npm -w packages/repo-fetcher run dev",
+    "fetch": "npm -w packages/repo-fetcher run dev",
+    "localapi": "npm -w packages/github-lens-api run dev",
     "synth": "npm run synth --workspace=cdk",
     "typecheck": "npm run typecheck --workspaces",
     "build": "npm run build --workspace=repo-fetcher --workspace=github-lens-api",
@@ -20,6 +21,7 @@
     "aws-sdk-client-mock": "^2.0.0"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.190.0",
     "@types/jest": "^28.1.6",
     "@types/node": "^18.6.3",
     "dotenv": "^16.0.1",
--- a/packages/common/github/github.ts
+++ b/packages/common/github/github.ts
@@ -141,8 +141,6 @@ export const getTeamlistMembersInOrg = async (
 	);
 };
 
-//Get a team
-//SingleTeamResponse
 export const getTeamBySlug = async (
 	client: Octokit,
 	teamSlug: string,
@@ -155,9 +153,6 @@ export const getTeamBySlug = async (
 	return response.data;
 };
 
-//Get a single repository
-//https://octokit.github.io/rest.js/v19#repos
-//SingleRepoResponse
 export const getInfoForRepo = async (
 	client: Octokit,
 	owner: string,
@@ -171,11 +166,11 @@ export const getInfoForRepo = async (
 	return response.data;
 };
 
-let teamCounter = 0;
-const numberOfTeams = 2;
 export const listTeamsForLocalDevelopment = async (
 	client: Octokit,
 ): Promise<TeamsResponse> => {
+	let teamCounter = 0;
+	const numberOfTeams = 2;
 	return await client.paginate(
 		client.teams.list,
 		{
@@ -192,11 +187,11 @@ export const listTeamsForLocalDevelopment = async (
 	);
 };
 
-let repoCounter = 0;
-const numberOfRepos = 2;
 export const listRepositoriesForLocalDevelopment = async (
 	client: Octokit,
 ): Promise<RepositoriesResponse> => {
+	let repoCounter = 0;
+	const numberOfRepos = 2;
 	return await client.paginate(
 		client.repos.listForOrg,
 		{
--- a/packages/common/github/github.ts
+++ b/packages/common/github/github.ts
@@ -2,7 +2,7 @@ import { createAppAuth } from '@octokit/auth-app';
 import { throttling } from '@octokit/plugin-throttling';
 import { Octokit } from '@octokit/rest';
 import type { GetResponseDataTypeFromEndpointMethod } from '@octokit/types';
-import type { Config } from '../config';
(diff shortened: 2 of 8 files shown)


packages/common/github/github.ts
```diff
@@ -99,17 +111,141 @@ export const listTeams = async (client: Octokit): Promise<TeamsResponse> => {
...
+		{
+			org: 'guardian',
+			team_slug: teamSlug,
+			per_page: defaultPageSize,
+		},
+		(response) => response.data,
+	);
+};
+
+//Get a team
```
kenoir, line 144: I think this comment is redundant, you can work out what it does and what it returns by the function name and signature.

packages/common/github/github.ts
```diff
@@ -99,17 +111,141 @@ export const listTeams = async (client: Octokit): Promise<TeamsResponse> => {
...
+): Promise<SingleRepoResponse> => {
+	const response = await client.repos.get({
+		org: 'guardian',
+		owner: owner,
+		repo: repo,
+	});
+	return response.data;
+};
+
+let teamCounter = 0;
```
kenoir, line 174: These should be moved inside the `listTeamsForLocalDevelopment` function.

packages/common/github/github.ts
```diff
@@ -99,17 +111,141 @@ export const listTeams = async (client: Octokit): Promise<TeamsResponse> => {
...
+	const response = await client.teams.getByName({
+		org: 'guardian',
+		team_slug: teamSlug,
+		per_page: defaultPageSize,
+	});
+	return response.data;
+};
+
+//Get a single repository
+//https://octokit.github.io/rest.js/v19#repos
```
kenoir, line 159: As above, I think you can remove these comments.

packages/common/github/github.ts
```diff
@@ -99,17 +111,141 @@ export const listTeams = async (client: Octokit): Promise<TeamsResponse> => {
...
+			teamCounter += response.data.length;
+			if (teamCounter >= numberOfTeams) {
+				done();
+			}
+			return response.data;
+		},
+	);
+};
+
+let repoCounter = 0;
```
kenoir, line 195: As above, these should be within the scope of `listRepositoriesForLocalDevelopment` to avoid confusion.

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "A lens into your GitHub organisation",
   "scripts": {
     "test": "jest --detectOpenHandles",
-    "start": "npm -w packages/repo-fetcher run dev",
+    "fetch": "npm -w packages/repo-fetcher run dev",
+    "localapi": "npm -w packages/github-lens-api run dev",
```
kenoir, line 8: This task is usually called `run`.

packages/repo-fetcher/src/handler.ts
```diff
@@ -47,24 +67,103 @@ export const main = async (): Promise<void> => {
...
 
-	console.log(`Found ${teamNames.length} github teams`);
+	const stage = config.stage;
+	console.log(` stage is ${stage}`);
+
+	console.log(` running ListTeams`);
+	//teamNames array of teamNames
+	const teamNames = await getTeams(config.stage, client);
+	const teamNamesFileName = join(__dirname, '../../../test/teamNames.json');
+	if (stage === 'DEV' && !fs.existsSync(teamNamesFileName)) {
```
nicl, line 78: Discussed IRL but for the record, here's a way we could re-use some of this logic:

```
// Make an API call and cache if not already found
const withCache = async <A>(
	cacheName: string,
	fn: () => Promise<A>,
): Promise<A> => {
	if (!fs.existsSync(cacheName)) {
		const data = await fn();
		writeFileSync(cacheName + '.json', JSON.stringify(data), {
			flag: 'w',
		});

		return data;
	}

	const data = readFileSync(
		join(__dirname, `../../../test/${cacheName}.json`),
		'utf8',
	);

	return Promise.resolve(JSON.parse(data) as A);
};

// e.g.
const allRepos: TeamRepoResponse = await withCache('getReposForTeam', () =>
	getReposForTeam(client, teamSlug),
);
```

packages/github-lens-api/src/app.ts
```diff
@@ -19,9 +22,124 @@ export function buildApp(): Express {
...
+				{
+					endpoint: 'teams',
+					info: 'galaxies info about teamss',
+					testlink: 'http://localhost:3232/teams',
+				},
+			]);
+		});
+
+		const repos = readFileSync(
+			join(__dirname, '../../../test/repos.json'),
```
nicl, line 80: Discussed IRL, but it might be worth introducing a small interface here to make it easier to change this for local/remote and clarify things:

```
interface Store {
	getItem(name: string): Promise<string>;
}
class LocalStore implements Store {
	getItem(name: string): Promise<string> {
		// fetch from /test
	}
}

// then e.g.
const repos = store.get("repos")
```

packages/common/github/github.ts
JuliaBrigitte: done

packages/common/github/github.ts
JuliaBrigitte: done

packages/common/github/github.ts
JuliaBrigitte: done

packages/common/github/github.ts
JuliaBrigitte: done

packages/repo-fetcher/src/handler.ts
JuliaBrigitte: Thank you, did not manage to implement Nic's suggestions before I headed to Wales, will do once I am back.

packages/github-lens-api/src/app.ts
JuliaBrigitte: Thank you, did not manage to implement Nic's suggestions before I headed to Wales, will do once I am back.

PR #20, opened 2022-10-21
Add Localstack S3 interaction

kenoir:
## What does this change?

This change adds `localstack` configuration to allow the `repo-fetcher` lambda to store data in a local S3 instance, allowing the `github-lens` api lambda to also access this local data for ease of development.

It's also possible to interact with the local S3 instance to download and review any data the application uploads.

This change could be an improvement over internalising the caching logic within the application as it:

- Reduces the complexity of the application code making it easier to understand, work with & maintain.
- Developers working with `localstack` are presented with an experience closer to dealing with AWS APIs

It does require developers to be familiar with and use `docker-compose` in so far as it should be installed and there should be an easy way of starting it, but this should not be onerous compared to carrying the logic in the application going forward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.zip
 .idea
+.localstack
 .DS_Store
 data
 node_modules
--- a/README.md
+++ b/README.md
@@ -24,6 +24,20 @@ The setup script will download a starter `.env`.
 
 ### Running locally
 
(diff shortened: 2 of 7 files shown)


kenoir:
Closing in favour of using a DEV S3 bucket (similar DX but no localstack required!)

PR #22, opened 2022-10-24
Use configured bucket from environment locally

kenoir:
## What does this change?

This change is intended to improve the DX of this repository by allowing the lambda running locally to use the INFRA stack S3 bucket with the "DEV" prefix. It makes specifying a bucket name mandatory and expects this to be made available in a local `.env` file as follows:

```
DATA_BUCKET_NAME=github-lens-data-INFRA
```

There is only an INFRA stack for this application and the value of `DATA_BUCKET_NAME` should be that in the currently created stack.

Note: When this is deployed and a stack is created next we'll need to update the .env file in S3 to reference the S3 bucket created in the INFRA stack.

### DIFF
--- a/packages/cdk/lib/__snapshots__/github-lens.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/github-lens.test.ts.snap
@@ -865,6 +865,7 @@ Object {
     "githublensdatabucketGithublens7C07FDE7": Object {
       "DeletionPolicy": "Retain",
       "Properties": Object {
+        "BucketName": "github-lens-data-infra",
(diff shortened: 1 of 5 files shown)


nicl (approved):
Happy to merge but, just a thought, would prefer us to just use functions here. `S3Ops` isn't really doing anything - and separate (free-floating) functions are simpler and easier to test/use etc.

kenoir:
> Happy to merge but, just a thought, would prefer us to just use functions here

This is fair, the use of S3Ops was borrowed from `tracker` but I agree is unnecesary.

PR #24, opened 2022-10-27
Small refactor to improve readability

nicl:
Tweaks some namings and replaces RepoAndOwner class with an interface. The main aim is to make the handler more readable - some of this is subjective for sure but overall I think this is an improvement.

### DIFF
--- a/packages/repo-fetcher/src/handler.ts
+++ b/packages/repo-fetcher/src/handler.ts
@@ -1,10 +1,7 @@
 import type { Octokit } from '@octokit/rest';
 import { getS3Client, putObject } from '../../common/aws/s3';
 import { getConfig } from '../../common/config';
-import type {
-	RepositoriesResponse,
-	TeamRepoResponse,
-} from '../../common/github/github';
+import type { TeamsResponse } from '../../common/github/github';
 import {
 	getOctokit,
 	getReposForTeam,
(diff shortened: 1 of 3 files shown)


packages/repo-fetcher/src/handler.ts
```diff
@@ -1,31 +1,37 @@
...
+		const adminRepos: string[] = getAdminReposFromResponse(teamRepos);
+		return adminRepos.map((repoName) => ({
+			teamSlug: team.slug,
+			repoName,
+		}));
+	});
+
+	const flattened = (await Promise.all(teamRepositories)).flat();
+
+	return flattened.reduce<Record<string, string[] | undefined>>((acc, repo) => {
```
akash1810, line 30: Would `Record<string, string[]>` be an easier type than `Record<string, string[] | undefined` here? Where the empty array is the `undefined` case.

nicl: The problem is, we can't guarantee what is returned will always return a `string[]` for a given key (lookup) string.

Typescript will happily let you declare:

`Record<string, string[]>`

but really it's just an object, and there's no guarantee that x[foo] will return anything (as `foo` may not exist in `x`).

akash1810: True, however we're currently using nullish coalescing to handle that case, I think?

akash1810 (approved):
Added some comments, none of which are blocking.

packages/repo-fetcher/src/handler.ts
nicl: Ah to be clear, the value when a string is present will always be string[] but the `undefined` case is what is returned when you lookup a string in the record that doesn't exist.

PR #28, opened 2022-10-27
Make API private

nicl:
API Gateway supports private endpoints that can be accessed from within a VPC. See:

https://aws.amazon.com/blogs/compute/introducing-amazon-api-gateway-private-endpoints/

We'll use this to limit access to other services and also ourselves (requires being on the VPN).

See also: https://github.com/guardian/deploy-tools-platform/pull/623.

Update, this now works. It required some changes to the resource policy and also the domain handling. The API itself needs to be private, and have the default API endpoint enabled - as it is used by the VPC endpoint.

Depends on https://github.com/guardian/deploy-tools-platform/pull/623.

*Note: https://docs.aws.amazon.com/vpc/latest/reachability/what-is-reachability-analyzer.html was incredibly useful in debugging some of the networking issues here. https://aws.amazon.com/premiumsupport/knowledge-center/api-gateway-troubleshoot-403-forbidden/ was another useful resource here.*

Co-authored-by: Natasha Thrale <novembertang@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,10 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'npm'
 
+      - uses: guardian/actions-read-private-repos@v0.1.0
+        with:
+          private-ssh-keys: ${{ secrets.PRIVATE_INFRASTRUCTURE_CONFIG_DEPLOY_KEY }}
+
       - name: Run script/ci
         run: ./scripts/ci.sh
 
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -9,5 +9,5 @@ new GithubLens(app, 'GithubLens-INFRA', {
 	stage: 'INFRA',
 	env: { region: 'eu-west-1' },
 	domainName: 'github-lens.gutools.co.uk',
-	vpceId: 'vpce-0cb20360ab01108de',
+	vpceId: 'vpce-04f40fb1a4a0a23a6',
 });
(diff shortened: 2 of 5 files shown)


kenoir (approved):

.github/workflows/ci.yml
```diff
@@ -35,6 +35,10 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'npm'
 
+      - uses: guardian/actions-read-private-repos@v0.1.0
```
nicl, line 38: Possibly not required.

packages/cdk/lib/github-lens.ts
```diff
@@ -1,19 +1,36 @@
...
+import { LogGroup } from 'aws-cdk-lib/aws-logs';
 
 interface GithubLensProps extends GuStackProps {
+	// Domain name to use for the app/stage.
 	domainName: string;
+
+	// ID of a virtual private endpoint used to connect to this private API
+	// Gateway. For more info, see:
+	// https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html#apigateway-private-api-create-interface-vpc-endpoint.
+	vpceId: string;
```
akash1810, line 33: Can this be read via an SSM Parameter?

packages/cdk/bin/cdk.ts
```diff
@@ -9,4 +9,5 @@ new GithubLens(app, 'GithubLens-INFRA', {
 	stage: 'INFRA',
 	env: { region: 'eu-west-1' },
 	domainName: 'github-lens.gutools.co.uk',
+	vpceId: 'vpce-04f40fb1a4a0a23a6',
```
akash1810, line 12: This feels like private info, can we push it into SSM Parameters?

nicl:
Closing in favour of https://github.com/guardian/github-lens/pull/39. It's proved unnecessarily complex to support private access with a custom domain on API Gateway.

PR #29, opened 2022-10-27
feat: Allow collection of data for one GH team

akash1810:
## What does this change?

This project heavily relies on the GitHub API. This API is rate limited.

In an attempt to improve the local DX, in this change we add a new configuration property `GITHUB_TEAM`. When set, we reduce the amount of API calls made to hopefully yield a faster feedback loop when developing locally.

## How to test

- Checkout branch
- Set `GITHUB_TEAM` env var
- Run repo-fetcher

### DIFF
--- a/packages/common/src/config.ts
+++ b/packages/common/src/config.ts
@@ -1,7 +1,7 @@
 import * as dotenv from 'dotenv';
 import { decrypt } from './aws/kms';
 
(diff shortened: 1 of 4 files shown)


kenoir (approved):
Satisfying sensible solution 👍

kenoir:
Merging this as I would like to be able to start using it!

PR #31, opened 2022-10-28
Add getObject function to S3

kenoir:
## What does this change?

This change:
- Removes the error catching from `putObject` as it seems unclear what this is doing, and more expected to handle errors above this function.
- Adds a `getObject` function in preparation for returning data from the github-lens API.

## How to test

Tests should pass :)


nicl (approved):
Agree better for callers of Put to handle the error as appropriate.

PR #35, opened 2022-11-01
Output contents of repos.json on /repos

kenoir:
## What does this change?

This change returns the contents of the `repos.json` stored by repo-fetcher on the `/repos` endpoint in the `github-lens-api` endpoint. This allows the repo data to be read from the API!

Screenshot 2022-11-01 at 11 38 06

## How to test

Tests pass, after infra deployed the `/repos` endpoint will return data.

### DIFF
--- a/packages/github-lens-api/src/app.ts
+++ b/packages/github-lens-api/src/app.ts
@@ -1,10 +1,15 @@
+import path from 'path';
 import { json as jsonBodyParser } from 'body-parser';
+import { getObject, getS3Client } from 'common/aws/s3';
 import cors from 'cors';
 import express from 'express';
 import type { Express } from 'express';
+import asyncHandler from 'express-async-handler';
+import type { Config } from './config';
 
-export function buildApp(): Express {
+export function buildApp(config: Config): Express {
 	const app = express();
+	const s3Client = getS3Client(config.region);
 
 	app.use(jsonBodyParser());
 
@@ -19,9 +24,15 @@ export function buildApp(): Express {
 		res.status(200).json({ status: 'OK', stage: 'INFRA' });
 	});
 
-	app.get('/repos', (req: express.Request, res: express.Response) => {
-		res.status(200).json({});
-	});
+	app.get(
+		'/repos',
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const repoFileLocation = path.join(config.dataKeyPrefix, 'repos.json');
+			await getObject(s3Client, config.dataBucketName, repoFileLocation);
+
+			res.status(200).json({});
+		}),
+	);
 
 	return app;
 }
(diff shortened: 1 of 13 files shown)


packages/github-lens-api/src/handler.ts
```diff
@@ -1,13 +1,27 @@
...
+import { configureLogging } from 'common/log/log';
 import { buildApp } from './app';
+import { getConfig } from './config';
 
-configureLogging(getLogLevel(process.env['LOG_LEVEL']));
+const config = getConfig();
+const s3Client = getS3Client(config.region);
 
-dotEnvConfig();
+configureLogging(config.logLevel);
```
JuliaBrigitte, line 12: Glad this changed, because it did not work locally as it was before.

packages/github-lens-api/src/app.ts
```diff
@@ -19,9 +21,12 @@ export function buildApp(): Express {
 		res.status(200).json({ status: 'OK', stage: 'INFRA' });
 	});
 
-	app.get('/repos', (req: express.Request, res: express.Response) => {
-		res.status(200).json({});
-	});
+	app.get(
+		'/repos',
```
JuliaBrigitte, line 25: Maybe add app.get('/', for the poor people like me who click on Listening on port 3232
Access via http://localhost:3232
in the IDE and then get 
Cannot GET /

packages/repo-fetcher/src/handler.ts
```diff
@@ -66,7 +67,9 @@ export const main = async (): Promise<void> => {
 	);
 
 	const s3Client = getS3Client(config.region);
-	await putObject(s3Client, config.dataBucketName, config.dataKeyPrefix, repos);
+	const repoFileLocation = path.join(config.dataKeyPrefix, 'repos.json');
```
JuliaBrigitte, line 70: Maybe not the right place for this comment, but I am wondering if we might want to include a field info_last_retrieved in our json?

JuliaBrigitte (approved):

packages/github-lens-api/src/app.ts
kenoir: I like the self-describing API idea, I think that's one for a future PR.

packages/repo-fetcher/src/handler.ts
kenoir: Yes, I think so.

packages/github-lens-api/src/app.ts
kenoir: Addressed by https://github.com/guardian/github-lens/pull/36

packages/repo-fetcher/src/handler.ts
kenoir: Addressed by https://github.com/guardian/github-lens/pull/37

PR #36, opened 2022-11-02
Add router describing request handler

kenoir:
## What does this change?

This change adds a special request handler that can parse the Express routing stack and report a list of endpoints with their methods.

Requesting `/` will return the following:

```json
[
  {
    "path": "/",
    "methods": [
      "get"
    ]
  },
  {
    "path": "/healthcheck",
    "methods": [
      "get"
    ]
  },
  {
    "path": "/repos",
    "methods": [
      "get"
    ]
  }
]
```

This is intended to make it easier for developers or folk working with the API to discover and browse endpoints in this service.

### DIFF
--- a/packages/github-lens-api/src/app.ts
+++ b/packages/github-lens-api/src/app.ts
@@ -1,32 +1,38 @@
 import { json as jsonBodyParser } from 'body-parser';
 import type { Repository } from 'common/github/github';
 import cors from 'cors';
-import express from 'express';
 import type { Express } from 'express';
+import express, { Router } from 'express';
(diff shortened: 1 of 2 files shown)


JuliaBrigitte:
There must have been a misunderstanding, now we have two solutions:
https://github.com/guardian/cdk-metadata/pull/25

Maybe you can check mine and I check yours.

kenoir:
> There must have been a misunderstanding, now we have two solutions:

Two solutions for two APIs in different projects & languages so I think it's fine!

nicl (approved):

PR #37, opened 2022-11-02
Add last-modified date to the data retrieved from S3

kenoir:
## What does this change?

This change adds a `lastModified` field to the object returned from S3 using the `RetrievedObject` interface.

The last modified date is determined from the S3 metadata for a retrieved object.

The intention is to improve the usefulness of the data returned from the API i.e.f if the data in S3 hasn't been refreshed it's useful to know that!

The API response looks like this:

Screenshot 2022-11-02 at 15 30 31

**Note:** In a future PR we may want to introduce a separate response type and parse the `RetrievedObject` into that type. 

## How to test

`npm run test`

### DIFF
--- a/packages/common/src/aws/s3.ts
+++ b/packages/common/src/aws/s3.ts
@@ -28,11 +28,16 @@ export const putObject = async <T>(
 	console.info(`Item successfully uploaded to: s3://${bucketName}/${key}`);
 };
 
+export interface RetrievedObject<T> {
+	payload: T;
+	lastModified?: Date;
+}
+
 export const getObject = async <T>(
 	s3Client: S3Client,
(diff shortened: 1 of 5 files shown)


packages/common/src/aws/s3.ts
```diff
@@ -28,18 +28,24 @@ export const putObject = async <T>(
 	console.info(`Item successfully uploaded to: s3://${bucketName}/${key}`);
 };
 
+export interface RetrievedObject<T> {
```
akash1810, line 31: minor: Might `GetObjectResponse` be a more descriptive name here? In that it includes the HTTP verb?

akash1810 (approved):

PR #39, opened 2022-11-08
Make API private by moving to ALB/EC2

nicl:
**Update: test with https://github-lens.gutools.co.uk/healthcheck !**

Security Groups (network restrictions) are used to limit API access to within the VPC and peered Guardian networks - specifically the VPN.

API Gateway can be used to achieve this too but it is difficult to use a custom domain when doing so. See e.g.

https://georgemao.medium.com/enabling-private-apis-with-custom-domain-names-aws-api-gateway-df1b62b0ba7c

which requires an ALB for this.

If, as it seems, an ALB is required anyway, then using API Gateway seems to only add complexity here.

The shift to EC2 is not strictly speaking required but provides the following:

* consistency with Cloudformation Lens (nee cdk-metadata)
* ability to use our CDK GuEc2App pattern (without this, the CDK is even more custom/verbose)
* a larger response payload (lambda behind ALB is limited to 1mb)

It is clear though that even using our GuEc2App pattern is non-trivial atm in terms of user data and we should improe this story.

The cost impact of switching to EC2 here is $7/month for a single t4g.nano instance (not reserved).

Longer-term it may make sense to move to a monorepo and share the ALB across the services.gutools.co.uk-related services.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3120,23 +3120,6 @@
       "version": "1.0.3",
       "license": "MIT"
     },
-    "node_modules/@types/aws-lambda": {
-      "version": "8.10.107",
-      "resolved": "https://registry.npmjs.org/@types/aws-lambda/-/aws-lambda-8.10.107.tgz",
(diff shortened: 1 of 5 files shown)


akash1810:
> This is a lot easier than I would have expected!

💯 ! Great work @nicl !

PR #40, opened 2022-11-09
list valid endpoints for invalid api calls

JuliaBrigitte:
## What does this change?

list valid endpoints for invalid api calls

## How to test

Call with invalid endpoint


akash1810 (approved):

kenoir:
I believe this should still `404` all invalid endpoints, even if we do return the list of valid endpoints in the response body.

PR #41, opened 2022-11-14
Add Services API

nicl:
Just a 'Hello, world' really atm - adds a healthcheck.

Keen to add Google Auth but want to do this by extending the CDK pattern rather than bespoke here (which is verbose).


packages/cdk/lib/services-api.ts
```diff
@@ -0,0 +1,86 @@
...
+systemctl start ${app}
+`;
+
+		const ec2 = new GuEc2App(this, {
+			app,
+			access: {
+				scope: AccessScope.INTERNAL,
+				cidrRanges: [Peer.ipv4('10.0.0.0/8')], // TODO replace with Google auth once integrated to this pattern.
+			},
+			instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.NANO),
```
kenoir, line 61: 🐜 💪

packages/services-api/src/app.ts
```diff
@@ -0,0 +1,30 @@
...
+			credentials: true,
+		}),
+	);
+
+	router.get('/healthcheck', (req: express.Request, res: express.Response) => {
+		res.status(200).json({ status: 'OK', stage: 'INFRA' });
+	});
+
+	//handle all invalid routes by showing all available routes
+	router.get('*', getDescribeRouterHandler(router));
```
kenoir, line 25: Not for here, but I reckon the `getDescribeRouterHandler` should `404` while still dispensing the routes in the body.

kenoir (approved):

packages/services-api/src/app.ts
kenoir: https://github.com/guardian/github-lens/pull/42

packages/services-api/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "description": "API for Guardian services",
+  "scripts": {
+    "test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects services-api",
+    "prebuild": "rm -rf dist",
+    "build": "esbuild src/handler.ts --bundle --outdir=dist --platform=node --target=node16 --external:aws-sdk",
+    "dev": "LOCAL=true ts-node ./src/handler",
+    "start": "npm run dev",
+    "typecheck": "tsc -noEmit"
+  },
+  "devDependencies": {
```
akash1810, line 13: Some/most of these devDeps are common across the other workspaces, and shouldn't need repeating. Not a blocker, but we should revise how things are setup, pushing as much to the root package.json, if only to have a simpler version bumping story.

PR #42, opened 2022-11-14
Return 404 when path not found

kenoir:
## What does this change?

`getDescribeRouterHandler` is used to list available routes when a user provides a path that is not found. This change returns the correct status code for not found routes (404).

### DIFF
--- a/packages/github-lens-api/src/app.ts
+++ b/packages/github-lens-api/src/app.ts
@@ -22,8 +22,6 @@ export function buildApp(
 		}),
 	);
 
-	router.get('/', getDescribeRouterHandler(router));
-
 	router.get('/healthcheck', (req: express.Request, res: express.Response) => {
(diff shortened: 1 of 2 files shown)


packages/github-lens-api/src/app.ts
```diff
@@ -22,8 +22,6 @@ export function buildApp(
 		}),
 	);
 
-	router.get('/', getDescribeRouterHandler(router));
```
nicl, line 25: It's quite nice having this on the base path. Perhaps we should refactor to separate out the path list code from the specific status code and keep this, while also returning a 404 on the `*` catchall?

kenoir: > Perhaps we should refactor to separate out the path list code from the specific status code and keep this
I implemented it that way on the first go, but it felt unnecessarily complex. Seemed simpler to handle all undefined routes (including `/`) the same way.

nicl (approved):

PR #43, opened 2022-11-14
Repository setup

kenoir:
## What does this change?

This change adds:

- **Dependabot**: Stay up to date, avoiding vulnerabilities ensuring best practise (e.g. ensure GuCDK updates)
- **PR template**: Make it easier for others to contribute
- **Snyk integration**: Provide vulnerability monitoring
- **Updated .gitgnore**: More specific, machine-generated .gitignore to prevent accidental additions and reproducible rules.

It's intended to improve the DX in this repository, but also as a bit of a test to see what we might add by default to others new repositories.


.gitignore
```diff
@@ -1,9 +1,186 @@
-*.zip
```
akash1810, line 1: Not a huge fan of these generic .gitignore files, as they're overly verbose (we'll add webpack or svelte to this repo, for example). We've also lost the `*.zip` rule in this change; was that intentional?

kenoir: My reasoning for preferring these generated `.gitignore` files is that I prefer the "one size fits all approach", I don't have to understand in too much detail what a particular framework / IDE or language does. I can hone it if I need to (like that *.zip), and this file is rarely visited so not much mental overhead in the verbosity.

akash1810: Still a bit skeptical, but not enough to block. 🚢

akash1810 (approved):

.gitignore
kenoir: If we end up changing this file much in the future, that's a smell we should revert to the less verbose form.

PR #49, opened 2022-11-15
add endpoint /reposByName/:name

JuliaBrigitte:
## What does this change?
add endpoint /reposByName/:name (use e.g. add endpoint /reposByName/riff-raff
small refactor to make interface repository easier to find in preparation to adding more


packages/github-lens-api/src/app.ts
```diff
@@ -33,6 +33,20 @@ export function buildApp(
 		}),
 	);
 
+	router.get(
+		'/reposByName/:name',
```
akash1810, line 37: Are we likely to have other repository focused endpoints? For example get repositories by team? If so, would a URL of `/repos/by-name/:name` be simpler, as we can then have `/repos/by-team/:team`?

akash1810 (approved):
Added a non-blocking comment, but LGTM overall! 👍🏽

packages/github-lens-api/src/app.ts
kenoir: This is useful reading for a guide to REST API naming (not necessarily strictly applied, but useful to stick roughly to): https://restfulapi.net/resource-naming/

packages/common/src/github/github.ts
```diff
@@ -34,25 +34,6 @@ export type TeamRepoResponse = GetResponseDataTypeFromEndpointMethod<
 >;
 export type RepositoryResponse = RepositoriesResponse[number];
```
nicl, line 35: Not part of this PR, but this was interesting to me for a couple of reasons:

1) it's an interesting Typescript type feature (to index into another type and yield a type)
2) we're exporing Github's own types from this package. I would prefer we export our own types so that Github's don't leak but not something to change here.

packages/github-lens-api/src/app.ts
nicl: Agree with @kenoir - `/repos/:name` looks good here.

packages/github-lens-api/src/app.ts
```diff
@@ -33,6 +33,20 @@ export function buildApp(
...
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const repoDataJson = await repoData;
+			const searchForName = req.params.name;
```
nicl, line 40: Minor, but I think this const is probably redundant here - it's not clear what gain there is over inlining.

packages/github-lens-api/src/app.ts
```diff
@@ -33,6 +33,20 @@ export function buildApp(
...
+			const jsonResponse = repoDataJson.payload.filter((item) =>
+				item.name.includes(searchForName),
+			);
+			//res.send(searchForName);
```
nicl, line 44: (To remove.)

packages/github-lens-api/src/app.ts
```diff
@@ -33,6 +33,20 @@ export function buildApp(
 		}),
 	);
 
+	router.get(
+		'/reposByName/:name',
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const repoDataJson = await repoData;
+			const searchForName = req.params.name;
+			const jsonResponse = repoDataJson.payload.filter((item) =>
```
nicl, line 41: I'd be tempted to rename this to drop the `json` prefix, as it isn't actually JSON at this point.

nicl (approved):

packages/github-lens-api/src/app.ts
```diff
@@ -33,6 +33,20 @@ export function buildApp(
 		}),
 	);
 
+	router.get(
+		'/reposByName/:name',
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const repoDataJson = await repoData;
```
nicl, line 39: Possibly rename to drop the `json` suffix, as this isn't actually JSON here.

nicl: Ideally we'd not pass this data as a Promise but `await` in the `handler.ts` file. I know this requires 'top-level await' though, and not sure how easy it is to configure TS to support that, so ignore if proves too tricky.

packages/github-lens-api/src/app.ts
JuliaBrigitte: Actually what I implemented was a search, so maybe I will change it and do both.

JuliaBrigitte:
Did alter routes to match discussion. Added separate search

PR #55, opened 2022-11-19
Add teams data

kenoir:
## What does this change?

This change stores a list of teams in S3 along with the repositories they own so this data can be made available in the github-lens API.  It also adds `/teams` and `/teams/:name` endpoints.

Screenshot 2022-11-24 at 10 40 17

## Why?

More complete view on GitHub, and this data can then be used by `services.gutools` to identify which P&E teams own which repositories.

### DIFF
--- a/packages/github-lens-api/src/app.ts
+++ b/packages/github-lens-api/src/app.ts
@@ -1,6 +1,6 @@
 import { json as jsonBodyParser } from 'body-parser';
 import type { RetrievedObject } from 'common/aws/s3';
-import type { Repository } from 'common/model/repository';
+import type { Repository, Team } from 'common/model/repository';
 import cors from 'cors';
 import type { Express } from 'express';
 import express, { Router } from 'express';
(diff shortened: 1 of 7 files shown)


packages/github-lens-api/src/app.ts
```diff
@@ -60,6 +61,31 @@ export function buildApp(
...
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const teamsData = await teamData;
+			const jsonResponse = teamsData.payload.filter(
+				(item) => item.slug === req.params.name,
+			);
+			if (jsonResponse.length !== 0) {
+				res.status(200).json(jsonResponse);
+			} else {
+				res
+					.status(200)
```
nicl, line 83: Should this 404?

packages/repo-fetcher/src/transformations.ts
```diff
@@ -37,7 +37,7 @@ export const asRepo = (
 		is_template: repo.is_template,
 		topics: repo.topics,
 		default_branch: repo.default_branch,
-		owners: owners,
+		owners: owners ? owners : [],
```
nicl, line 40: `owners ?? []`?

nicl (approved):

PR #56, opened 2022-11-21
List services for teams

nicl:
This is it, the big one, the one you've all been waiting for...

P+E Teams (from Galaxies) are combined with Cloudformation stacks and Github Repo information. The link points are:

* the 'gu:repo' tag on Cfn stacks
* the 'owners' info from Github Lens, which lists Github Teams that are admins for a repo
* the 'primaryGithubTeam' property on Galaxy teams

Note, this does not attempt any caching yet so each request makes a couple of API calls behind the scenes. We should implement that in a follow-up though it may not be a concern for a while practically speaking given the expected low volume of requests.

Screenshot 2022-11-21 at 10 33 51

### DIFF
--- a/packages/cdk/lib/services-api.ts
+++ b/packages/cdk/lib/services-api.ts
@@ -43,23 +43,6 @@ export class ServicesApi extends GuStack {
 			galaxiesBucketParam.valueAsString,
 		);
 
-		const cloudformationLensUrlParam = new CfnParameter(
-			this,
-			'cloudformation-lens-url',
-			{
-				description:
-					'Cloudformation Lens URL including protocol but not trailing slash.',
-				default: `/${this.stage}/${this.stack}/${app}/cloudformation-lens-url`,
-				type: 'AWS::SSM::Parameter::Value<String>',
-			},
-		);
-
-		const githubLensUrlParam = new CfnParameter(this, 'github-lens-url', {
-			description: 'Github Lens URL including protocol but not trailing slash.',
-			default: `/${this.stage}/${this.stack}/${app}/github-lens-url`,
-			type: 'AWS::SSM::Parameter::Value<String>',
-		});
-
 		const applicationPort = 8900;
 		const handler = 'handler.js';
 
@@ -72,8 +55,6 @@ Description=Github Lens API
 Environment="PORT=${applicationPort}"
 Environment="STAGE=${props.stage}"
 Environment="GALAXIES_BUCKET_NAME=${galaxiesBucketParam.valueAsString}"
-Environment="CLOUDFORMATION_LENS_URL=${cloudformationLensUrlParam.valueAsString}"
-Environment="GITHUB_LENS_URL=${githubLensUrlParam.valueAsString}"
 ExecStart=/usr/bin/node /${handler}
 
 [Install]
--- a/packages/cdk/lib/services-api.ts
+++ b/packages/cdk/lib/services-api.ts
@@ -43,6 +43,23 @@ export class ServicesApi extends GuStack {
 			galaxiesBucketParam.valueAsString,
 		);
 
+		const cloudformationLensUrlParam = new CfnParameter(
+			this,
+			'cloudformation-lens-url',
+			{
+				description:
+					'Cloudformation Lens URL including protocol but not trailing slash.',
+				default: `/${this.stage}/${this.stack}/${app}/cloudformation-lens-url`,
+				type: 'AWS::SSM::Parameter::Value<String>',
+			},
+		);
+
+		const githubLensUrlParam = new CfnParameter(this, 'github-lens-url', {
+			description: 'Github Lens URL including protocol but not trailing slash.',
+			default: `/${this.stage}/${this.stack}/${app}/github-lens-url`,
+			type: 'AWS::SSM::Parameter::Value<String>',
+		});
+
 		const applicationPort = 8900;
 		const handler = 'handler.js';
 
@@ -55,6 +72,8 @@ Description=Github Lens API
 Environment="PORT=${applicationPort}"
 Environment="STAGE=${props.stage}"
 Environment="GALAXIES_BUCKET_NAME=${galaxiesBucketParam.valueAsString}"
+Environment="CLOUDFORMATION_LENS_URL=${cloudformationLensUrlParam.valueAsString}"
+Environment="GITHUB_LENS_URL=${githubLensUrlParam.valueAsString}"
 ExecStart=/usr/bin/node /${handler}
 
 [Install]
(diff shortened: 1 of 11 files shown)


packages/cdk/lib/services-api.ts
```diff
@@ -43,6 +43,23 @@ export class ServicesApi extends GuStack {
...
+				default: `/${this.stage}/${this.stack}/${app}/cloudformation-lens-url`,
+				type: 'AWS::SSM::Parameter::Value<String>',
+			},
+		);
+
+		const githubLensUrlParam = new CfnParameter(this, 'github-lens-url', {
+			description: 'Github Lens URL including protocol but not trailing slash.',
+			default: `/${this.stage}/${this.stack}/${app}/github-lens-url`,
+			type: 'AWS::SSM::Parameter::Value<String>',
+		});
```
akash1810, line 61: We're already [defining these in VCS](https://github.com/guardian/github-lens/blob/97e613f7d5021ae814f0443cdf4429b1fe279c1c/packages/cdk/bin/cdk.ts#L12), could we pass the values straight in, removing the CFN Params?

packages/services-api/src/services.ts
```diff
@@ -0,0 +1,125 @@
...
+	if (!repoTag) {
+		return [];
+	}
+
+	// find the repo and return its owners.
+	const owners = repos.find((repo) => repo.full_name === repoTag)?.owners;
+	return owners ?? [];
+};
+
+export const groupStacksByStage = (stacks: Stack[]): Stack[][] => {
```
akash1810, line 105: minor: Would a `Record<string, Stack[]>` be a simpler type here. It would make retrieval of the grouping stage value easier, as the value is the key, rather than reaching into the first element of the array.

```json
{
  "CODE": [
    { "stack": "deploy", "stage": "CODE" },
    { "stack": "security", "stage": "CODE" },
  ],
  "PROD": [
    { "stack": "deploy", "stage": "PROD" },
    { "stack": "security", "stage": "PROD" },
  ]
}
```

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "ES2021",
     "module": "commonjs",
     "lib": ["ES2021", "dom"],
-    "moduleResolution": "node",
+    "moduleResolution": "nodenext",
```
akash1810, line 6: Oooh, why is this needed?

packages/services-api/src/services.ts
```diff
@@ -0,0 +1,125 @@
...
+	readonly githubLensUrl: string;
+
+	constructor(cloudformationLensUrl: string, githubLensUrl: string) {
+		this.cloudformationLensUrl = cloudformationLensUrl;
+		this.githubLensUrl = githubLensUrl;
+	}
+
+	async list(): Promise<Service[]> {
+		const stacksResp = await fetch(`${this.cloudformationLensUrl}/stacks`);
+		const stacks = (await stacksResp.json()) as Stack[];
```
akash1810, line 71: Does this correctly cast the `Date` fields in `Stack`? Or do we want to use something like [Zod](https://zod.dev/)?

nicl: Yeah, try this: `JSON.stringify(new Date())`.

packages/services-api/src/app.ts
```diff
@@ -34,21 +41,38 @@ export function buildApp(config: Config): Express {
...
+
 	// TODO include Cloudformation stacks for each team to get to MVP!
 
 	router.get(
 		'/teams',
 		asyncHandler(async (req: express.Request, res: express.Response) => {
 			const teams = await galaxiesApi.getTeams();
-			res.status(200).json(teams);
+			const services = await servicesApi.list();
+			const teamsWithServices = teams.map((team) => {
```
kenoir, line 56: I think we should eventually do this upstream: https://github.com/guardian/github-lens/pull/55

tsconfig.json
nicl: Thanks, have now removed - was experimenting with adding ESM support, but abandoned for now.

packages/services-api/src/app.ts
nicl: So there are P+E teams and then Github Teams. Here, we're dealing with the former, whereas #55 is the latter if I've understood correctly? Exposing github teams in Github Lens is definitely useful, but I'm not sure it helps here.

packages/services-api/src/services.ts
nicl: Good spot! This currently works by accident - the type system thinks the dates are `Date` and, because we never manipulate them, they just pass through. How about we just model the date fields as `string` for now and explore Zod etc. later if/when we need something richer?

akash1810: Generally, I'm not a huge fan of using the `string` type as a catch all, as, for example, it makes comparisons more complicated.

However, the suggestion is very pragmatic 👍🏽 . Could we add some todos around it as well?

packages/services-api/src/app.ts
kenoir: Perhaps I should say we should also join P&E team -> GitHub team -> GitHub repositories "admined" by team.

packages/services-api/src/services.ts
nicl: Have added a note. Completely agree that string isn't a good choice if/when we start doing stuff with these dates.

packages/services-api/src/app.ts
nicl: Ah okay thanks. We currently achieve that through the Github Lens `/repos` endpoint, but it would be easier with a `/teams` endpoint in GH Lens yep.

packages/services-api/src/services.ts
```diff
@@ -0,0 +1,130 @@
...
+	id: number;
+	name: string;
+	full_name: string;
+	private: boolean;
+	description: string | null;
+	created_at: string | null;
+	updated_at: string | null;
+	pushed_at: string | null;
+	size: number | undefined;
+	language: string | null | undefined;
```
akash1810, line 24: The typing here is interesting; is this down to GitHub's types? Or could we simplify with:

```ts
language?: string;
```

packages/services-api/src/services.ts
```diff
@@ -0,0 +1,130 @@
...
+	const groups: Map<string, Stack[]> = new Map();
+
+	stacks.forEach((stack) => {
+		const app = stack.tags['App'];
+		const stackTag = stack.tags['Stack'];
+		const accountId = stack.accountId;
+
+		// Requirement: stacks should have app/stack/stage tags for grouping to
+		// work.
+		const key =
```
akash1810, line 120: minor: would `serviceName` be a more descriptive variable name here?

akash1810 (approved):
Couple of non-blocking comments.

packages/services-api/src/services.ts
nicl: This mirrors the existing types (from GH Lens) - which include null apparently. I've not changed that here, but we could definitely investigate as a follow-up if this can be rationalised, though I suspect this follows the GH types.

packages/services-api/src/services.ts
nicl: It's not technically the service name - just a unique key for a service.

PR #57, opened 2022-11-22
chore(deps): Simplify some dependencies by moving them to the root

akash1810:
## What does this change?
Pushes the following dependencies to the root package.json:
  - `@guardian/eslint-config-typescript`
  - `@guardian/prettier`
  - `typescript`
  - `jest`
  - `ts-jest`
  - `@types/jest`
  - `esbuild`

These dependencies are used in all (/the majority of workspaces). By installing them once, we achieve a simpler dependency updating story as bumping happens in one place.

These dependencies are also updated to the latest available version.

Also:
  - Removes `tslib` as unused
  - Removes `ts-node-dev` in favour of `ts-node`
  - Makes the CI script less verbose

## Why?
To tell a simpler dependency story.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13304,7 +13304,6 @@
         "supertest": "^6.1.3",
         "ts-jest": "^27.0.3",
         "ts-node-dev": "^1.1.6",
-        "tslib": "^2.3.0",
         "typescript": "^4.8.4"
       }
     },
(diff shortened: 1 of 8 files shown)


kenoir (approved):
👏 I noticed this at the weekend. Great to see it fixed.

PR #60, opened 2022-11-24
Migrate cloudformation-lens here

kenoir:
## What does this change?

This change pulls https://github.com/guardian/cloudformation-lens into this repository (along with its commit history). 

We are treating `cloudformation-lens` as a npm workspace from the top level, adding a `package.json` to fulfil the standard tasks (`build`, `test`) to simplify the top level build (it seems to work ok!).

This code is best reviewed by looking at the code changes, not the commit history (bar the last few commits to make the build work here)!

*Note:* The cloudformation-lens stack will need to be destroyed before this change is applied in order to allow this deployment to use the `https://cloudformation-lens.gutools.co.uk/` domain name.

## Why?

`cloudformation-lens` is called by the services API and it generally makes sense to keep all this related code (and services that will be deployed together) here.


packages/cloudformation-lens/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "cloudformation-lens",
+  "version": "0.0.1",
+  "description": "API for cloudformation stacks",
+  "scripts": {
+    "test": "go test ./...",
+    "build": "go test ./... && env GOOS=linux GOARCH=arm64 go build -o cloudformation-lens main.go",
+    "dev": "go run main.go --profile deployTools --in-memory",
+    "start": "npm run dev",
+    "lint": "true",
```
nicl, line 10: `go vet ./...` is a good option here.

packages/cloudformation-lens/package.json
```diff
@@ -0,0 +1,13 @@
...
+    "typecheck": "true"
```
nicl, line 11: `go build` is a good option for this.

packages/cloudformation-lens/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "cloudformation-lens",
+  "version": "0.0.1",
+  "description": "API for cloudformation stacks",
+  "scripts": {
+    "test": "go test ./...",
+    "build": "go test ./... && env GOOS=linux GOARCH=arm64 go build -o cloudformation-lens main.go",
```
nicl, line 7: I'd be tempted to drop the arch stuff here as my expectation would be that it would build it for my own system. But perhaps this is more consistent with the other scripts?

kenoir: This will build the binary that ends up getting deployed, and I don't think guarantee the CI architecture will be the same as the EC2 instance architecture it ends up on. I'd be concerned there will be a mismatch resulting in a broken deployment.

packages/cloudformation-lens/package.json
kenoir: Practically using `go build` will result in unnecessary execution of the command in CI. If you're developing on this service, I don't think you'll ever run `npm run typecheck` here so i'd rather just avoid doing anything.

nicl (approved):

PR #63, opened 2022-11-25
Fix snyk monitoring, don't exclude packages

kenoir:
## What does this change?

Snyk monitoring [has broken](https://github.com/guardian/service-catalogue/actions/runs/3547547273/jobs/5957785700#step:13:16), we need to update `package-lock.json`.

This change also excludes only the npm workspaces already covered by monitoring the root `package.json`. This allows us to cover`cloudformation-lens` which is written in go!

## Why?

So we can continue to monitor and remediate vulnerabilities.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -14,6 +14,6 @@ jobs:
       DEBUG: true
       ORG: guardian-devtools
       SKIP_NODE: false
-      NODE_PACKAGE_JSON_FILES_MISSING_LOCK: package.json
+      NODE_PACKAGE_JSON_FILES_MISSING_LOCK: service-catalogue/package.json
     secrets:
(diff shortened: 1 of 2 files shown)


akash1810 (commented):
Why does `package-lock.json` receive an update with this change?

kenoir:
> Why does `package-lock.json` receive an update with this change?

Looking at the output of the broken build in main, Snyk complains `package-lock.json` is out of sync with `package.json`. This change regenerates it to make Snyk happy.

nicl (approved):

akash1810:
> > Why does `package-lock.json` receive an update with this change?
> 
> Looking at the output of the broken build in main, Snyk complains `package-lock.json` is out of sync with `package.json`. This change regenerates it to make Snyk happy.

This feels a bit strange; should the general CI complain about this too, as we're using `npm ci`? 🤔

PR #64, opened 2022-11-25
Updates the app name to "service catalogue"

kenoir:
## What does this change?

Changes the app name to "service catalogue" so that riff-raff will call the project by its new name when deploying services. 

This should have no impact on the deployed services.

## Why?

To reduce confusion when the project has a different name than that deployed under.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Upload to riff-raff
         uses: guardian/actions-riff-raff@v2
         with:
-          app: github-lens
+          app: service-catalogue
           buildNumber: ${{ env.GITHUB_RUN_NUMBER }}
           projectName: deploy::github-lens
           configPath: packages/cdk/cdk.out/riff-raff.yaml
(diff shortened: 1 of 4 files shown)


NovemberTang (approved):

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Upload to riff-raff
         uses: guardian/actions-riff-raff@v2
         with:
-          app: github-lens
+          app: service-catalogue
```
akash1810, line 48: This'll have no effect. [`guardian/actions-riff-raff`](https://github.com/guardian/actions-riff-raff/blob/d8a011784c7d73d759bb35c9d255e6cc9267e77e/src/config.ts) will only look at `app` if `projectName` is not set.

akash1810: It looks like we've not updated the tag on actions-riff-raff since https://github.com/guardian/actions-riff-raff/pull/20 was merged. That would have helped here! Will update it now.

akash1810: Tag updated. We should see a [summary](https://github.com/guardian/actions-riff-raff/pull/20#issuecomment-1310876567) in the GHA build from now on 🤞🏽

PR #65, opened 2022-11-25
Updates the project name to "service catalogue"

kenoir:
## What does this change?

This change https://github.com/guardian/service-catalogue/pull/64 doesn't really update the name riff-raff uses, we need to 

## Why?

To _really_ reduce confusion when the project has a different name than that deployed under.


akash1810 (approved):
From https://github.com/guardian/service-catalogue/actions/runs/3548528033
image

img

PR #66, opened 2022-11-25
Search for repo by names reinstated

JuliaBrigitte:
## What does this change?
Add an optional search after endpoint repos
search by name with ?name=searchString

Thanks to Kenny for pointing out that this actually did not make it into main at all.

### DIFF
--- a/packages/github-lens-api/src/app.ts
+++ b/packages/github-lens-api/src/app.ts
@@ -99,7 +99,7 @@ export function buildApp(
 				break;
 			case '/repos':
 				info =
-					'Show all repos, with their team owners, optional search by name with ?name=searchString';
+					'Show all repos, with their team owners, optionally search by name with ?name=^repo-name-regex.*';
 				break;
 			case '/repos/:name':
 				info = 'Show repo and its team owners, if it exists';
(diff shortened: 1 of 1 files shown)


packages/github-lens-api/src/app.ts
```diff
@@ -95,7 +99,7 @@ export function buildApp(
 				break;
 			case '/repos':
 				info =
-					'Show all repos, with their team owners';
+					'Show all repos, with their team owners, optional search by name with ?name=searchString';
```
kenoir, line 102:
```suggestion
					'Show all repos, with their team owners, optional search by name with ?name=repo-name';
```

kenoir (approved):
Great! Would be nice in the future to do this on all endpoints in a standardised way (search in named field), but that needs a lot more thinking about.

packages/github-lens-api/src/app.ts
```diff
@@ -32,13 +32,17 @@ export function buildApp(
...
 				);
-				res.status(200).json(jsonResponse);
-			} else {
+				if (jsonResponse.length !== 0) {
+					res.status(200).json(jsonResponse);
+				} else {
+					res
+						.status(200)
+						.json({ searchString: searchString, info: 'no results found in repos' });
+				}			} else {
```
akash1810, line 45: let's add the eslint/prettier the the build to avoid this sort of formatting, which is a little difficult to immediately parse.

PR #68, opened 2022-11-28
Return objects not lists on /thing/:identifier endpoints

kenoir:
## What does this change?

Currently if you look for `/team/content-platforms` (or any member or repo) you get a list rather than an object, this change returns a less surprising object.

## Why?

Less surprising result format in API.


kenoir:
Closing as too far out of sync (this should just be recreated).

PR #69, opened 2022-11-28
Refactor to make routes in app.ts easier to grasp

JuliaBrigitte:
## What does this change?
Refactor to separate logic out of app.ts into controllers

## Why?
Keeping the routes separate for the implementation and the implementation in logical groups makes the code easier to understand und is good practice

## Further improvement
Maybe the calls to get the data could be moved too, but I ran into problems there and feel this refactor already adds worthwhile clarity.

### DIFF
--- a/packages/github-lens-api/src/controller/DescribeApiController.ts
+++ b/packages/github-lens-api/src/controller/DescribeApiController.ts
@@ -1,0 +1,63 @@
+import {getDescribeRouterHandler} from "common/expressRoutes";
+import type { Router } from 'express';
+import type express from 'express';
+
+interface InnerLayer {
+    method: string;
+}
+
+interface InnerRoute {
+    path: string;
+    stack: InnerLayer[];
+}
+
+interface RouterLayer {
+    route: InnerRoute | undefined;
+}
+
+function describeRoutes(path: string) {
+    let info = '';
+    switch (path) {
+        case '/healthcheck':
+            info = 'Display healthcheck';
+            break;
+        case '/repos':
+            info =
+                'Show all repos, with their team owners, optionally search by name with ?name=^repo-name-regex.*';
+            break;
+        case '/repos/:name':
+            info = 'Show repo and its team owners, if it exists';
+            break;
+        case '/archivedRepos':
+            info =
+                'Show all archived repos';
+            break;
+        case '/archivedReposNamesOnly':
+            info =
+                'Show all archived repos, names only (for tracker)';
+            break;
+        case '/teams':
+            info =
+                'Show all teams, with the repositories they own';
+            break;
+        case '/teams/:slug':
+            info = 'Show team and the repositories it owns, if it exists';
+            break;
+        case '/members':
+            info =
+                'Show member, with the teams they are in';
+            break;
+        case '/members/:login':
+            info = 'Show member and the teams they are in, if it exists';
+            break;
+        default:
+            info = 'No path info supplied';
+            break;
+    }
+    return info;
+}
+
+export const getRouteDescriptions = (router: Router) => {
+    return getDescribeRouterHandler(router, describeRoutes)
+};
+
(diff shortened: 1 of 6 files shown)


packages/github-lens-api/src/controller/DescribeApiController.ts
```diff
@@ -0,0 +1,63 @@
...
+interface InnerRoute {
+    path: string;
+    stack: InnerLayer[];
+}
+
+interface RouterLayer {
+    route: InnerRoute | undefined;
+}
+
+function describeRoutes(path: string) {
```
nicl, line 18: Not specific to this PR at all, but more a note to myself to refactor this. It would be perhaps better to do this kind of thing as follows:

```
    const defaultDescription = 'No path info supplied';
	
    const paths: Record<string, string | undefined> = {
		'/healtcheck': 'Display healthcheck',
		// others here
	};

	return paths[path] ?? defaultDescription;
```

And call the method `routeDescription` instead perhaps.

packages/github-lens-api/src/controller/DescribeApiController.ts
```diff
@@ -0,0 +1,63 @@
...
+interface InnerLayer {
+    method: string;
+}
+
+interface InnerRoute {
+    path: string;
+    stack: InnerLayer[];
+}
+
+interface RouterLayer {
```
nicl, line 14: These are unused and can be deleted.

packages/github-lens-api/src/controller/TeamsController.ts
```diff
@@ -0,0 +1,20 @@
...
+export const getTeamBySlug = (req: express.Request, res: express.Response, teamsData: RetrievedObject<Team[]>) => {
+    const jsonResponse = teamsData.payload.filter(
+        (item) => item.slug === req.params.slug,
+    );
+    if (jsonResponse.length !== 0) {
+        res.status(200).json(jsonResponse);
+    } else {
+        res
+            .status(200)
+            .json({ teamSlug: req.params.slug, info: 'Team not found' });
```
nicl, line 18: Should we return a 404 in these cases?

packages/github-lens-api/src/controller/ReposController.ts
```diff
@@ -0,0 +1,33 @@
...
+export const getRepoByName = (req: express.Request, res: express.Response, reposData: RetrievedObject<Repository[]>) => {
+    const jsonResponse = reposData.payload.filter(
+        (item) => item.name === req.params.name,
+    );
+    if (jsonResponse.length !== 0) {
+        return res.status(200).json(jsonResponse);
+    } else {
+        return res
+            .status(200)
+            .json({ repoName: req.params.name, info: 'Repository not found' });
```
nicl, line 31: 404?

packages/github-lens-api/src/controller/MembersController.ts
```diff
@@ -0,0 +1,20 @@
+import type {RetrievedObject} from "common/aws/s3";
+import type {Member, Repository} from "common/model/github";
+import type express from "express";
+
+export const getAllMembers = (req: express.Request, res: express.Response, membersData: RetrievedObject<Member[]>) => {
```
nicl, line 5: Minor, but the `All` in these names is perhaps redundant - `getTeams` is clear enough here.

nicl (approved):
Some comments but no blockers. Though I recommend setting up auto-formatting in your IDE.

One final bit of (more subjective) feedback: I'm not so keen on introducing lots of files - specifically thinking of the file-per-controller introduced here. In some language communities it is favoured, but in languages such as TS, Scala, and Go, where multiple classes can coexist in a file, I tend to bias to fewer files provided the content is still related. As IDEs/editors are still file-based for the most part, and lots of files with only a few lines of content is harder to comprehend/scan quickly.

packages/github-lens-api/src/controller/DescribeApiController.ts
JuliaBrigitte: I have deleted them.

packages/github-lens-api/src/controller/ReposController.ts
JuliaBrigitte: Sure (to all 404)

PR #70, opened 2022-11-28
Add try catch for http://localhost:3233/teams/:id

JuliaBrigitte:
## What does this change?
Just added error handling for http://localhost:3233/teams/:id

## Why?
nicer that just seeing an exception

### DIFF
--- a/packages/services-api/src/app.ts
+++ b/packages/services-api/src/app.ts
@@ -68,11 +68,17 @@ export function buildApp(config: Config): Express {
 	router.get(
 		'/teams/:id',
 		asyncHandler(async (req: express.Request, res: express.Response) => {
-			const team = await galaxiesApi.getTeam(req.params.id);
-			const services = await servicesApi.forGithubOwner(team.primaryGithubTeam);
-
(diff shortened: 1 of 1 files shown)


packages/services-api/src/app.ts
```diff
@@ -68,11 +68,17 @@ export function buildApp(config: Config): Express {
 	router.get(
 		'/teams/:id',
 		asyncHandler(async (req: express.Request, res: express.Response) => {
-			const team = await galaxiesApi.getTeam(req.params.id);
-			const services = await servicesApi.forGithubOwner(team.primaryGithubTeam);
-
-			const resp: TeamResponse = { ...team, services };
-			res.status(200).json(resp);
+			try
```
kenoir, line 71: I'm not sure try/catch is the best way to handle this. Instead `getTeam` should return `undefined` if you try to find a team that doesn't exist, and then you can check for that here instead of catching all exceptions (which might not always originate from a team not found).

kenoir (changes requested):

PR #71, opened 2022-11-28
Add archived repos endpoints

JuliaBrigitte:
## What does this change?
Add archived repos endpoints

## Why?
Will used them when moving tracker to use service catalogue

## Further work
Might change to other path so not to have as many endpoints.
Need to crosscheck with the results by slightly different call in tracker that leads less results.


packages/github-lens-api/src/controller/ReposController.ts
```diff
@@ -31,3 +31,30 @@ export const getRepoByName = (req: express.Request, res: express.Response, repos
...
+export const getArchivedRepos = (req: express.Request, res: express.Response, reposData: RetrievedObject<Repository[]>) => {
+    const jsonResponse = reposData.payload.filter(
+        (item) => item.archived === true,
+    );
+    if (jsonResponse.length !== 0) {
+        return res.status(200).json(jsonResponse);
+    } else {
+        return res
+            .status(200)
+            .json({ repoName: req.params.name, info: 'no repos found' });
```
akash1810, line 44: Could the shape of the response be the same on the happy and unhappy path? That is, on the unhappy path, could we respond with the empty array? This format will make it easier for consumers as they can have fewer branches.

packages/github-lens-api/src/app.ts
```diff
@@ -35,22 +45,35 @@ export function buildApp(
...
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const reposData=await getReposData(res, ghData);
+			if (reposData !== undefined) getArchivedRepos(req, res, reposData);
 		}),
 	);
 
+	router.get(
+		'/archivedReposNamesOnly',
+		asyncHandler(async (req: express.Request, res: express.Response) => {
+			const reposData=await getReposData(res, ghData);
```
kenoir, line 71: I think you need to merge `main` and utilise the formatting changes from here: https://github.com/guardian/service-catalogue/pull/83

packages/github-lens-api/src/app.ts
```diff
@@ -35,22 +45,35 @@ export function buildApp(
...
 	router.get(
 		'/repos/:name',
 		asyncHandler(async (req: express.Request, res: express.Response) => {
-			const reposData = (await ghData).repos;
-			if (reposData === undefined) {
-				res.status(500).json({ error: 'Unable to retrieve repository data!'});
-				return;
-			}
-				getRepoByName(req, res, reposData);
+			const reposData=await getReposData(res, ghData);
```
kenoir, line 55: I am not a fan of the way the files in controllers are being used here. My preference would be to (a) do all mutation of `Response` in `app.ts` and push any other logic into other files or (b) push the whole handler for the routes into the controller.

I find deciding on the response partly here and partly in the controller files is quite confusing.

JuliaBrigitte:
Sorry, made a bit of a mess of the rebase and had to force push my branch.

packages/github-lens-api/src/controller/DescribeApiController.ts
```diff
@@ -0,0 +1,38 @@
+import { getDescribeRouterHandler } from 'common/expressRoutes';
+import type { Router } from 'express';
+
+function describeRoutes(path: string) {
```
nicl, line 4: A map would probably be easier to read here. E.g.

```
{
  '/healthcheck': 'Display healthcheck',
  '/repos/:name': 'Show all...'
}
```

nicl (approved):
LGTM! 🚢

PR #83, opened 2022-11-29
chore(prettier): Configure Prettier

akash1810:
## What does this change?
Since @guardian/eslint-config v2.0.0 we have to run Prettier ourselves.

In this change, we configure ESlint to run Prettier.

See:
  - https://github.com/guardian/csnx/releases/tag/%40guardian%2Feslint-config%402.0.0
  - https://github.com/prettier/eslint-plugin-prettier

## Why?
Prettier is an opinionated formatter... so we don't have to be! It produces a consistent code style.

It takes care of decisions on quite mark style (`"` or `'`), tabs vs spaces, trailing commas on objects, etc.


kenoir (approved):
Nice one.

PR #89, opened 2022-12-01
chore: Use departmental tsconfig

akash1810:
## What does this change?
Use [@guardian/tsconfig](https://github.com/guardian/csnx/tree/main/libs/%40guardian/tsconfig) to reduce the amount of configuration.

This config has highlighted some code paths not returning expected values. These are also fixed.

## Why?
- Less config
- An easier path to ESM... I think!
- Makes it easier to use packages that are now published as ESM only (e.g. [node-fetch](https://github.com/guardian/service-catalogue/pull/87#discussion_r1036853767))

### DIFF
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -1,8 +1,8 @@
-import 'source-map-support/register';
-import { GuRootExperimental } from '@guardian/cdk/lib/experimental/constructs/root';
-import { CloudFormationLens } from '../lib/cloudformation-lens';
-import { GithubLens } from '../lib/github-lens';
-import { ServicesApi } from '../lib/services-api';
+import 'source-map-support/register.js';
+import { GuRootExperimental } from '@guardian/cdk/lib/experimental/constructs/root.js';
+import { CloudFormationLens } from '../lib/cloudformation-lens.js';
+import { GithubLens } from '../lib/github-lens.js';
+import { ServicesApi } from '../lib/services-api.js';
 
 const app = new GuRootExperimental();
 
--- a/packages/common/src/aws/kms.test.ts
+++ b/packages/common/src/aws/kms.test.ts
@@ -22,14 +22,4 @@ describe('decrypt', function () {
 
 		expect(actualOutput).toBe(expectedOutput);
 	});
-
-	it('returns undefined if the decrypt operation fails', async function () {
-		const expectedOutput = 'plaintext';
-
-		kmsMock.on(DecryptCommand).rejects();
-
-		const actualOutput = await decrypt('keyId', expectedOutput);
-
-		expect(actualOutput).toBe(undefined);
-	});
 });
--- a/packages/common/src/aws/kms.ts
+++ b/packages/common/src/aws/kms.ts
@@ -8,7 +8,7 @@ const kmsClient = new KMSClient({
 export const decrypt = async (
 	cipherText: string,
 	keyId: string,
-): Promise<string | undefined> => {
+): Promise<string> => {
 	const decryptCommand = new DecryptCommand({
 		CiphertextBlob: Buffer.from(cipherText, 'base64'),
 		KeyId: keyId,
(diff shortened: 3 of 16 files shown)


packages/common/src/aws/kms.test.ts
```diff
@@ -22,14 +22,4 @@ describe('decrypt', function () {
 
 		expect(actualOutput).toBe(expectedOutput);
 	});
-
-	it('returns undefined if the decrypt operation fails', async function () {
```
akash1810, line 26: This is testing an oddity of AWS's types, the code path should never happen, so removed the test.

packages/common/src/aws/kms.ts
```diff
@@ -8,7 +8,7 @@ const kmsClient = new KMSClient({
 export const decrypt = async (
 	cipherText: string,
 	keyId: string,
-): Promise<string | undefined> => {
+): Promise<string> => {
```
akash1810, line 11: This type is easier for callers to handle; the `undefined` case should never happen (see added comment below).

packages/common/src/github/github.ts
```diff
@@ -69,6 +69,8 @@ export const getOctokit = (config: GitHubConfig): Octokit => {
 					console.log(`Retrying after ${retryAfter} seconds!`);
 					return true;
 				}
+
+				return false;
```
akash1810, line 73: Ensure all code paths return a value (same for below).

packages/github-lens-api/src/controller/repos.ts
```diff
@@ -7,22 +7,22 @@ export const getAllRepos = (
...
-		if (jsonResponse.length !== 0) {
-			res.status(200).json(jsonResponse);
-		} else {
-			return res.status(200).json({
-				searchString: searchString,
-				info: 'no results found in repos',
-			});
-		}
-	} else {
+	if (typeof req.query.name === 'undefined') {
```
akash1810, line 10: Returning early is easier to read.

packages/services-api/src/app.ts
```diff
@@ -67,13 +67,17 @@ export function buildApp(config: Config): Express {
 
 	router.get(
 		'/teams/:id',
-		asyncHandler(async (req: express.Request, res: express.Response) => {
-			const team = await galaxiesApi.getTeam(req.params.id);
-			const services = await servicesApi.forGithubOwner(team.primaryGithubTeam);
+		asyncHandler(
+			async (req: express.Request<{ id: string }>, res: express.Response) => {
```
akash1810, line 71: Type annotation added to `req`, to ensure `req.params.id` is type-safe.

packages/services-api/src/galaxies.ts
```diff
@@ -89,12 +89,14 @@ export class S3GalaxiesApi implements GalaxiesApi {
 
 		const teamsMapping = mappingResp.payload;
 
-		// Add the (for now) missing fields.
-		return Object.keys(teams).map((teamId) => {
+		return Object.entries(teams).map(([id, team]) => {
```
akash1810, line 92: `tsc` wasn't convinced `teams[teamId]` wasn't undefined when using `Object.keys`, it's happier with `Object.entries` 🤷🏽 .

packages/common/src/aws/kms.ts
```diff
@@ -18,14 +18,18 @@ export const decrypt = async (
...
+			return message;
 		}
+
+		console.log('Ciphertext successfully decrypted');
+		return Buffer.from(Plaintext).toString();
 	} catch (e) {
-		console.error(`Decryption failed: ${(e as Error).message}`);
+		const message = `Decryption failed: ${(e as Error).message}`;
+		console.error(message);
+		throw new Error(message);
```
akash1810, line 33: If we get here, we should fail fast, so `throw` (returning `undefined` requires extra logic by the caller).

packages/common/src/aws/kms.ts
```diff
@@ -18,14 +18,18 @@ export const decrypt = async (
 		const { Plaintext } = await kmsClient.send(decryptCommand);
 		console.log('Decrypt command sent successfully');
 
-		if (Plaintext) {
-			console.log('Ciphertext successfully decrypted');
-			const plaintextString = Buffer.from(Plaintext).toString();
-			return plaintextString;
-		} else {
-			console.error('Plaintext is missing from DecryptCommandOutput!');
+		if (!Plaintext) {
```
kenoir, line 21: 👌 nice one -  I wondered why I was forced to deal with this case by the type system!

kenoir, line 21: 👌 nice one -  I wondered why I was forced to deal with this case by the type system!

packages/cdk/bin/cdk.ts
```diff
@@ -1,8 +1,8 @@
-import 'source-map-support/register';
-import { GuRootExperimental } from '@guardian/cdk/lib/experimental/constructs/root';
-import { CloudFormationLens } from '../lib/cloudformation-lens';
-import { GithubLens } from '../lib/github-lens';
-import { ServicesApi } from '../lib/services-api';
+import 'source-map-support/register.js';
+import { GuRootExperimental } from '@guardian/cdk/lib/experimental/constructs/root.js';
```
akash1810, line 2: I _think_ once GuCDK is published as ESM, we won't need these explicit extensions.

kenoir: Oh wowza, this is quite unpleasant.

kenoir (approved):

packages/cdk/bin/cdk.ts
akash1810: > Oh wowza, this is quite unpleasant.

Indeed!

akash1810:
Deployed, and everything seem to be running still. Merging.

PR #90, opened 2022-12-01
Implementation and testing of initial repocop rules

kenoir:
## What does this change?

This PR implements a repository metadata object, and a flexible way to implement arbitrary rules based on the data provided.

It also autogenerates a markdown document for a human readable way to look at the rules, quick remediations and justifications for them. There is a handler for running this process, and a pre-commit hook that runs a snapshot test that notifies the user if the markdown needs to be regenerated.

## What this PR does not include

- Serialisation/deserialisation logic for repository objects returned from github lens
- Infrastructure/a specific execution mechanism (ie lambda, scheduled container, EC2)
- Data storage
- Notification/action of the back of rules
- CI

### Why repocop

DevX has ways of checking whether repositories adhere to particular practices. This logic all exists in tracker, which means we de facto rely on devX to communicate violations of best practice to teams manually, and we don't have a good way of deciding the appropriate cadence for notifying people. Repocop will notify teams within a day or so of a repository being changed as it runs overnight. This means

1.  Teams do have time to make sure repos are set up properly as repocop doesn't run to the next day.
2. Significant performance benefits/runtime reduction as 90% of repos have not been modified in the last 3 months.
3. Frequently updated repositories are more likely to be business critical, and they will be evaluated more often

All repositories will be evaluated during the initial invocation.

### Why scala 3

This is already a multilanguage repo. It's also important that DevX are comfortable using languages/frameworks that are commonly used across the department. Scala 3 is one of those, and we don't have a project that uses that currently. This project in particular makes use of Scala 3's new enum construct to enumerate repository rules


packages/repocop/src/main/scala/Repository.scala
```diff
@@ -0,0 +1,33 @@
+package com.gu.repocop
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter.ISO_DATE_TIME
+import java.time.temporal.ChronoUnit
+import scala.util.{Failure, Success, Try}
+
+case class Repository(
+                       name: String,
+                       Private: Boolean,
```
akash1810, line 10: Why is this upper cased?

packages/repocop/src/main/scala/Rules.scala
```diff
@@ -0,0 +1,49 @@
...
+    case hasValidTopic extends RepoRule(
+      statement = repoContainsProductionStatus,
+      violationMessage = "Valid production status topic required",
+      ruleJustification = singleLineStrippedString(
+        """Currently, we mandate that production P+E repos need to be reliably integrated with Snyk. In order to work
+          | out which projects contain production code, we ask that all choose from a list of status tags"""
+      )
+    )
+
+    case defaultIsMain extends RepoRule(
```
akash1810, line 29: `defaultBranchName` might be a little more descriptive here (and in the markdown table).

packages/repocop/src/main/scala/Rules.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    case hasOwner extends RepoRule(
+      statement = _.owners.nonEmpty,
+      violationMessage = "Requires a valid team set as an owner",
+      ruleJustification = singleLineStrippedString(
+        """Repositories are generally required to have a team owner for smoother handover in case someone leaves or
+          | moves to another team. When somebody leaves the org, any individual access they had to a repository is
+          |maintained, so there is a security case for this rule as well as a maintainability one."""
+      )
+    )
```
akash1810, line 18: Should we be more assertive by removing 'generally'?

packages/repocop/src/main/scala/Rules.scala
```diff
@@ -0,0 +1,49 @@
+package com.gu.repocop
+
+object Rules {
+
+  def singleLineStrippedString(s: String): String = s.stripMargin.replace("\n", "")
+
+  enum RepoRule(statement: Repository => Boolean, val violationMessage: String, val ruleJustification: String = ""):
```
akash1810, line 7: Should we also add an (optional?) remediation? For the topic rule, for example, we don't currently enumerate the set of topics we're looking for, so it'll be difficult for someone to fix their repo.

packages/repocop/src/main/scala/Repository.scala
```diff
@@ -0,0 +1,33 @@
+package com.gu.repocop
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter.ISO_DATE_TIME
+import java.time.temporal.ChronoUnit
+import scala.util.{Failure, Success, Try}
+
+case class Repository(
```
akash1810, line 8: Worth adding a comment to say this should be kept in sync with the TypeScript model?

packages/repocop/src/main/scala/main.scala
```diff
@@ -0,0 +1,24 @@
...
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter.ISO_DATE_TIME
+import com.gu.repocop.MarkdownHelpers.createPage
+
+import java.io.*
+@main
+def main(): Unit = {
+
+  //need to filter out archived repos immediately
+  println("hi")
```
akash1810, line 14: 👋🏽 

What should this do?

akash1810 (commented):
Looks great! Do we want to build this project in CI?

packages/repocop/src/main/scala/Repository.scala
NovemberTang: lower case private is a reserved keyword and keeping the field names consistent makes serialisation 100x easier as you can derive the transformation automatically

packages/repocop/src/main/scala/Rules.scala
NovemberTang: enumerating topics is coming in the next PR!

packages/repocop/src/main/scala/main.scala
NovemberTang: nothing at the moment. to keep this PR manageable i've just created the first few rules and tested them. Eventually this will contain the request, deserialisation and processing, but that's a separate concern and will be built up over the course of the next week or so

packages/repocop/src/main/scala/Repository.scala
akash1810: We could use backtick escaping?

```scala
case class Repository(
  `private`: Boolean
)
```

NovemberTang: yea that works too

akash1810 (commented):
Looks great!

I think we should get this building in CI soon, but happy for that to be separate. Do we also need to update the Snyk workflow to have it pick up the `buiild.sbt`?

akash1810 (approved):
Looks great!

I think we should get this building in CI soon, but happy for that to be separate. Do we also need to update the Snyk workflow to have it pick up the `buiild.sbt`?

PR #91, opened 2022-12-01
Report build tool

nicl:
## What does this change?

Includes the build tool if found. See also:

* https://github.com/guardian/riff-raff/pull/937
* https://github.com/guardian/actions-riff-raff/pull/23

## Why?

To understand usage of `guardian/actions-riff-raff` and other build tools.

### DIFF
--- a/packages/cloudformation-lens/account/account.go
+++ b/packages/cloudformation-lens/account/account.go
@@ -19,6 +19,7 @@ import (
 type Features struct {
 	GuardianCDKVersion    *string  `json:"guardianCdkVersion"`
 	GuardianDNSRecordSets []string `json:"guardianDnsRecordSets"`
+	GuardianBuildTool     *string  `json:"guardianBuildTool"`
 }
 
 type Stack struct {
@@ -137,6 +138,7 @@ func getStacks(ctx context.Context, client *cloudformation.Client, cache cache.C
 			DevxFeatures: Features{
 				GuardianCDKVersion:    guardianCDKVersion(tags, metadataMap),
 				GuardianDNSRecordSets: domains,
+				GuardianBuildTool:     guardianBuildTool(tags),
 			},
 		}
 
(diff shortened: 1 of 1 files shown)


kenoir (approved):

akash1810:
I'm not sure this is a property of a service, but of a repository. That is, rather than reading a CFN tag, I think this should be exposed as a property in Lens, with data-fetcher parsing the file content of a repository.

nicl:
@akash1810 There are tradeoffs each way here:

1. We look at build tools per repo
How do we measure usage accurately here for monorepos? Or for repos that are part of builds in another repo but don't have one themself (such as Janus)?

2. We look at build tools per service
What about usage for libraries etc? (Also an interesting question).

I think ultimately we'll want a sense of both, but services.gutools.co.uk is (as the name implies!) primary focused on running services so we should, where possible, prefer runtime data.

akash1810:
> I think ultimately we'll want a sense of both, but services.gutools.co.uk is (as the name implies!) primary focused on running services so we should, where possible, prefer runtime data.

Pedantic, but this is a change to CFN-Lens. I'd also suggest runtime and build time have separate concerns. Might we want to formalise the boundaries of data for the various APIs that make up the Service Catalogue? (Especially as we're adding a couple more soon for RepoCop).

The aim of the PR is described as:

> To understand usage of `guardian/actions-riff-raff` and other build tools.

I'm not convinced a CFN tag will provide the meaningful coverage as we know there's a significant number of stacks that are _not_ deployed via Riff-Raff.

A [GitHub search](https://github.com/search?q=org%3Aguardian+actions-riff-raff+language%3AYAML+path%3A%2F%5E.github%5C%2Fworkflows%5C%2F%2F&type=code&l=.github%2Fworkflows%2F) would be more accurate for `guardian/actions-riff-raff`. Need to think the other build tools...

PR #97, opened 2022-12-02
Adds CDK described infra for repocop

JuliaBrigitte:
## What does this change?
Start to use cdk, implements scheduled lambda that uses starter application

## Why?
Repocop did not yet use cdk


akash1810 (approved):
👍🏽 

We should add a [test](https://github.com/guardian/service-catalogue/blob/main/packages/cdk/lib/services-api.test.ts) for the CDK stack too.

kenoir:
Testing deployment: https://riffraff.gutools.co.uk/deployment/view/6a931e4c-acd5-4040-aab6-4b331dfd6520

PR #98, opened 2022-12-05
Add Repocop CDK test

kenoir:
## What does this change?

Follows https://github.com/guardian/service-catalogue/pull/97, to add test coverage for Repocop CDK stack.

## Why?

So we notice and test infra changes in repocop.

### DIFF
--- a/packages/cdk/lib/repocop.test.ts
+++ b/packages/cdk/lib/repocop.test.ts
@@ -1,0 +1,15 @@
+import { App } from 'aws-cdk-lib';
+import { Template } from 'aws-cdk-lib/assertions';
+import { Repocop } from './repocop';
+
+describe('The Repocop stack', () => {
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/repocop.test.ts
```diff
@@ -0,0 +1,15 @@
+import { App } from 'aws-cdk-lib';
+import { Template } from 'aws-cdk-lib/assertions';
+import { Repocop } from './repocop';
+
+describe('The Repocop stack', () => {
+	it('matches the snapshot', () => {
+		const app = new App();
+		const stack = new Repocop(app, 'GithubLens', {
```
akash1810, line 8:
```suggestion
		const stack = new Repocop(app, 'Repocop', {
```

akash1810 (approved):

PR #102, opened 2022-12-05
Simplify .gitignore by excluding .idea (Intellij)

JuliaBrigitte:
## What does this change?
Exclude more files, that are not intended to be checked in

## Why?
Akash suggested I do this to reduce the noise

## Please comment!
I followed the lead of what was already thereand added files to  .gitignore, but am not quite sure why we do not just put
.idea/*
in the .gitignore. Please do comment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,63 +28,21 @@ go.work
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+.idea
+
 # Project-specific stuff
-packages/repocop/.idea/*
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -28,7 +28,20 @@ go.work
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# Project-specific stuff
+packages/repocop/.idea/*
```
akash1810, line 32: We can simplify this quite a lot with just:

```
.idea
.bsp
```

Removing the path specific rules. I don't think we care about any files in these directories. Some files were [specifically added](https://github.com/guardian/service-catalogue/pull/85), but for the most part, I think we can ignore them.

PR #113, opened 2022-12-08
Consistent api endpoint output for github lens api

JuliaBrigitte:
## What does this change?

Make the github lens api endpoints more consistent

## Why?

To make user experience less confusing

### DIFF
--- a/packages/github-lens-api/src/controller/members.ts
+++ b/packages/github-lens-api/src/controller/members.ts
@@ -15,11 +15,11 @@ export const getMembersByLogin = (
 	res: express.Response,
 	membersData: RetrievedObject<Member[]>,
 ) => {
-	const member = membersData.payload.filter(
+	const member = membersData.payload.find(
 		(item) => item.login === req.params.login,
 	);
-	if (member.length !== 0) {
-		res.status(200).json({ ...membersData, payload: member[0] });
+	if (member) {
+		res.status(200).json({ ...membersData, payload: member });
 	} else {
 		res
 			.status(404)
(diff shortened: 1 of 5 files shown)


packages/github-lens-api/src/controller/members.ts
```diff
@@ -15,11 +15,11 @@ export const getMembersByLogin = (
 	res: express.Response,
 	membersData: RetrievedObject<Member[]>,
 ) => {
-	const jsonResponse = membersData.payload.filter(
+	const member = membersData.payload.filter(
```
kenoir, line 18: I think `find` would make more sense here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

packages/github-lens-api/src/controller/teams.ts
```diff
@@ -15,11 +15,11 @@ export const getTeamBySlug = (
 	res: express.Response,
 	teamsData: RetrievedObject<Team[]>,
 ) => {
-	const jsonResponse = teamsData.payload.filter(
+	const team = teamsData.payload.filter(
```
kenoir, line 18: See above suggestion to use `find`.

packages/github-lens-api/src/controller/repos.ts
```diff
@@ -24,15 +24,15 @@ export const getAllRepos = (
 		},
 	];
 
-	const repos = reposData.payload.filter((repo) => {
+	const repo = reposData.payload.filter((repo) => {
```
kenoir, line 27: Nitpick: this makes sense to be plural `repos` as we expect multiple results.

PR #115, opened 2022-12-09
Use map for api description

JuliaBrigitte:
## What does this change?
Use map (instead of case) for api description as in github lens api

## Why?
To make codebase more consistent


packages/services-api/src/controller/describe.ts
```diff
@@ -0,0 +1,18 @@
+import { getDescribeRouterHandler } from 'common/expressRoutes';
+import type { Router } from 'express';
+
+function describeRoutes(path: string): string {
+	const infoByRoute: Record<string, string> = {
+		'/healthcheck': 'Display healthcheck',
+		'/teams': 'Show all P&E teams, with the services they own',
+		'/teams/:id': 'Show team and the services it owns, if it exists',
```
NovemberTang, line 8: why are we switching to id? from a developer/product point of view, if we only support one of these, i'm more likely to remember the name rather than the numerical id?

JuliaBrigitte: Not sure, I just saw that it was :id and adjusted the description, it did not work anyway when I tried today, will ask Nic Monday.

packages/services-api/src/app.ts
```diff
@@ -4,8 +4,9 @@ import cors from 'cors';
 import type { Express } from 'express';
 import express, { Router } from 'express';
 import asyncHandler from 'express-async-handler';
-import { getDescribeRouterHandler } from '../../common/src/expressRoutes';
+import { getHealthCheckHandler } from 'github-lens-api/src/controller/healthcheck';
```
akash1810, line 7: Does this import work? It implies `services-api` knows about `github-lens-api`; it's [tsconfig](https://github.com/guardian/service-catalogue/blob/7657b78e696ebaf6ec2b0297871a44496dc3e857/packages/services-api/tsconfig.json#L5) suggests it doesn't.

packages/services-api/src/controller/describe.ts
JuliaBrigitte: See https://services.gutools.co.uk/
"http://services.gutools.co.uk/teams/:id"
So this was the case before.

nicl (approved):
Looks good!

I wonder if we can give a more descriptive name to `.../describe.ts`? I.e something that says what it describes - normally files would have a noun name rather than verb.

An alternative is to simply move the map back into `app.ts`. It's not really a 'controller' in this case, so that might be more appropriate anyway.

(Neither are blockers though.)

kenoir (approved):

PR #116, opened 2022-12-13
Update filters to work more generally on API endpoints

kenoir:
## What does this change?

Updates the way filters work so they can be applied across multiple endpoints. This change allows filtering of repository data on the following endpoints:

- `/teams/:slug`
- `/teams`

It renames the filter parameters to refer to specifically the type of data they are filtering on the API responses, i.e.

- `name` -> `repoName`
- `isArchived` -> `repoIsArchived`

It also adds the following filter that can be applied across the existing endpoints;

- `repoIsNotArchived`: Display only repositories not archived.
- `repoNotOwned`: Display only repositories with no owners.

## Why?

Generalises the use of filters in a (hopefully) sensible way, and provides utility to other services wanting to consume this data.

### DIFF
--- a/packages/github-lens-api/src/app.ts
+++ b/packages/github-lens-api/src/app.ts
@@ -7,7 +7,7 @@ import { getRouteDescriptions } from './controller/describe';
 import { getHealthCheckHandler } from './controller/healthcheck';
 import { getMembers, getMembersByLogin } from './controller/members';
 import { getAllRepos, getRepoByName } from './controller/repos';
-import { getTeamBySlug } from './controller/teams';
+import { getAllTeams, getTeamBySlug } from './controller/teams';
 import type { GitHubData } from './data';
 
 export function buildApp(ghData: Promise<GitHubData>): Express {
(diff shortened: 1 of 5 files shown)


packages/github-lens-api/src/app.ts
```diff
@@ -57,7 +57,7 @@ export function buildApp(ghData: Promise<GitHubData>): Express {
 				res.status(500).json({ error: 'Unable to retrieve teams data!' });
 				return;
 			}
-			res.status(200).json(teamsData);
+			getAllTeams(req, res, teamsData);
```
kenoir, line 60: This change also adds a fix where the "controller" function `getAllTeams` was not called.

packages/github-lens-api/src/filters.ts
```diff
@@ -0,0 +1,52 @@
...
+	},
+	{
+		paramName: 'repoIsArchived',
+		fn: (repo: Repository) => repo.archived ?? false,
+	},
+	{
+		paramName: 'repoIsNotArchived',
+		fn: (repo: Repository) => !(repo.archived ?? false),
+	},
+];
```
akash1810, line 26: This seems a little verbose; a single query string of `?archived=<true|false>` would be preferred here, IMO.

I'd be surprised if there isn't an expressjs middleware (or similar) that provides type-safety to query strings too. That is, something that allows us to describe `archived` as a boolean type, giving us both variants.

If there isn't https://zod.dev/ should work, I think.

As discussed IRL, happy to get this in and refactor later.

akash1810 (approved):

PR #117, opened 2022-12-13
chore: Add Scalafmt

akash1810:
## What does this change?
Scalafmt is a code formatter for Scala. See https://scalameta.org/scalafmt/.

## Why?
Mainly because I can never remember how to get IntelliJ to not add a million indents when creating a case class!

```scala
case class Person(
                  name: String) {

}
```

vs

```scala
case class Person(
  name: String
) {

}
```


.scalafmt.conf
```diff
@@ -0,0 +1 @@
+packages/repocop/.scalafmt.conf
```
akash1810, line 1: This is a symlink for those who use open the project from the root.

kenoir (approved):
Neat.

PR #119, opened 2022-12-13
Adds team filters

kenoir:
## What does this change?

This change adds the ability to filter the teams endpoints for valid / engineering teams based on the `validGithubTeams.ts` data.

## Why?

To allow this data to be used by repocop.

### DIFF
--- a/packages/github-lens-api/src/validGithubTeams.ts
+++ b/packages/github-lens-api/src/validGithubTeams.ts
@@ -64,3 +64,14 @@ export const validGithubTeams = [
 	{ name: 'it-australia', engineering: false },
 	{ name: 'glabs-au', engineering: false },
 ];
+
+export const engineeringTeamSlugs = validGithubTeams
+	.filter((team) => team.engineering)
+	.map((team) => {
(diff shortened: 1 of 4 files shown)


packages/github-lens-api/src/validGithubTeams.ts
```diff
@@ -64,3 +64,14 @@ export const validGithubTeams = [
 	{ name: 'it-australia', engineering: false },
 	{ name: 'glabs-au', engineering: false },
 ];
+
+export const engineeringTeamSlugs = validGithubTeams
+	.filter((team) => team.engineering)
+	.map((team) => {
+		const { name } = team;
+		return name;
+	});
```
akash1810, line 73: minor: if you wanted to be more terse:

```ts
export const engineeringTeamSlugs = validGithubTeams
	.filter((team) => team.engineering)
	.map(({ name }) => name);
```

akash1810 (approved):
Should we add details of the new filters to the PR description too? Just so it's easy to discover in future.

PR #120, opened 2022-12-13
chore: Update to Node 18, latest LTS

akash1810:
> **Note**
> This requires a successful build of the new [arm64-focal-node18-devx recipe](https://amigo.gutools.co.uk/recipes/arm64-focal-node18-devx)

## What does this change?
In this change, we move to Node 18 for the data-fetcher lambda, the lens-api, and the services-api. The Node 18 lambda runtime comes with the AWS SDK v3, so they get moved to `devDependencies`.

## Why?
It's always good to use LTS versions.


akash1810:
I [deployed](https://riffraff.gutools.co.uk/deployment/view/8a493804-b5b0-4495-9e10-3add6accca87) this. The EC2 services were happy. The lambda, however, is now timing out:

> Task timed out after 300.11 seconds

PR #124, opened 2022-12-14
Add repocop/vulnerability api adr

NovemberTang:
## What does this change?

Moves the ADR from somewhere in the filesystem to sit alongside the actual code

## Why?

Findability, better visibility of changes


kenoir:
[Preview link](https://github.com/guardian/service-catalogue/blob/5884769e53017e2596435c9bd22af978e1546fb1/ADR/repocop-and-vulnerability-api-arch.md)

PR #125, opened 2022-12-14
Make filter params accept boolean values

kenoir:
## What does this change?

Allow some of the existing parameters to take a boolean value, so we can have fewer parameters and they are used in a less surprising way. The implementation naively looks for the addition of `=false` to the param, and uses only positive language so that the original param without the addition of boolean also works.

e.g.

- `/repos?repoIsArchived=false`: returns all repos where archived is false
- `/repos?repoIsArchived=true`: returns all repos where archived is true
- `/repos?repoIsArchived`: returns all repos where archived is true
- `/repos?repoIsArchived=purplehippos`: _also_ returns all repos where archived is true

Although this is _a bit weird_ I think it is good enough for our purposes.

## Why?

Slightly less surprised developers! 😮

### DIFF
--- a/packages/github-lens-api/src/filters.ts
+++ b/packages/github-lens-api/src/filters.ts
@@ -33,14 +33,14 @@ export const repoFilters: RepoFilter[] = [
 		paramName: 'repoIsOwned',
 		fn: (repo: Repository, paramValue: string) => {
 			const repoIsOwned = repo.owners.length !== 0;
-			return paramValue === 'true' ? repoIsOwned : !repoIsOwned;
+			return paramValue === 'false' ? !repoIsOwned : repoIsOwned;
 		},
 	},
 	{
 		paramName: 'repoIsArchived',
 		fn: (repo: Repository, paramValue: string) => {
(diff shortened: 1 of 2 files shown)


packages/github-lens-api/src/filters.ts
```diff
@@ -30,16 +30,18 @@ export const repoFilters: RepoFilter[] = [
 		fn: (repo: Repository, paramValue: string) => !!repo.name.match(paramValue),
 	},
 	{
-		paramName: 'repoNotOwned',
-		fn: (repo: Repository) => repo.owners.length === 0,
+		paramName: 'repoIsOwned',
+		fn: (repo: Repository, paramValue: string) => {
```
akash1810, line 34: Ideally, would `paramValue` be `boolean` here?

akash1810 (approved):
👍🏽

We might be able to further improve here, but auto-generating the strings in `describe.ts`, but not a blocker.

akash1810:
> We might be able to further improve here, but auto-generating the strings in `describe.ts`, but not a blocker.

https://github.com/guardian/service-catalogue/pull/128

PR #127, opened 2022-12-14
Ask developers to explain how changes were tested

NovemberTang:
## What does this change?
Adds a test section to the PR template

## Why?
It's not always clear from the code, and is extra helpful in multilingual repos where not everybody is familiar with all the test frameworks

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,5 @@
 ## What does this change?
 
 ## Why?
+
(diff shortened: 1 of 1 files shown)


kenoir:
> It's not always clear from the code, and is extra helpful in multilingual repos where not everybody is familiar with all the test frameworks

Or a useful reminder where things might not have any tests.

kenoir (approved):

PR #130, opened 2022-12-15
simplify hook setup

NovemberTang:
## What does this change?

extracts pre hook commit to its own file

## Why?

easier to read, easier to test

## How has it been verified?

deleted the hook from .git, reran the setup script, verified that the hook still works.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -1,12 +1,20 @@
 #!/bin/zsh
 
+MSG='
+RepoCop changes detected"
+
+Running markdown snapshot test. If this fails, regenerate the markdown file, stage the changes, and commit again
+Run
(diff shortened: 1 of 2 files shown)


kenoir (approved):

scripts/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/bin/zsh
+
+
+REPOCOP_CHANGES=$(git status --short packages/repocop/)
+if [[ -n "$REPOCOP_CHANGES" ]]
+then
+    printf "RepoCop changes detected"
+    printf "\n\nRunning markdown snapshot test. If this fails, regenerate the markdown file, stage the changes, and commit again"
```
kenoir, line 8: Can you add the command to regenerate here?

PR #131, opened 2022-12-20
Reduce requests made to github

NovemberTang:
## What does this change?

Affected GitHub requests, assuming ~3000 repos
Previously, we were making multiple requests for every repository to gather all the metadata fresh. Now, we run the initial requests to get the basic metadata, and only make further requests for information if we detect that the repo has been updated.

With a several day gap between successful runs, the calls to the language and commit endpoints have been reduced from about 3200 each to about 80 each, a reduction of about 98% 🥳🥳🥳. **n.b no changes have been made do how we interact with other endpoints. I'd guesstimate that it would be closer to 90% once all unchanged requests have been factored in**

## Why?

The data fetcher has been failing for the last few days as recent changes mean we are now exceeding the rate limit. This had not been detected previously as the local config runs only a subset of the repositories. Only making additional requests for a subset of repositories keeps us under the limit, and greatly reduces the running time of the lambda.

## How has it been verified?

Unit tests have been written to verify that timestamp matching works. Extra logging has been added to the handler so users can sense check the number of modified repos.

### DIFF
--- a/packages/common/src/github/github.ts
+++ b/packages/common/src/github/github.ts
@@ -221,17 +221,7 @@ export async function getLastCommitForRepositories(
 		repositories
 			.filter((repository) => {
 				const repositoryIsEmpty = repository.size === 0;
-				if (repositoryIsEmpty) {
-					console.log(
-						`Repository ${repository.name} is empty so there is also no last commit`,
-					);
-				}
 				const hasDefaultBranch = repository.default_branch !== undefined;
-				if (!hasDefaultBranch) {
-					console.log(
-						`Repository ${repository.name} has no default branch so there is also no last commit`,
-					);
-				}
 				return hasDefaultBranch && !repositoryIsEmpty;
 			})
 			.map(async ({ name, default_branch }) => {
--- a/packages/common/src/github/github.ts
+++ b/packages/common/src/github/github.ts
@@ -94,7 +94,7 @@ export const getOctokit = (config: GitHubConfig): Octokit => {
 	return _octokit;
 };
 
-export const listRepositories = async (
+export const getReposFromGitHub = async (
 	client: Octokit,
 ): Promise<RepositoriesResponse> => {
 	return await client.paginate(
--- a/packages/common/src/github/github.ts
+++ b/packages/common/src/github/github.ts
@@ -97,7 +97,7 @@ export const getOctokit = (config: GitHubConfig): Octokit => {
 export const getReposFromGitHub = async (
 	client: Octokit,
 ): Promise<RepositoriesResponse> => {
-	return await client.paginate(
+	const repos: RepositoriesResponse = await client.paginate(
 		client.repos.listForOrg,
 		{
 			org: 'guardian',
@@ -105,6 +105,8 @@ export const getReposFromGitHub = async (
 		},
 		(response) => response.data,
 	);
+	console.log(`Found ${repos.length} repositories on Github`);
+	return repos;
 };
 
 export const getTeam = async (
--- a/packages/github-data-fetcher/src/handler.ts
+++ b/packages/github-data-fetcher/src/handler.ts
@@ -17,7 +17,7 @@ import {
 import { configureLogging, getLogLevel } from 'common/log/log';
 import type { Commit, Member, Repository, Team } from 'common/model/github';
 import { getConfig } from './config';
-import { foundUnchangedMatchOnGithub } from './repoMatching';
+import { isCachedRepositoryStale } from './repoMatching';
 import { asMember, asRepo, getAdminReposFromResponse } from './transformations';
 
 // Returns a map of repoName -> admins (a list of team slugs).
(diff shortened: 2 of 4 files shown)


packages/common/src/github/github.ts
```diff
@@ -219,17 +221,7 @@ export async function getLastCommitForRepositories(
 		repositories
 			.filter((repository) => {
 				const repositoryIsEmpty = repository.size === 0;
-				if (repositoryIsEmpty) {
-					console.log(
-						`Repository ${repository.name} is empty so there is also no last commit`,
-					);
```
akash1810, line 225: For a future PR: these log lines provide an indication of repositories that should be deleted. Maybe these checks should be made in it's own discrete step? and/or encoded into RepoCop?

packages/github-data-fetcher/src/handler.ts
```diff
@@ -71,8 +74,20 @@ export interface TeamsAndRepositories {
 	members: Member[];
 }
 
+async function getReposFromS3(
+	S3client: S3Client,
```
akash1810, line 78: minor: could this be `s3Client` to keep variables to camelCase? (A little surprised the linter didn't flag this...)

packages/github-data-fetcher/src/handler.ts
```diff
@@ -82,71 +97,67 @@ async function getGHData(
...
+					repositoryLastCommit[repository.name],
+				);
+
+				adminTeamSlugs.forEach((adminSlug: string) => {
+					if (acc[adminSlug] == undefined) {
+						acc[adminSlug] = [repo];
+					} else {
+						acc[adminSlug]?.push(repo);
+					}
+				});
```
akash1810, line 153: Appreciate this isn't introduced in this PR, but `acc[adminSlug]?.push` is usually an indication that we can refactor for readability. Happy to attempt this outside of this PR.

packages/github-data-fetcher/src/handler.ts
```diff
@@ -173,7 +184,34 @@ export const main = async (): Promise<void> => {
...
+	const unchangedRepos: Repository[] = oldRepos.filter((oldRepo) =>
+		foundUnchangedMatchOnGithub(oldRepo, currentRepos),
+	);
+	console.log(
+		`${unchangedRepos.length} repositories are unchanged since the last successful run`,
+	);
+
+	const reposThatNeedUpdating: RepositoriesResponse = currentRepos.filter(
+		(newRepo) => !unchangedRepos.map((r) => r.name).includes(newRepo.name),
+	);
```
akash1810, line 206: minor: we're performing the same `map` on each iteration of `filter` here. We can refactor for performance, but mainly for readability/brevity:

```ts
const unchangedRepoNames = unchangedRepos.map((r) => r.name);

const reposThatNeedUpdating = currentRepos.filter((repo) => !unchangedRepoNames.includes(repo.name));
```

packages/github-data-fetcher/src/repoMatching.ts
```diff
@@ -0,0 +1,41 @@
...
+		const oldPush: number = new Date(oldRepo.pushed_at).getTime();
+		const matchingUpdateTime = newUpdate == oldUpdate;
+		const matchingPushTime = newPush == oldPush;
+		return matchingPushTime && matchingUpdateTime;
+	} else {
+		return false;
+	}
+}
+
+export function foundUnchangedMatchOnGithub(
```
akash1810, line 29: minor: this function name isn't immediately obvious, I think because the name doesn't necessarily match the `boolean` return type.  Would `isCachedRepositoryStale` be an alternative?

packages/github-data-fetcher/src/repoMatching.ts
```diff
@@ -0,0 +1,41 @@
...
+		newRepo.pushed_at &&
+		oldRepo.updated_at &&
+		oldRepo.pushed_at
+	) {
+		const newUpdate: number = new Date(newRepo.updated_at).getTime();
+		const oldUpdate: number = new Date(oldRepo.updated_at).getTime();
+		const newPush: number = new Date(newRepo.pushed_at).getTime();
+		const oldPush: number = new Date(oldRepo.pushed_at).getTime();
+		const matchingUpdateTime = newUpdate == oldUpdate;
+		const matchingPushTime = newPush == oldPush;
```
akash1810, line 22: Prefer `===` to compare both the value and the type at runtime. From the annotations, the types do align, but I think that's only compile time safety.

akash1810 (approved):
Brilliant! 🏆 

Added a couple of non-blocking comments.

packages/common/src/github/github.ts
NovemberTang: one of the fields exposed by the api (ironically, i get rid of it in #132 ) is repo size, which would do the same thing i reckon. I'll modify that PR to keep it in.

NovemberTang: ok i've checked and i actually kept `size`. phew!

PR #132, opened 2022-12-20
Remove fields, make dates strings

NovemberTang:
## What does this change?

1. Removed 8 fields that are not needed to support any of our use cases
2. Changed Repository date fields to strings.
3. Introduced a method to handle nulls from GitHub

## Why?

1. At the moment, we are only using a subset of the fields that GH provides us with. For maintainability, we agreed only to support fields that have concrete use cases (such as usage in RepoCop), so I've removed fields that don't get used. To keep the data fetcher decoupled from the API, it would be worth exploring storing all of the data from github, and have the github-lens api expose only the fields we care about, but this can be done at a later date.
2. Previously, the Repository object was only being used for writing values out to JSON. Since #131 , we are using it to read data in. Currently, that data is not being strictly typechecked, meaning date fields like `updated_at`, are actually being loaded as strings, which isn't being caught by compilers, and could cause runtime errors. Better typechecking of the values loaded from s3 can be addressed in a follow-up PR, but for now the types in the interface are accurate.
3. Less possible type variation makes my brain hurt less

## How has it been verified?

All unit tests still pass, checked new types manually, added a unit test to verify behaviour of asRepo


packages/github-data-fetcher/src/transformations.test.ts
```diff
@@ -32,4 +32,20 @@ describe('repository objects', function () {
...
+		expect(actual.pushed_at).toStrictEqual('2011-01-26T19:06:43Z');
+		expect(actual.updated_at).toStrictEqual('2011-01-26T19:14:43Z');
+		expect(actual.created_at).toStrictEqual('2011-01-26T19:01:12Z');
+		expect(actual.size).toStrictEqual(108);
+		expect(actual.owners).toStrictEqual([]);
+		expect(actual.archived).toStrictEqual(false);
+		expect(actual.languages).toStrictEqual([]);
+		expect(actual.topics).toStrictEqual(['production']);
+		expect(actual.default_branch).toStrictEqual('master');
+		expect(actual.lastCommit).toStrictEqual(undefined);
```
akash1810, line 48: `toStrictEqual(undefined)` is a bit odd here. `undefined` is the absence of data, whereas `null` is the intentional omission of data (more [here](https://www.geeksforgeeks.org/undefined-vs-null-in-javascript/)). Should this be [`toBeUndefined`](https://jestjs.io/docs/expect#tobeundefined)?

packages/github-data-fetcher/src/transformations.ts
```diff
@@ -27,25 +14,28 @@ export const asMember = (member: MemberResponse, teams: string[]): Member => {
...
 		name: repo.name,
-		full_name: repo.full_name,
 		private: repo.private,
-		description: repo.description,
-		created_at: parseDateString(repo.created_at),
-		updated_at: parseDateString(repo.updated_at),
-		pushed_at: parseDateString(repo.pushed_at),
+		created_at: eliminateNull<string>(repo.created_at),
+		updated_at: eliminateNull<string>(repo.updated_at),
+		pushed_at: eliminateNull<string>(repo.pushed_at),
```
akash1810, line 36: Given our reliance on these fields, would it be easier to filter out the `null` or `undefined` instances, rather than allow them?

akash1810: Could we just drop them, as in not write them to the S3 object?

kenoir (approved):

PR #133, opened 2023-01-02
Refactor Repocop

nicl:
*Note, while I believe this represents an improvement (and have commented lines to explain specific changes), this is inevitably subjective to some extent. At the least, I hope it is thought-provoking even if we do not merge.*

Some misc refactorings to (hopefully) improve readability through naming changes, simplifying the data model (to use case classes over enums), and remove some tests. Scala 3 features are preferred so e.g. 'given' is used instead of the now discouraged 'implicit' keyword.

Note, I've assumed that Github API responses return a consistent date format. The previous error handling suggests this may not be the case in practice, so we may need to modify this part to handle errors again.


packages/repocop/src/main/scala/GHLensAPI.scala
```diff
@@ -5,42 +5,41 @@ import io.circe.parser.parse
...
 
-  def getRepos: Either[Throwable, List[Repository]] = request.map { response =>
-    if (response.is2xx) {
-      extractRepoListsFromText(response.text())
-    } else {
-      val msg = "Non 2xx status code from github lens"
-      println(msg)
-      Left(new Exception(msg))
+  def getRepos: Either[Throwable, List[Repository]] = {
+    val result = Try(
```
nicl, line 15: Let's inline this. The previous helper method 'request' was poorly named and only used here.

packages/repocop/src/main/scala/GHLensAPI.scala
```diff
@@ -5,42 +5,41 @@ import io.circe.parser.parse
...
-    val parsed: Either[Error, Json] = parse(requestText)
-    val payload: Either[Error, Json] =
-      parsed.map(_.hcursor.get[Json]("payload")).joinRight
-
-    val parsingResult: Either[Error, List[Repository]] =
-      (for {
-        json <- payload
-      } yield json.as[List[Repository]]).joinRight
-    parsingResult
+    given dtDecoder: Decoder[LocalDateTime] =
```
nicl, line 36: `given` instead of `implicit` - generally it seems like `implicit` should never be used now in Scala 3.

packages/repocop/src/main/scala/GHLensAPI.scala
```diff
@@ -5,42 +5,41 @@ import io.circe.parser.parse
...
+      Decoder.decodeLocalDateTimeWithFormatter(ISO_DATE_TIME)
+    given decoder: Decoder[Repository] = deriveDecoder[Repository]
+
+    for {
```
nicl, line 40: (For comprehension is more concise here than the previous version.)

packages/repocop/src/main/scala/MarkdownHelpers.scala
```diff
@@ -14,25 +13,13 @@ object MarkdownHelpers {
 
     preamble + generateMarkdownTable(rules)
   }
-  private def generateMarkdownTable(rules: List[Rules.RepoRule]): String = {
-    val repoRules: List[(String, (String, String))] = rules
```
nicl, line 18: This whole bit was quite hard to read before (because of tuple handling) and has been simplified.

packages/repocop/src/main/scala/Repository.scala
```diff
@@ -1,40 +1,22 @@
...
 import java.time.temporal.ChronoUnit
 import scala.util.{Failure, Success, Try}
 
 //This case class should always contain a subset of fields from packages/common/src/model/github.ts
 case class Repository(
     name: String,
-    created_at: String,
-    updated_at: String,
-    pushed_at: String,
+    created_at: LocalDateTime,
```
nicl, line 10: Depending on how consistent the GH API is we may need to make these be Options.

packages/repocop/src/main/scala/Rules.scala
```diff
@@ -1,79 +1,109 @@
...
-  enum RepoRule(
-      statement: Repository => Boolean,
-      val violationMessage: String,
-      val ruleJustification: String = ""
-  ):
-    def evaluate(r: Repository): Boolean = statement(r)
+case class Rule(
+    name: String,
+    description: String,
+    check: Repository => Option[Error]
```
nicl, line 12: This is the key model change here - rather than returning a boolean, we now return an (optional) error. This allows for contextual error messages, which aren't used yet but are likely to be useful in the future.

packages/repocop/src/main/scala/Rules.scala
```diff
@@ -1,79 +1,109 @@
...
+    errorMessage: Repository => String
+): Repository => Option[Error] = { (repo: Repository) =>
+  {
+    if (!pred(repo)) Some(Error(name, errorMessage(repo)))
+    else None
+  }
+}
+
+object Rules {
+  val hasOwner = Rule(
```
nicl, line 27: Enums are no longer used as they are not really a suitable fit here - rules aren't a Sum type.

packages/repocop/src/main/scala/Rules.scala
```diff
@@ -1,79 +1,109 @@
...
-      Rules.RepoRule.values.map(_.evaluate(repository)).toList
-    ruleNames.zip(ruleBools).toMap
+    teamRepos.map(repo => {
+      Report(repo.name, checkRepo(all, repo))
+    })
   }
 
-  def evaluateRulesForRepos(repos: List[Repository]): List[EvaluatedRepo] = {
-    repos.map(repo => EvaluatedRepo(repo.name, evaluateRulesForRepo(repo)))
+  def checkRepo(rules: List[Rule], repo: Repository): List[Error] = {
```
nicl, line 88: For cases like these, where the method is contained in an object, we can improve naming by leveraging the object name. E.g.

```
Rules.evaluateRulesForRepo  // is verbose and repeats the 'Rules' part
Rules.checkRepo // it is obvious that we are checking rules here already
```

(`Rules.evaluateForRepo` would also be a good name here.)

packages/repocop/src/test/scala/MarkdownSpec.scala
```diff
@@ -3,19 +3,14 @@ package com.gu.repocop
...
-
-    val filename = "RepoRules.md"
-    val source = Source.fromFile(filename)
-    val fileContent: String = source.getLines().toList.mkString("\n") + "\n"
-    source.close()
-
-    MarkdownHelpers.createPage(
-      Rules.RepoRule.values.toList
-    ) shouldEqual fileContent
+    val fileContent = Source.fromFile("RepoRules.md").mkString
```
nicl, line 13: (Rather than splitting lines and then rejoining we can just use `.mkString` here.)

packages/repocop/src/main/scala/Repository.scala
```diff
@@ -1,40 +1,22 @@
...
-      LocalDateTime.parse(created_at, ISO_DATE_TIME),
-      LocalDateTime.parse(updated_at, ISO_DATE_TIME),
-      LocalDateTime.parse(pushed_at, ISO_DATE_TIME)
-    ).max
-  )
-
-  private val midnightYesterday: LocalDateTime =
-    LocalDateTime.now().truncatedTo(ChronoUnit.DAYS).minusDays(1)
-
-  val rerunRepocop: Boolean = dateOfLastChange match
```
nicl, line 32: It didn't quite feel like the right place for this logic - it should be separate from the Repository itself - so have moved it into the Rules object.

packages/repocop/src/main/scala/Handler.scala
```diff
@@ -1,7 +0,0 @@
-package com.gu.repocop
-
-class Handler {
-  def handler: String = {
-    main()
```
NovemberTang, line 5: cdk references this handler class/method combo so the lambda won't work unless it's brought back, or it's put in a different class and the reference is changed.

packages/repocop/src/test/scala/RepositorySpec.scala
```diff
@@ -1,70 +0,0 @@
-package com.gu.repocop
-
-import org.scalatest.flatspec.AnyFlatSpec
-import org.scalatest.matchers.should.Matchers
-
-import java.time.temporal.ChronoUnit
-import java.time.{LocalDate, LocalDateTime}
-
-class RepositorySpec extends AnyFlatSpec with Matchers {
```
NovemberTang, line 9: Why have all of these tests been deleted?

packages/repocop/src/main/scala/Handler.scala
nicl: Ah okay, I was curious what this was used for but that makes sense!

packages/repocop/src/test/scala/RepositorySpec.scala
nicl: I moved the `shouldCheck` logic to the `Rules` object, as it is it not clear that it should be coupled to the Repository definition. And have added a test in `RulesSpec.scala` - https://github.com/guardian/service-catalogue/pull/133/files#diff-97392ea19ede7f9863cdc11203cf507b6ba93497b71ef9e4ffe0d4bba06ff6aeR62. It might be worth us expanding on this though/testing more cases as in this file.

It's possible I've got the logic wrong here though! The new code doesn't work if lastModified is unset (see comment on whether the date fields should be optional). This will need updating if the date fields genuinely can all be empty.

NovemberTang: at the very least i think its worth checking the negative case (that nothing happens if all 3 dates are before the cutoff)?

packages/repocop/src/main/scala/Rules.scala
nicl, line 12: Should this be an Either? Do we want to retain happy path stuff like when the check ran?

packages/repocop/src/main/scala/GHLensAPI.scala
nicl: See also: https://docs.scala-lang.org/scala3/reference/contextual/#.

packages/repocop/src/main/scala/Repository.scala
nicl: TODO - let's check this and potentially make these Options (it sounds like this will be required).

kenoir (approved):
I think we can merge this 👍

NovemberTang (approved):

PR #138, opened 2023-02-20
Sets a timeout for the CI action

NovemberTang:
Co-authored-by: @tjsilver 

## What does this change?

The default timeout for a GHA is 360 minutes (6 hours). Normally this step completes in about 5 minutes, so we've lowered the timeout

## Why?

6 hours seemed like too many.

I think this could be rolled out to other actions too.

## How has it been verified?

Initially the timeout was set to 1 minute, which failed as there was not enough time for the process complete. It was then set to 10 minutes, and passed.
Screenshot 2023-02-20 at 16 05 18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ on:
       - main
 jobs:
   CI:
+    timeout-minutes: 1
     runs-on: ubuntu-latest
 
     # See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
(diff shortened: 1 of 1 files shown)


akash1810 (commented):
Not blocking, but I'm curious to hear why we need this PR. For example, did we have a long running build? Or are we testing this feature out?

akash1810 (approved):
Not blocking, but I'm curious to hear why we need this PR. For example, did we have a long running build? Or are we testing this feature out?

NovemberTang:
> Not blocking, but I'm curious to hear why we need this PR. For example, did we have a long running build? Or are we testing this feature out?

Both! and also cost saving purposes! we had a build where the `setup-java` action had been hanging for ages and at that point I'd rather it just fail quickly (ish) rather than six hours later for my convenience and for $$$.

PR #147, opened 2023-03-16
Introduce Cloudquery

nicl:
## What does this change?

Cloudquery is a great way to collect AWS, and other, sources of data and push it into a destination - in this case Postgres.

This commit represents an initial spike and there will be plenty of follow up work to tidy things up and broaden its behaviour, e.g. by targeting additional accounts.

See also:

https://www.cloudquery.io/

## Why?

The initial aims is to validate that we can successfully run Cloudquery
[See trello card about spike](https://trello.com/c/1bMquzu6/1813-spike-cloudquery)

Even from some initial queries it is clear that this tool can help us replace some of our custom tracking code.

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -1,3 +1,6 @@
+import fs from 'fs';
+import type { GuAutoScalingGroupProps } from '@guardian/cdk/lib/constructs/autoscaling';
+import { GuAutoScalingGroup } from '@guardian/cdk/lib/constructs/autoscaling';
 import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import { GuStack } from '@guardian/cdk/lib/constructs/core';
 import {
@@ -11,7 +14,9 @@ import {
 	InstanceSize,
 	InstanceType,
 	Port,
+	UserData,
 } from 'aws-cdk-lib/aws-ec2';
+import { Effect, ManagedPolicy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
 import type { DatabaseInstanceProps } from 'aws-cdk-lib/aws-rds';
 import { DatabaseInstance, DatabaseInstanceEngine } from 'aws-cdk-lib/aws-rds';
 import {
@@ -45,6 +50,11 @@ export class CloudQuery extends GuStack {
 		};
 
 		const db = new DatabaseInstance(this, 'PostgresInstance1', dbProps);
+		const dbSecret = db.secret?.secretName;
+
+		if (!dbSecret) {
+			throw new Error('DB Secret is missing');
+		}
 
 		const applicationToPostgresSecurityGroup = new GuSecurityGroup(
 			this,
@@ -52,6 +62,7 @@ export class CloudQuery extends GuStack {
 			{ app, vpc },
 		);
 
+		// Used by downstream services that read CloudQuery data, namely Grafana.
 		new StringParameter(this, 'PostgresAccessSecurityGroupParam', {
 			parameterName: `/${stage}/${stack}/${app}/postgres-access-security-group`,
 			simpleName: false,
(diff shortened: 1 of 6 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -0,0 +1,100 @@
...
+			`cat > aws.yaml << EOL
+${awsYaml}
+EOL`,
+			`cat > postgresql.yaml << EOL
+${postgresqlYaml}
+EOL`,
+			`# Replace password + db host`,
+			`HOST=$(aws secretsmanager get-secret-value --secret-id ${dbSecret} --region ${this.region} | jq -r '.SecretString|fromjson|.host')`,
+			`sed -i "s/£HOST/$HOST/g" postgresql.yaml`,
+			`PASSWORD=$(aws secretsmanager get-secret-value --secret-id ${dbSecret} --region ${this.region} | jq -r '.SecretString|fromjson|.password|@uri')`,
```
akash1810, line 71: I think we should be able to get these values from the `DatabaseInstance` we're creating above, removing these two secrets.

packages/cdk/lib/cloudquery/aws.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - aws_rds_cluster_parameter_groups
+    - aws_rds_db_parameter_groups
+    - aws_rds_engine_versions
+    - aws_servicequotas_services
+  spec:
+    regions:
+      - eu-west-1
+    accounts:
+      - id: deployTools
+        local_profile: deployTools
```
akash1810, line 29: Could we use EC2 instance profiles here?

packages/cdk/lib/cloudquery/postgresql.yaml
```diff
@@ -0,0 +1,9 @@
+kind: destination
+spec:
+  name: 'postgresql'
+  registry: 'github'
+  path: 'cloudquery/postgresql'
+  version: 'v3.0.0'
+  spec:
+    #TODO put credentials and adress later
```
akash1810, line 8: Could we IAM authentication?

packages/cdk/lib/cloudquery/postgresql.yaml
```diff
@@ -0,0 +1,9 @@
...
+    connection_string: 'postgresql://postgres:£PASSWORD@£HOST:5432/postgres?sslmode=disable'
```
akash1810, line 9: Why are we disabling `sslmode`?

packages/cdk/lib/cloudquery/bootstrap.py
```diff
@@ -0,0 +1 @@
+# TODO: replace user data bash stuff with this to be create python script
```
akash1810, line 1: Why are we using a python script here?

packages/cdk/lib/cloudquery.ts
```diff
@@ -0,0 +1,100 @@
...
+
+		const awsYaml = fs.readFileSync(__dirname + '/cloudquery/aws.yaml', {
+			encoding: 'utf-8',
+		});
+		const postgresqlYaml = fs.readFileSync(
+			__dirname + '/cloudquery/postgresql.yaml',
+			{
+				encoding: 'utf-8',
+			},
+		);
```
akash1810, line 54: This mechanism of creating configuration differs from most other services which opt to have Riff-Raff upload the templated files to S3, then the user data downloads them, and substitutes values.

Any reason to use this mechanism?

akash1810:
This PR is still a WIP, so converting it to draft so that it doesn't get picked up by the daily [prnouncer message](prnouncer).

packages/cdk/lib/cloudquery/postgresql.yaml
akash1810: Docs are here - https://aws.amazon.com/premiumsupport/knowledge-center/rds-postgresql-connect-using-iam/. I've not done this before, so keen to pair to learn!

packages/cdk/lib/cloudquery/aws.yaml
kenoir: Suspect this will use the AWS default credentials chain from the golang AWS SDK, so it will use the instance profile by default if running in EC2. If you're using this config locally setting an env var `AWS_PROFILE=deplyTools` I think would work.

packages/cdk/lib/cloudquery.ts
```diff
@@ -64,5 +74,61 @@ export class CloudQuery extends GuStack {
...
+			`cat > postgresql.yaml << EOL
+${postgresqlYaml}
+EOL`,
+			`# Replace password + db host`,
+			`HOST=$(aws secretsmanager get-secret-value --secret-id ${dbSecret} --region ${this.region} | jq -r '.SecretString|fromjson|.host')`,
+			`sed -i "s/£HOST/$HOST/g" postgresql.yaml`,
+			`PASSWORD=$(aws secretsmanager get-secret-value --secret-id ${dbSecret} --region ${this.region} | jq -r '.SecretString|fromjson|.password|@uri')`,
+			`sed -i "s/£PASSWORD/$PASSWORD/g" postgresql.yaml`,
+			`echo "#!/bin/sh\n/cloudquery sync /aws.yaml /postgresql.yaml" > /etc/cron.daily/cloudQuery`,
+			`chmod +x /etc/cron.daily/cloudQuery`, // TODO ship logs.
```
akash1810, line 108: I think we can achieve log shipping by using [systemd timers](https://github.com/guardian/recommendations/blob/0381138afe5b6be4843951458cad7a4fa142fca1/scheduled-jobs.md#unix--ec2) rather than `cron`, and setting the [`log-format`](https://www.cloudquery.io/docs/reference/cli/cloudquery) flag to `json`. These two should result in JSON logs in `journald`, and allow us to use [devx-logs](https://github.com/guardian/devx-logs).

packages/cdk/lib/cloudquery.ts
kenoir: It feels less complicated to me to do it this way, why put Riff-Raff & S3 in the middle? Potentially this won't scale with [limits on the size of userdata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html) though.

packages/cdk/lib/cloudquery.ts
nicl: Really good shout - suggest doing this in a follow up PR.

packages/cdk/lib/cloudquery/bootstrap.py
nicl: Placeholder to make user data easier to maintain but also made RR happy - as it requires an artifact to upload.

Suggest tidying in a follow-up PR. Also a Trello card here: https://trello.com/c/FtzyAPGI/1849-cloudquery-migrate-user-data-to-script

packages/cdk/lib/cloudquery/postgresql.yaml
nicl: 👍 we should do this in a follow-up.

packages/cdk/lib/cloudquery.ts
nicl: We tried this but it is not possible to reference these at runtime (i.e. in UserData) as far as we could tell.

NovemberTang (approved):

packages/cdk/lib/cloudquery/bootstrap.py
akash1810: > Placeholder to make user data easier to maintain but also made RR happy - as it requires an artifact to upload.

(too late, as merged) worth adding this as a comment in the file?

packages/cdk/lib/cloudquery.ts
akash1810: > It feels less complicated to me to do it this way

It is also a step away from the norm, potentially reducing long term maintainability?

> why put Riff-Raff & S3 in the middle?

These files can be the artifacts that Riff-Raff requires for an auto-scaling deployment. I'd suggest this is an improvement on the empty python file we're using.

packages/cdk/lib/cloudquery.ts
akash1810: I think `HOST` might match the [`instanceEndpoint` property](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_docdb.DatabaseInstance.html#instanceendpoint)?

If we move to IAM Authentication, the need for the password shouldn't be necessary too? If we can't move to IAM Authentication, I wonder if a [`DatabaseSecret`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_docdb.DatabaseSecret.html), and it's [`secretValue` property](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_docdb.DatabaseSecret.html#secretvalue) should be used?

PR #150, opened 2023-03-21
Allow things to talk to CloudQuery database

akash1810:
## What does this change?
There are two things that need to talk to the CloudQuery database:
  1. CloudQuery itself (write)
  2. Grafana (read)

This change creates a Security Group that allows connection to the database, and places its ID in an SSM Parameter to allow Grafana to easily obtain it.

This follows on from the changes in #149.

### DIFF
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -6,6 +6,7 @@ exports[`The CloudQuery stack matches the snapshot 1`] = `
     "gu:cdk:constructs": [
       "GuVpcParameter",
       "GuSubnetListParameter",
+      "GuSecurityGroup",
     ],
     "gu:cdk:version": "TEST",
   },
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
```diff
@@ -22,6 +23,65 @@ exports[`The CloudQuery stack matches the snapshot 1`] = `
...
+    "PostgresAccessSecurityGroupCloudqueryE959A23F": {
+      "Properties": {
+        "GroupDescription": "CloudQuery/PostgresAccessSecurityGroupCloudquery",
+        "SecurityGroupEgress": [
+          {
+            "CidrIp": "0.0.0.0/0",
+            "Description": "Allow all outbound traffic by default",
+            "IpProtocol": "-1",
+          },
+        ],
```
akash1810, line 35: This stack now has 2 egress rules allowing `0.0.0.0/0`. This feels a bit redundant, but it's all auto-generated, so 🤷🏽.

packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
```diff
@@ -195,16 +251,11 @@ exports[`The CloudQuery stack matches the snapshot 1`] = `
...
+            "PostgresAccessSecurityGroupCloudqueryE959A23F",
             "GroupId",
           ],
         },
-        "ToPort": {
-          "Fn::GetAtt": [
-            "PostgresInstance16DE4286E",
-            "Endpoint.Port",
-          ],
-        },
```
akash1810, line 207: This will be a no-op.

packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
```diff
@@ -57,6 +117,7 @@ exports[`The CloudQuery stack matches the snapshot 1`] = `
             ],
           ],
         },
+        "Port": "5432",
```
akash1810, line 120: [Updates to `Port` causes replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-port), which explains why the [continuous deployment failed](https://riffraff.gutools.co.uk/deployment/view/5ac95e4e-9cd5-4ef9-92fd-dd9910aa47fe), as Riff-Raff protects stateful resources by default.

It also explains why a [subsequent "dangerous" deployment](https://riffraff.gutools.co.uk/deployment/view/e805aa06-5be4-4e9c-932a-8fe5bb259cdb) resulted in data loss!

PR #151, opened 2023-03-23
Add developer playground account as Cloudquery source

nicl:
This is to test multi-account read.

Once happy, we should switch to using Organisational Units as sources and (commented) code has been added to achieve this. This can be switched in a follow-up PR.

See also:

https://www.cloudquery.io/docs/plugins/sources/aws/multi-account#aws-organizations

TODO: test on INFRA.


packages/cdk/lib/cloudquery/ReadOrgUnits.yaml
```diff
@@ -0,0 +1,45 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
akash1810, line 1: Can we define this template with GuCDK? Appreciate its quite small, however using GuCDK will give you CI/CD for free, and it'll also ensure all the standard tags are reliably applied.

nicl: Not sure if it is possible to delegate OU info to the Deploy Tools account - will have a look.

kenoir: This is possible, just paired with @nicl following https://docs.aws.amazon.com/organizations/latest/userguide/orgs_delegate_policies.html.

nicl: @kenoir has helped me with this so will remove this stack now.

packages/cdk/lib/cloudquery.ts
```diff
@@ -88,6 +89,24 @@ export class CloudQuery extends GuStack {
...
+		);
+
+		const devPlaygroundAccountID = new CfnParameter(
+			this,
+			'devPlaygroundAccountIDParam',
+			{
+				type: 'String',
+				description: 'Account ID for developerPlayground',
+			},
+		);
```
akash1810, line 108: Use of `CfnParameter`s in a CDK stack can be avoided here with https://github.com/guardian/private-infrastructure-config.

Its generally preferred not to use `CfnParameter`s as the value is only known at runtime, making testing difficult.

kenoir (approved):

packages/cdk/lib/cloudquery.ts
```diff
@@ -157,6 +180,26 @@ EOL`,
...
+			new PolicyStatement({
+				effect: Effect.ALLOW,
+				resources: ['*'],
+				actions: [
+					'organizations:ListAccounts',
+					'organizations:ListAccountsForParent',
+					'organizations:ListChildren',
+				],
+			}),
+		);
```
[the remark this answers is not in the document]
nicl: Will remove this code and add it to the PR description instead.

akash1810 (approved):

PR #152, opened 2023-03-24
Add us-east-1 to CloudQuery regions scanned

kenoir:
## What does this change?

Add `us-east-1` to CloudQuery regions scanned

## Why?

Some tables such as `aws_organizations_accounts` and `aws_waf_web_acls` the services for which are "global" will not get filled out unless you specify to query the magic original `us-east-1` region. This is necessary to get a good complete view of our infrastructure.

## How has it been verified?

Tested configuration with `cloudquery` locally.

### DIFF
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -823,6 +823,7 @@ spec:
   spec:
     regions:
       - eu-west-1
+      - us-east-1
     accounts:
       - id: deployTools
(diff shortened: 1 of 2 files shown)


akash1810 (approved):
not blocking - https://github.com/guardian/infosec-platform/blob/main/policies/DenyAccessToNonApprovedRegions.json defines what regions we have enabled. Worth scanning them all too?

kenoir:
> not blocking - https://github.com/guardian/infosec-platform/blob/main/policies/DenyAccessToNonApprovedRegions.json defines what regions we have enabled. Worth scanning them all too?

I think so, but that might cause unexpected results right now as trying this locally it seems to take much longer scanning extra regions.

NovemberTang (approved):

PR #156, opened 2023-03-30
refactor: replace cron with systemd timer

akash1810:
This replaces the daily cron with a systemd timer because: 
- Use of systemd is the [recommended](https://github.com/guardian/recommendations/blob/main/scheduled-jobs.md) way to run cron tasks
- It pushes logs into journald, which allows us to utilise [devx-logs](https://github.com/guardian/devx-logs) to ship these to Central ELK (we'll do this in a separate PR)

## How has it been verified?
- Deploy the cloudquery service
- Observe logs in journald (`journalctl -u cloudquery`)


tjsilver (approved):
Tested together on the box and all works as expected.

PR #157, opened 2023-03-31
feat(cloudquery): Ship logs to Central ELK

akash1810:
## What does this change?
Since #156, CloudQuery's logs are available in `journald`.

In this change we [configure CloudQuery](https://www.cloudquery.io/docs/reference/cli/cloudquery) to produce logs in JSON format. We also add the `LogKinesisStreamName`, `SystemdUnit` tags to the autoscaling group.

With these three changes, we can take advantage of [devx-logs](https://github.com/guardian/devx-logs) - devx-logs will automatically[^1] produce some configuration for [fluentbit](https://fluentbit.io/) to forward our logs on to Kinesis, allowing them to be ingested into the Central ELK stack.

## Why?
Having logs shipped to Central ELK increase our visibility[^2].

## How has it been verified?
- Deploy it ([done](https://riffraff.gutools.co.uk/deployment/view/7149a5c5-b352-4b91-8e31-654f181e4a12))
- View [logs in Central ELK](https://logs.gutools.co.uk/s/devx/goto/d6d28b20-cfeb-11ed-9d78-d12928f91852) (see screenshot below)

image

[^1]: devx-logs will [tail logs from journald, and expects the logs to be in JSON format](https://github.com/guardian/devx-logs/blob/ee2b0b9b3423eb6c86e2de3f875e7aa2da0b6949/fluentbit/application-logs.conf)
[^2]: It's shown immediate value too! From the screenshot, we can see CloudQuery is erroring!

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -1,8 +1,10 @@
+import { MetadataKeys } from '@guardian/cdk/lib/constants';
 import type { GuAutoScalingGroupProps } from '@guardian/cdk/lib/constructs/autoscaling';
 import { GuAutoScalingGroup } from '@guardian/cdk/lib/constructs/autoscaling';
 import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import {
 	GuDistributionBucketParameter,
+	GuLoggingStreamNameParameter,
 	GuStack,
 } from '@guardian/cdk/lib/constructs/core';
 import {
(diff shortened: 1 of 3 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -163,6 +165,13 @@ export class CloudQuery extends GuStack {
 
 		const asg = new GuAutoScalingGroup(this, 'asg', asgProps);
 
+		Tags.of(asg).add(
```
nicl, line 168: Probably worth us commenting these tag lines to explain the rationale.

akash1810: Good idea. Done.

NovemberTang (approved):

PR #158, opened 2023-03-31
Cache go deps

nicl:
## What does this change?

Attempts to fix Go build caching. See: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs.

## Why?

To speed up the build.

Note, this is likely to see at best a 30s improvement.

## How has it been verified?

TODO - will run a few times and compare.


tjsilver (approved):
💯

PR #159, opened 2023-03-31
Implement RDS IAM auth for Cloudquery

tjsilver:
> **Note**
> It might be easiest to review [commit by commit](https://github.com/guardian/service-catalogue/pull/159/commits).

## What does this change?
Switches to use [IAM authentication](https://repost.aws/knowledge-center/rds-postgresql-connect-using-iam) when connecting to the Postgres database. IAM authentication is preferred as it produces temporary credentials (valid for 15 minutes), in place of long-lived credentials.

As the credentials are temporary, we issue them each time the CloudQuery scheduled task runs. If we were to only issue them in the user data, they would have expired by the time CloudQuery would run.

## Why?
Use of the root credentials is not best practice. With this change, we're using temporary credentials, for a dedicated user.

## How has it been verified?
The branch has been deployed, and CloudQuery has successfully run and written to the database.

### DIFF
--- a/config/cloudquery/cloudquery.sh
+++ b/config/cloudquery/cloudquery.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e
+
+/cloudquery --log-format json --log-console sync /aws.yaml /postgresql.yaml
--- a/config/cloudquery/cloudquery.sh
+++ b/config/cloudquery/cloudquery.sh
@@ -15,7 +15,8 @@ PG_PASSWORD="$(aws rds generate-db-auth-token --hostname $RDS_HOST --port 5432 -
 # See `postgresql.yaml`
 # See https://www.postgresql.org/docs/11/libpq-connect.html#LIBPQ-CONNECT-SSLMODE for sslmode options
 # See https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution
-export PG_CONNECTION_STRING="postgresql://cloudquery:$PG_PASSWORD@$RDS_HOST:5432/postgres?sslmode=verify-full"  > /connection_string
+echo "user=cloudquery password=$PG_PASSWORD host=$RDS_HOST port=5432 dbname=postgres sslmode=verify-full" > /connection_string
+
 
 # Run cloudquery
 /cloudquery --log-format json --log-console sync /aws.yaml /postgresql.yaml
--- a/config/cloudquery/cloudquery.sh
+++ b/config/cloudquery/cloudquery.sh
@@ -1,15 +1,21 @@
 #!/usr/bin/env bash
 
 set -e
+# This is created in the Cloudformation
+# TODO: At the moment this is hard-coded, but if we want multiple stages in the future we would need to change this
+SSM_PATH=/INFRA/deploy/cloudquery/postgres-instance-endpoint-address
 
-# NOTE: The `RDS_HOST` env var is set in userdata
+# Get the address of the postgres instance endpoint from SSM, because it's more secure and future-proof
+RDS_HOST="$(aws ssm get-parameter --name $SSM_PATH --region eu-west-1 | jq .Parameter.Value -r)"
 
 # Generate temp credentials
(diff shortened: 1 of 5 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -233,6 +243,18 @@ export class CloudQuery extends GuStack {
...
+
+		asg.addToRolePolicy(
+			new PolicyStatement({
+				effect: Effect.ALLOW,
+				resources: [
+					`arn:aws:rds-db:${this.region}:${this.account}:dbuser:${attrDbiResourceId}/cloudquery`,
+				],
+				actions: ['rds-db:connect'],
+			}),
+		);
```
akash1810, line 256: This policy comes from https://repost.aws/knowledge-center/rds-postgresql-connect-using-iam.

Note: The `cloudquery` database user was manually created via:

```sql
CREATE USER cloudquery WITH LOGIN; 
GRANT rds_iam TO iamuser;
```

akash1810: Ideally we'd be able to use the [`grantConnect` function](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-rds.DatabaseInstance.html#grantwbrconnectgrantee), however there is a [bug with it](https://github.com/aws/aws-cdk/issues/11851), so do it manually for now. (We should contribute a patch upstream!)

config/cloudquery/cloudquery.sh
```diff
@@ -0,0 +1,22 @@
...
+RDS_HOST="$(aws ssm get-parameter --name $SSM_PATH --region eu-west-1 | jq .Parameter.Value -r)"
+
+# Generate temp credentials
+PG_PASSWORD="$(aws rds generate-db-auth-token --hostname $RDS_HOST --port 5432 --region eu-west-1 --username cloudquery)"
+
+# Build connection string on disk for CloudQuery to read
+# See `postgresql.yaml`
+# See https://www.postgresql.org/docs/11/libpq-connect.html#LIBPQ-CONNECT-SSLMODE for sslmode options
+# See https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution
+echo "user=cloudquery password=$PG_PASSWORD host=$RDS_HOST port=5432 dbname=postgres sslmode=verify-full" > /connection_string
```
akash1810, line 18: This is a Postgres connection string in Database Source Name (DSN) form. This is needed because CloudQuery was not able to parse a traditional connection string with a password generated from `aws rds generate-db-auth-token`. See https://github.com/guardian/service-catalogue/pull/159/commits/b35ca921c50218d017540324fc75dd827a7fc3e0.

packages/cdk/lib/cloudquery.ts
```diff
@@ -76,6 +76,13 @@ export class CloudQuery extends GuStack {
 			tier: ParameterTier.STANDARD,
 			dataType: ParameterDataType.TEXT,
 		});
+		new StringParameter(this, 'PostgresInstanceEndpointAddress', {
+			parameterName: `/${stage}/${stack}/${app}/postgres-instance-endpoint-address`,
+			simpleName: false,
+			stringValue: db.dbInstanceEndpointAddress,
+			tier: ParameterTier.STANDARD,
+			dataType: ParameterDataType.TEXT,
+		});
```
akash1810, line 85: This SSM Parameter can also be used by downstream services, such as Grafana.

packages/cdk/lib/cloudquery.ts
```diff
@@ -233,6 +243,18 @@ export class CloudQuery extends GuStack {
...
+			new PolicyStatement({
+				effect: Effect.ALLOW,
+				resources: [
+					`arn:aws:rds-db:${this.region}:${this.account}:dbuser:${attrDbiResourceId}/cloudquery`,
+				],
+				actions: ['rds-db:connect'],
+			}),
+		);
+
 		db.secret.grantRead(asg);
```
akash1810, line 258: We should be able to drop this now. Suggest we do so in a follow up PR.

akash1810 (approved):

PR #161, opened 2023-04-03
feat(cloudquery): Verify RDS SSL certificate

akash1810:
## What does this change?
Verify the full SSL certificate on the RDS connection.

To achieve this, we download and add the signing certificate authority (CA) to the OS trust store. This means the CA is trusted OS wide.

An alternative to this could be to set the [`sslrootcert`](https://www.postgresql.org/docs/11/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) value in the postgres connection string. Opted against that because it means we have to remember a file path, which is fragile. For example, when we move to an AMIgo recipe, we'd need to ensure these two paths line up, and stay in sync.

## Why?
Security!

From https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html:

> You can use Secure Socket Layer (SSL) or Transport Layer Security (TLS) from your application to encrypt a connection to a DB instance running MariaDB, Microsoft SQL Server, MySQL, Oracle, or PostgreSQL.
> 
> SSL/TLS connections provide one layer of security by encrypting data that moves between your client and a DB instance. Using a server certificate provides an extra layer of security by validating that the connection is being made to an Amazon RDS DB instance. It does so by checking the server certificate that is automatically installed on all DB instances that you provision.

## How has it been verified?
- Deploy it ([done](https://riffraff.gutools.co.uk/deployment/view/49f86e98-0c4b-4767-82d3-ce785819c5c1))
- Start cloudquery (`sudo systemctl start cloudquery`)
- Observe the logs showing postgres is being written to (`sudo journalctl -f -u cloudquery` OR look in Central ELK)

### DIFF
--- a/config/cloudquery/postgresql.yaml
+++ b/config/cloudquery/postgresql.yaml
@@ -12,4 +12,5 @@ spec:
   
   spec:
     #TODO put credentials and adress later
-    connection_string: 'postgresql://postgres:£PASSWORD@£HOST:5432/postgres?sslmode=disable'
+    # See https://www.postgresql.org/docs/11/libpq-connect.html#LIBPQ-CONNECT-SSLMODE for sslmode options
+    connection_string: 'postgresql://postgres:£PASSWORD@£HOST:5432/postgres?sslmode=verify-full'
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -148,6 +148,10 @@ export class CloudQuery extends GuStack {
 			`PASSWORD=$(aws secretsmanager get-secret-value --secret-id ${dbSecret} --region ${this.region} | jq -r '.SecretString|fromjson|.password|@uri')`,
 			`sed -i "s/£PASSWORD/$PASSWORD/g" postgresql.yaml`,
 
+			// Install RDS certificate
+			'curl https://s3.amazonaws.com/rds-downloads/rds-ca-2019-root.pem -o /usr/local/share/ca-certificates/rds-ca-2019-root.crt',
+			'update-ca-certificates',
+
 			'systemctl enable cloudquery.timer',
 			'systemctl start cloudquery.timer',
 		);
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -982,6 +982,8 @@ PASSWORD=$(aws secretsmanager get-secret-value --secret-id ",
                 },
                 " | jq -r '.SecretString|fromjson|.password|@uri')
 sed -i "s/£PASSWORD/$PASSWORD/g" postgresql.yaml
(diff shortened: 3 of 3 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -148,6 +148,10 @@ export class CloudQuery extends GuStack {
 			`PASSWORD=$(aws secretsmanager get-secret-value --secret-id ${dbSecret} --region ${this.region} | jq -r '.SecretString|fromjson|.password|@uri')`,
 			`sed -i "s/£PASSWORD/$PASSWORD/g" postgresql.yaml`,
 
+			// Install RDS certificate
+			'curl https://s3.amazonaws.com/rds-downloads/rds-ca-2019-root.pem -o /usr/local/share/ca-certificates/rds-ca-2019-root.crt',
+			'update-ca-certificates',
```
akash1810, line 153: Followed instructions from https://superuser.com/a/719047.

config/cloudquery/postgresql.yaml
```diff
@@ -6,4 +6,4 @@ spec:
   version: 'v3.0.0'
   spec:
     #TODO put credentials and adress later
-    connection_string: 'postgresql://postgres:£PASSWORD@£HOST:5432/postgres?sslmode=disable'
+    connection_string: 'postgresql://postgres:£PASSWORD@£HOST:5432/postgres?sslmode=verify-ca'
```
akash1810, line 9: The value `verify-ca` is taken from [the `pgxpool` docs](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#ParseConfig), which CloudQuery uses.

packages/cdk/lib/cloudquery.ts
akash1810: The `rds-ca-2019` certificate authority expires in 2024, and does not support automatic rotation, meaning we'll have to [perform the rotation manually](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html) (with potential downtime).

We should move to the `rds-ca-rsa2048-g1` which _does_ support automatic rotation!

See:
  - https://aws.amazon.com/about-aws/whats-new/2023/01/amazon-rds-new-ssl-tls-certificates-certificate-controls/
  - https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.RegionCertificateAuthorities

NovemberTang: do we have work planned to do this migration?

NovemberTang (approved):

packages/cdk/lib/cloudquery.ts
akash1810: > do we have a ticket planned to do this migration?

Not (yet) on trello, but we do have https://github.com/guardian/cdk/issues/1786.

akash1810: Created the Trello card now - https://trello.com/c/gqPcnFh7.

PR #164, opened 2023-04-11
refactor(cloudquery): Use `/opt/cloudquery` rather than `/` as base directory

akash1810:
Download files to `/opt/cloudquery` rather than `/`. Downloading files to a dedicated directory makes things easier to manage, and differentiate system files from our files.

## How has it been verified?
- [Deployed the branch](https://riffraff.gutools.co.uk/deployment/view/204be484-3520-40d0-9a1a-253a8eae8ef4)
- Ran CloudQuery via `systemctl start cloudquery`
- Observed the logs showing CloudQuery continues to run successfully (`journalctl -f -u cloudquery`)

### DIFF
--- a/config/cloudquery/cloudquery.service
+++ b/config/cloudquery/cloudquery.service
@@ -2,7 +2,7 @@
 Description=CloudQuery
 
 [Service]
-ExecStart=/cloudquery.sh
+ExecStart=/opt/cloudquery/cloudquery.sh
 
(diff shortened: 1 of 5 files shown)


tjsilver (approved):
I observed the same `failed to CloseAndRecv client`error on the instance as we saw before but could see that cloudquery was running successfully.

PR #167, opened 2023-04-12
feat(cloudquery): Verify checksum of cloudquery binary

akash1810:
## What does this change?
Adds a check for the checksum of the CloudQuery binary to the user data.

## Why?
It's good practice to verify the checksum of downloaded third-party binaries.

## How has it been verified?
TBD.


tjsilver (approved):
👍

PR #168, opened 2023-04-12
chore(deps): Update CloudQuery to v2.5.3

akash1810:
## What does this change?
This is the [latest version](https://github.com/cloudquery/cloudquery/releases?q=cli) of the CloudQuery CLI.

## Why?
Keeping up to date is good!

## How has it been verified?
The checksum was obtained from https://github.com/cloudquery/cloudquery/releases/download/cli-v2.5.3/checksums.txt and also confirmed via:

```sh
curl -sL https://github.com/cloudquery/cloudquery/releases/download/cli-v2.5.3/cloudquery_linux_arm64 | shasum -a 256
```


nicl (approved):
🌻

PR #170, opened 2023-04-12
Run Cloudquery in all of our AWS accounts

nicl:
## What does this change?

Run Cloudquery in all of our AWS accounts.

## Why

So that we can inspect all of our AWS accounts to understand how we are using them.

## How

This uses the OU support that Cloudquery provides to query all of our AWS accounts without having to manually specify them. The docs on how this works are here:

https://www.cloudquery.io/docs/plugins/sources/aws/multi-account#aws-organizations

Cloudquery discovers accounts by querying AWS for the OU members, and then assumes a role in each target accounts to query it.

Note, the process isn't entirely automatic; the cloudquery-access role must be present in target accounts, and this is only present in accounts the aws-account-setup manages. See:

https://github.com/guardian/aws-account-setup/pull/58

## How has it been verified?

TODO - suggestion is that we deploy this branch for a day and see how it goes.

### DIFF
--- a/config/cloudquery/aws.yaml
+++ b/config/cloudquery/aws.yaml
@@ -25,7 +25,7 @@ spec:
     regions:
       - eu-west-1
       - us-east-1
-  org:
-    member_role_name: cloudquery-access # See: https://github.com/guardian/aws-account-setup/pull/58.
-    organization_units:
-      - £TARGET_ORG_UNIT
+    org:
+      member_role_name: cloudquery-access # See: https://github.com/guardian/aws-account-setup/pull/58.
+      organization_units:
+        - £TARGET_ORG_UNIT
--- a/config/cloudquery/aws.yaml
+++ b/config/cloudquery/aws.yaml
@@ -21,6 +21,7 @@ spec:
     - aws_rds_db_parameter_groups
     - aws_rds_engine_versions
     - aws_servicequotas_services
+  concurrency: 100000 # this is known to work with a t4g.medium and our current sources but not optimised.
   spec:
     regions:
       - eu-west-1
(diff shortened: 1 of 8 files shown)


nicl:
@NovemberTang Organisation Unit - this is the grouping abstraction that AWS uses for collections of accounts. E.g. we have the following ones:

https://github.com/guardian/private-infrastructure-config/blob/main/src/organisational-units.ts#L3

packages/cdk/lib/cloudquery.ts
```diff
@@ -197,7 +179,7 @@ export class CloudQuery extends GuStack {
 			vpcSubnets: { subnets: privateSubnets },
 			minimumInstances: 1,
 			userData: userData,
-			instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
+			instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
```
akash1810, line 182: Is this in reaction to seeing a `t4g.small` struggle with the new accounts? Or just a preemptive change?

nicl: @akash1810 it is crashing yep :( . I am still tweaking the numbers here to be honest - Cloudquery has a 'concurrency' parameter to control memory usage too which we may need to set. Medium term we are probably better off making this a Fargate task and then configuring quite high memory/cpu - it would still end up cheaper mostly likely as we'd only pay for the duration of the task.

akash1810 (approved):

config/cloudquery/aws.yaml
```diff
@@ -21,12 +21,12 @@ spec:
     - aws_rds_db_parameter_groups
     - aws_rds_engine_versions
     - aws_servicequotas_services
+  concurrency: 100000 # this is known to work with a t4g.medium and our current sources but not optimised.
```
nicl, line 24: *not* known - but I am hopeful. Will adjust post-merge!

PR #169, opened 2023-04-12
chore(deps): Update cloudquery/aws to v16.1.0

akash1810:
## What does this change?
This is a major bump, with [breaking changes](https://github.com/cloudquery/cloudquery/releases?q=plugins-source-aws). Reading https://www.cloudquery.io/docs/advanced-topics/migrations and following #162, CloudQuery should automatically perform the necessary steps to migrate.

## Why?
Keeping up to date.

## How has it been verified?
I've tested locally via #171.

PR #171, opened 2023-04-12
feat(cloudquery): Add a stack for local development 

akash1810:
## What does this change?
Adds a collection of Docker containers to enable running CloudQuery locally.

Locally, we have:
  - Postgres
  - Grafana
  - CloudQuery

## Why?
This improves the feedback loop, and means we're not blocked by having only the one production environment.

It also reduces the risk of version upgrades. See #169.

## How has it been verified?
Follow the instructions in [the README](https://github.com/guardian/service-catalogue/blob/97da54594318f7a8af7e7ec738fca7d73bfca3fa/packages/cloudquery/README.md). Do they work?


packages/cloudquery/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
akash1810, line 1: A `package.json` file appears to be the contract of the repository. See the [repocop directory](https://github.com/guardian/service-catalogue/blob/0db069f22ac37efda7bf835f1d52b6cc22afe070/packages/repocop/package.json).

packages/cloudquery/dev-config/cloudquery.yaml
```diff
@@ -0,0 +1,46 @@
+---
+kind: source
+spec:
+  name: 'aws'
+  path: 'cloudquery/aws'
+  version: 'v15.7.0'
+  destinations: ['postgresql']
+  skip_tables:
```
nicl, line 8: Rather than skipping tables here, I wonder if we should allow only a few tables that are helpful for testing purposes but less likely to hit rate limits/impact prod or other users?

nicl (approved):
This is great! Both functionality and also big fan of moving the Cloudquery stuff into a separate package - no more top-level config :)

Only question is around which tables and accounts we should index locally:

* +1 to only indexing dev playground
* but should we only index a couple of tables to speed things up and further reduce risk?

Given it's only dev playground I'm less concerned, but something to consider anyway.

packages/cloudquery/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Requirements
+- Docker
+
+## Running
+1. Get AWS credentials from Janus (check [cloudquery.yaml](./dev-config/cloudquery.yaml) for the AWS profiles needed)
+2. Run:
+   
+   ```sh
+   npm start --workspace cloudquery
```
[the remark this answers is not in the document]
akash1810: 👍🏽  good point! The [latest commit](https://github.com/guardian/service-catalogue/pull/171/commits/1d380c86ff442d9da15efa29296a5ee419a7b476) tries to address this. WDYT?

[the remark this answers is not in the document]
tjsilver: Fantastic!

packages/cloudquery/dev-config/cloudquery.yaml
akash1810: Sounds sensible. Will do this in a separate PR, if that's ok?

PR #180, opened 2023-04-28
Add github as data source

JuliaBrigitte:
## What does this change?
Add github as a source to cloudquery
repository cdk only for test purposes

(when I tried to comment aws back in, localhost:3000 did not load, but I might be in desperate need of a restart.)


akash1810 (approved):
Brilliant!

PR #182, opened 2023-05-02
docs(cloudquery): Add manually applied SQL to VCS

akash1810:
## What does this change?
When setting up CloudQuery, we manually executed a few SQL commands. For prosperity, add them to the repository. We should eventually employ a database tool like [flyway](https://documentation.red-gate.com/fd/migrate-184127460.html) or [prisma](https://www.prisma.io/), but for now just place them in VCS.

## Why?
The more we have in VCS the better!

## How has it been verified?
N/A.


packages/cloudquery/sql/02.sql
```diff
@@ -0,0 +1,16 @@
+-- Create users for Grafana CODE and PROD, with readonly access.
+-- See https://grafana.com/docs/grafana/latest/datasources/postgres/#database-user-permissions-important
+
+-- Unfortunately Grafana does not support RDS IAM authentication.
+-- See https://github.com/grafana/grafana/discussions/48170
+
+CREATE USER grafanareadercode WITH PASSWORD 'REDACTED';
```
akash1810, line 7: Not entirely sure what's best practice here...

nicl (commented):
I'm all for adding the SQL to VCS. But less keen on the myriad READMEs in subdirectories - it feels like worse DX than a single one under package/cloudquery.

Could we split this into 2 PRs or drop the README changes? Keen to discuss further that side anyway.

PR #183, opened 2023-05-04
Add guardian/cloudformation-template-summary plugin

nicl:
The plugin lives here:

https://github.com/guardian/cq-source-cloudformation-template-summary

We have published a v0.1.0 tag version for now, which will become a v1.0.0 once we are happy it is stable.

Note: the plan is to migrate this plugin into the official AWS plugin but that is not wholly within our control.

## What does this change?

Introduces our cloudformation-template-summary plugin, to add Cloudformation template metadata as a source.

For now, we use a v0.1.0 tag, which is based off [this PR](https://github.com/guardian/cq-source-cloudformation-template-summary/pull/1).

## Why?

To track GuCDK usage (see [here](https://github.com/guardian/cdk/pull/1394)).

## How has it been verified?

It has been run locally with the dev config successfully. But testing on PROD is also required as the auth mechanism (assumed roles) is different. We may also test this locally following some Janus changes (regular dev roles deny assuming other roles).

UPDATE: have successfully tested in INFRA! Search Grafana for the `cloudformation_template_summaries` table to see the data.

### DIFF
--- a/packages/cloudquery/dev-config/cloudquery.yaml
+++ b/packages/cloudquery/dev-config/cloudquery.yaml
@@ -53,6 +53,21 @@ spec:
     accounts:
       - id: 'developerPlayground'
         local_profile: 'developerPlayground'
+---
+kind: source
+spec:
+  name: 'template-summary'
+  path: 'guardian/cloudformation-template-summary'
+  version: 'v0.1.0'
+  tables: ['*']
+  destinations: ['postgresql']
+  spec:
+    regions:
+      - eu-west-1
+      - us-east-1
+    accounts:
+      - id: 'developerPlayground'
+        local_profile: 'developerPlayground'
 
 ---
 kind: source
--- a/packages/cloudquery/dev-config/cloudquery.yaml
+++ b/packages/cloudquery/dev-config/cloudquery.yaml
@@ -58,7 +58,7 @@ kind: source
 spec:
   name: 'template-summary'
   path: 'guardian/cloudformation-template-summary'
-  version: 'v0.1.0'
+  version: 'v0.3.0'
   tables: ['*']
   destinations: ['postgresql']
   spec:
--- a/packages/cloudquery/prod-config/cloudquery.sh
+++ b/packages/cloudquery/prod-config/cloudquery.sh
@@ -17,12 +17,17 @@ PG_PASSWORD="$(aws rds generate-db-auth-token --hostname $RDS_HOST --port 5432 -
 # See https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution
 echo "user=cloudquery password=$PG_PASSWORD host=$RDS_HOST port=5432 dbname=postgres sslmode=verify-full" > /opt/cloudquery/connection_string
 
-
 # Run cloudquery
 /opt/cloudquery/cloudquery \
   --log-format json \
   --log-console \
   sync \
-  /opt/cloudquery/template-summary.yaml \
   /opt/cloudquery/aws.yaml \
+  /opt/cloudquery/postgresql.yaml
+
+/opt/cloudquery/cloudquery \
+  --log-format json \
+  --log-console \
+  sync \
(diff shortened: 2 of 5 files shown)


packages/cloudquery/prod-config/template-summary.yaml
akash1810, line 1: I think you'll need to `cloudquery sync` this new file from [`cloudquery.sh`](https://github.com/guardian/service-catalogue/blob/main/packages/cloudquery/prod-config/cloudquery.sh).

The changes in https://github.com/guardian/service-catalogue/pull/184 might provide some inspiration.

akash1810 (commented):
Exciting!

packages/cloudquery/prod-config/template-summary.yaml
nicl: ah yep meant to do that thanks - will add.

NovemberTang (approved):

packages/cloudquery/dev-config/cloudquery.yaml
```diff
@@ -53,6 +53,21 @@ spec:
     accounts:
       - id: 'developerPlayground'
         local_profile: 'developerPlayground'
+---
+kind: source
+spec:
+  name: 'template-summary'
+  path: 'guardian/cloudformation-template-summary'
+  version: 'v0.1.0'
```
akash1810, line 61: This version differs from that in the prod config. Is that intentional?

packages/cloudquery/prod-config/cloudquery.sh
```diff
@@ -23,5 +23,6 @@ echo "user=cloudquery password=$PG_PASSWORD host=$RDS_HOST port=5432 dbname=post
   --log-format json \
   --log-console \
   sync \
+  /opt/cloudquery/template-summary.yaml \
```
akash1810, line 26: Any particular reason to collect this data in the same `sync` step as the core AWS stuff?

I ask because in #184 we were debating doing the same, but settled on a separate step, with no strong opinions.

packages/cloudquery/dev-config/cloudquery.yaml
nicl: Ah good spot! This is not intentional at all. While v0.1.0 works locally, v.0.3.0  contains some key fixes for PROD (which relies on OU lookup). Will align.

packages/cloudquery/prod-config/cloudquery.sh
nicl: No reason other than expediency. I think that a separate sync is better as potentially more robust - we certainly don't want to fail one source because another has errors. Will update.

packages/cloudquery/prod-config/cloudquery.sh
```diff
@@ -17,11 +17,17 @@ PG_PASSWORD="$(aws rds generate-db-auth-token --hostname $RDS_HOST --port 5432 -
...
-
 # Run cloudquery
 /opt/cloudquery/cloudquery \
   --log-format json \
   --log-console \
   sync \
   /opt/cloudquery/aws.yaml \
   /opt/cloudquery/postgresql.yaml
+
+/opt/cloudquery/cloudquery \
```
akash1810, line 28: I don't think this'll work reliably. In https://github.com/guardian/service-catalogue/pull/159 we moved to using IAM authentication when connecting to the database. This works by issuing a token that's [valid for 15 minutes](https://repost.aws/knowledge-center/rds-postgresql-connect-using-iam). If you reestablish a db connection after 15 minutes, the token will be invalid.

That is, if the first `cloudquery sync` takes over 15 minutes to complete, the generated token will have expired, and this second sync will fail.

A solution to this would be to reissue the token beforehand - #184 might offer some inspiration for how to do this.

akash1810: https://github.com/guardian/service-catalogue/pull/186.

PR #184, opened 2023-05-04
Feat(cloudquery): Add Github source config to PROD

tjsilver:


akash1810:
I wonder if an alternative deployment for this could be to use GitHub Actions, as described in https://www.cloudquery.io/docs/deployment/github-actions.

For example, what if each repository has a workflow that would perform a `cloudquery sync` with configuration like this:

```yaml
kind: source
spec:
  name: github
  path: cloudquery/github
  version: "v5.1.1"
  tables: ["*"]
  destinations: ["postgresql"]
  spec:
    repos: ["${GITHUB_REPOSITORY}"] # See https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
```

Pros:
  - Less likely to encounter API rate limiting
  - Can be run on merge to `main`, and/or on a schedule
  - Similar to how we're performing Snyk reporting

Cons:
  - The security group of the RDS database would need to allow connections from GitHub

akash1810:
https://github.com/guardian/service-catalogue/pull/131 might offer some inspiration for how to reduce the volume of calls made to the GitHub API.

PR #185, opened 2023-05-04
refactor(cloudquery): Use AWS CDK method to get DB access

akash1810:
## What does this change?
The `grantConnect` function was patched in [AWS CDK v2.77.0](https://github.com/aws/aws-cdk/releases/tag/v2.77.0), which we updated to in #178.

## Why?
Less code for us to maintain.

## How has it been verified?
N/A - this is a no-op.


tjsilver (approved):
Neat! 👍

PR #186, opened 2023-05-05
fix(cloudquery): Ensure each `sync` can authenticate with RDS

akash1810:
## What does this change?
Issue an RDS auth token before each `cloudquery sync`.

## Why?
In #159 we moved to using IAM authentication to connect to the database. This works by issuing a token that's [valid for 15 minutes](https://repost.aws/knowledge-center/rds-postgresql-connect-using-iam). Attempts to reconnect with this token after 15 minutes fail.

Each `cloudquery sync` establishes a new connection. This means if the first `cloudquery sync` takes over 15 minutes to complete, the second sync will fail to connect to RDS as the generated token would have expired.

By issuing a token before each sync, we can be sure the connection to RDS will be successful.


nicl (approved):
Nice spot! Thanks. I had indeed not encountered this yet.

PR #187, opened 2023-05-08
refactor(cloudquery): Obtain AWS root org at launch time

akash1810:
## What does this change?
Replace the `@guardian/private-infrastructure-config` library with an AWS CLI call to obtain the AWS Organisation.

## Why?
This removes a dependency.

More widely, I wonder if all the CloudQuery configuration should be set like this. That is, any private values should be obtained in the launch configuration, written to a file, and [CloudQuery reads that value](https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution).

The downside of this change is a "shift right", as the value only becomes known at launch time, rather than build time.

## How has it been verified?
- [Deployed](https://riffraff.gutools.co.uk/deployment/view/33c188fd-f1c5-4bc4-b2cb-edaa4bea186d)
- Verified the file via `cat /opt/cloudquery/aws-organisation-root`
- Verified CloudQuery still runs via `sudo systemctl start cloudquery && sudo journalctl -fu cloudquery`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,10 +50,6 @@ jobs:
           go-version: '1.20'
           cache-dependency-path: packages/cloudformation-lens/go.sum
 
-      - uses: guardian/actions-read-private-repos@v0.1.0
-        with:
-          private-ssh-keys: ${{ secrets.PRIVATE_INFRASTRUCTURE_CONFIG_DEPLOY_KEY }}
-
       - name: Run script/ci
         run: ./scripts/ci.sh
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -2157,11 +2157,6 @@
         "prettier": "^2.4.0"
       }
     },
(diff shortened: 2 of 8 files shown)


nicl (approved):
I'm happy with this change. But @akash1810 if we are doing this, does it really suggest that we should deprecate/discourage use of `private-infrastructure-config` full stop? Because the usage here, which we are removing, seems like the canonical example of when to use it.

akash1810:
@nicl you make a great point! The motivation here is to push towards one way to configure CloudQuery with:
- Config files tracked in VCS
- Secret config is read off disk by CloudQuery, making the files tracked in VCS immutable

Having this single way would make DEV and PROD config easier to reason about as in DEV we currently read secrets from env vars, and in PROD we mutate the config using `sed`.

WDYT are these sensible reasons for this change?

akash1810:
Closing as I've changed my view on this; using `private-infrastructure-config` makes #189 easier.

PR #189, opened 2023-05-17
feat(cloudquery): Run on ECS

akash1810:
Inspired by [our experiments](https://github.com/guardian/container-experiments) during this quarter's R&D sprint, this change moves away from using EC2 to run CloudQuery and towards ECS, on Fargate.

This brings some benefits:
- No idle time
  
  The EC2 instance sits idle for ~23 hours a day.
- Implementing different schedules is easier
  
  Though it's possible to implement different schedules using systemd timers on EC2, we'd likely need to manage multiple config files, which isn't fun.
  
  The changes in this PR allow us to schedule a job as follows:

  ```ts
  const customRateTables: CustomRateTable[] = [
    {
      schedule: Schedule.rate(Duration.hours(2)),
      tables: ['aws_s3_buckets'],
    },
    {
      schedule: Schedule.rate(Duration.minutes(30)),
      tables: ['aws_lambda_functions'],
    },
  ];
  ```

  Being able to collect data from different tables at different rates is going to be crucial to adding GitHub to CloudQuery due to the rate limiting of the GitHub API. If we collect data for [`github_teams`](https://www.cloudquery.io/docs/plugins/sources/github/tables/github_teams) at a slower rate than [`github_repositories`](https://www.cloudquery.io/docs/plugins/sources/github/tables/github_repositories), we reduce the chance of getting rate limited.

- Removal of known errors

  The EC2 collector currently produces error logs when collecting data for the [organization tables](https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_organizations) . This is because we're making the request from all accounts. We know that only the DeployTools account has [this permission](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_info-operations.html), so we now make the request from just that account.
  
  Similar to the above, this is possible when using EC2, however we'd need to manage multiple config files, which isn't fun.

- Reduced infrastructure

  With the EC2 version, we have an ASG purely to facilitate deployments via Riff-Raff. This is not needed in the ECS version.

  In the EC2 version, we also place configuration in S3. The implementation of the ECS version removes this, in favour of inlining the config.

## How has it been verified?
The stack has been [deployed](https://riffraff.gutools.co.uk/deployment/view/d560b09d-901d-4863-8260-bfbded79a4db). There are a couple of ways to verify it works:
- Observe the [logs in Central ELK](https://logs.gutools.co.uk/goto/6c10fdf0-f6e5-11ed-a27a-7333ae0cfb0a)
- View the configuration of the ECS Cluster in AWS

## Notes
Sadly, I don't think it will be possible to use [centralised logging container](https://aws.amazon.com/blogs/opensource/centralized-container-logging-fluent-bit/) on Fargate as:

> Fargate tasks do not support the `DAEMON` scheduling strategy.
> – https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_services.html#service_scheduler

This means each task has three containers:
1. One to vend temporary AWS RDS IAM tokens
2. One to perform the CloudQuery work
3. One to ship logs

We should keep an eye on the cost of this, as it might become cheaper to ship the logs to CloudWatch logs, and then use [cloudwatch-logs-management](https://github.com/guardian/cloudwatch-logs-management) to route them to Central ELK.

## TODO (non-blocking)
- Fine tuning, documentation (and renaming) of https://github.com/guardian/hackday-firelens, as the logs lines in Central ELK do not quite match those from the EC2 version, with missing fields.


packages/cdk/package.json
```diff
@@ -5,14 +5,17 @@
     "test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects cdk",
     "typecheck": "tsc -noEmit",
     "cdk": "cdk",
-    "synth": "cdk synth --path-metadata false --version-reporting false"
+    "synth": "cdk synth --path-metadata false --version-reporting false",
+    "deploy:cloudquery-ecs": "AWS_PROFILE=developerPlayground cdk deploy --path-metadata false --version-reporting false Cloudquery"
```
akash1810, line 9: Added purely for a shorter feedback loop. We can update the CFN stack via `npm run deploy:cloudquery-ecs` vs. committing, waiting for CI, and waiting for Riff-Raff.

Remove once the infrastructure is moved to the Deploy Tools account, this can be removed.

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,167 @@
...
+				},
+			}),
+		});
+
+		task.addFirelensLogRouter(`${id}Firelens`, {
+			image: firelensImage,
+			logging: LogDrivers.awsLogs({
+				streamPrefix: [stack, stage, app].join('/'),
+				logRetention: RetentionDays.ONE_DAY,
+			}),
```
akash1810, line 133: Send the log shipper's logs to CloudWatch.

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		const task = new FargateTaskDefinition(scope, `${id}TaskDefinition`);
+
+		task.addContainer(`${id}Container`, {
+			image: cloudqueryImage,
+			secrets: {
+				DB_HOST: Secret.fromSecretsManager(db.secret, 'host'),
+				DB_PASSWORD: Secret.fromSecretsManager(db.secret, 'password'),
+			},
```
akash1810, line 107: Whilst we're injecting secrets to the container, they are for the database's root user. As the CloudQuery image doesn't have the AWS CLI, we can't vend a temporary token.

One solution to this could be to use the [CLI in a separate container](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-docker.html), generate a token and store it in a shared volume between the two containers.

packages/cdk/lib/ecs/config.ts
```diff
@@ -0,0 +1,117 @@
...
+			member_role_name: 'cloudquery-access',
+			organization_units: [GuardianOrganisationalUnits.Root],
+		},
+	});
+}
+
+/**
+ * Create a CloudQuery configuration for a single AWS account.
+ * Use this for those services running across the organisation which are aggregated in a single account.
+ * For example, Access Analyzer.
```
akash1810, line 98: See https://github.com/guardian/aws-account-setup/pull/67.

packages/cdk/lib/ecs/config.ts
```diff
@@ -0,0 +1,117 @@
...
+			version: `v${Versions.CloudqueryPostgres}`,
+			migrate_mode: 'forced',
+			spec: {
+				connection_string: [
+					'user=postgres',
+					'password=${DB_PASSWORD}',
+					'host=${DB_HOST}',
+					'port=5432',
+					'dbname=postgres',
+					'sslmode=disable',
```
akash1810, line 33: We should continue to verify the certificate, see https://github.com/guardian/service-catalogue/pull/161.

packages/cdk/lib/ecs/config.test.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+	awsSourceConfigForAccount,
+	awsSourceConfigForOrganisation,
+	postgresDestinationConfig,
+} from './config';
+
+describe('Config generation, and converting to YAML', () => {
+	it('Should create a destination configuration', () => {
+		const config = postgresDestinationConfig();
+		expect(dump(config)).toMatchInlineSnapshot(`
```
akash1810, line 12: These tests allow us to gain confidence in the generation of the CloudQuery configuration.

packages/cdk/lib/ecs/cloudquery.ts
```diff
@@ -0,0 +1,182 @@
...
+		const vpc = GuVpc.fromIdParameter(this, 'PrimaryVpc', {
+			/*
+			CDK wants privateSubnetIds to be a multiple of availabilityZones.
+			We're pulling the subnets from a parameter at runtime.
+			We know they evaluate to 3 subnets, but at compile time CDK doesn't.
+
+			Set the number of AZs to 1 to avoid the error:
+			  `Error: Number of privateSubnetIds (1) must be a multiple of availability zones (2).`
+			 */
+			availabilityZones: ['ignored'],
```
akash1810, line 54: 😅

packages/cdk/package.json
```diff
@@ -5,14 +5,17 @@
...
   },
   "devDependencies": {
+    "@types/js-yaml": "^4.0.5",
     "@guardian/cdk": "50.3.3",
     "@guardian/private-infrastructure-config": "github:guardian/private-infrastructure-config#v2.3.0",
     "aws-cdk": "2.77.0",
     "aws-cdk-lib": "2.77.0",
     "constructs": "10.2.8",
+    "js-yaml": "^4.1.0",
```
akash1810, line 18: This dependency is needed to convert a TypeScript `Record` into YAML configuration for CloudQuery.

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,167 @@
...
+import { postgresDestinationConfig } from './config';
+import { Versions } from './versions';
+
+const cloudqueryImage = ContainerImage.fromRegistry(
+	`ghcr.io/cloudquery/cloudquery:${Versions.CloudqueryCli}`,
+);
+
+const firelensImage = ContainerImage.fromRegistry(
+	'ghcr.io/guardian/hackday-firelens:main',
+);
```
akash1810, line 29: TODO rename [this repository](https://github.com/guardian/hackday-firelens).

packages/cdk/lib/ecs/cloudquery.ts
```diff
@@ -0,0 +1,182 @@
...
+		const customRateTables: CustomRateTable[] = [
+			{
+				schedule: Schedule.rate(Duration.hours(2)),
+				tables: ['aws_s3_buckets'],
+			},
+			{
+				schedule: Schedule.rate(Duration.minutes(30)),
+				tables: ['aws_lambda_functions'],
+			},
+		];
```
akash1810, line 160: Fairly arbitrary rates here. We'd want to tune this list over time.

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,204 @@
...
+import { RetentionDays } from 'aws-cdk-lib/aws-logs';
+import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
+import { dump } from 'js-yaml';
+import type { CloudqueryConfig } from './config';
+import { postgresDestinationConfig } from './config';
+import { Versions } from './versions';
+
+const awsCliImage = ContainerImage.fromRegistry(
+	'public.ecr.aws/aws-cli/aws-cli',
+);
```
akash1810, line 25: This is the [official AWS CLI image](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-docker.html).

nicl:
Update: @akash1810 have added a commit with some changes. The main one is to change the scheduled task interface from:

```typescript
interface CustomRateTable {
    schedule: Schedule;
    tables: string[];
}
```

to something like:

```typescript
interface CloudquerySource {
	name: string;
	description: string;
	schedule: Schedule;
	config: CloudqueryConfig;
	policies?: PolicyStatement[];
	managedPolicies?: IManagedPolicy[];
}
```

This looks more complicated but allows:

* least privilege, as tasks can be passed just the policies they require
* support for non-AWS tasks, such as the upcoming Github, Fastly etc.
* better task identification (via the name/description)

The idea is that *everything* becomes a scheduled task here, including the standard fetch-everything-from-all-accounts Cloudquery task. So the code overall becomes simpler with fewer ifs/branches.

Obviously feel free to drop the commit if you are strongly against and want to discuss!

akash1810:
> This looks more complicated but allows:
> 
> * least privilege, as tasks can be passed just the policies they require
> * support for non-AWS tasks, such as the upcoming Github, Fastly etc.
> * better task identification (via the name/description)

Love every one of these! I was struggling to add GH to this pattern; I think these changes get us closer, but I think we'd need a way to set secrets too; suggest this refactor be it's PR though.

Regarding logging, that was happening prior to your change too. We should be able to reproduce this behaviour locally to help debugging (woop containers!). WDYT - is this shippable the current version of logging?

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,204 @@
...
+import { ScheduledFargateTask } from 'aws-cdk-lib/aws-ecs-patterns';
+import type { IManagedPolicy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import { RetentionDays } from 'aws-cdk-lib/aws-logs';
+import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
+import { dump } from 'js-yaml';
+import type { CloudqueryConfig } from './config';
+import { postgresDestinationConfig } from './config';
+import { Versions } from './versions';
+
+const awsCliImage = ContainerImage.fromRegistry(
```
NovemberTang, line 23: for a potential follow-up. could we put the AWS CLI/firelens logging in a container that has cloudquery as the base, self publish, and auth that way? Would save couple of containers? @Mark-McCracken

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,204 @@
...
+				'-c',
+				[
+					`PG_PASSWORD=$(/usr/local/bin/aws rds generate-db-auth-token --hostname $DB_HOST --port 5432 --region eu-west-1 --username ${dbUser})`,
+					`echo "user=${dbUser} password=$PG_PASSWORD host=$DB_HOST port=5432 dbname=postgres sslmode=verify-full" > /var/scratch/connection_string`,
+				].join(';'),
+			],
+			essential: false,
+		});
+
+		const worker = task.addContainer(`${id}Container`, {
```
NovemberTang, line 122: worker is a bit vague

packages/cdk/lib/ecs/task.ts
```diff
@@ -0,0 +1,204 @@
...
+			entryPoint: [''],
+			command: [
+				'/bin/bash',
+				'-c',
+				[
+					`PG_PASSWORD=$(/usr/local/bin/aws rds generate-db-auth-token --hostname $DB_HOST --port 5432 --region eu-west-1 --username ${dbUser})`,
+					`echo "user=${dbUser} password=$PG_PASSWORD host=$DB_HOST port=5432 dbname=postgres sslmode=verify-full" > /var/scratch/connection_string`,
+				].join(';'),
+			],
+			essential: false,
```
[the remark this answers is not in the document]
akash1810: From the docs:

> essential?
> Type: boolean (optional, default: true)
> 
> Specifies whether the container is marked essential.
> 
> If the essential parameter of a container is marked as true, and that container fails or stops for any reason, all other containers that are part of the task are stopped. If the essential parameter of a container is marked as false, then its failure does not affect the rest of the containers in a task. All tasks must have at least one essential container.
> 
> If this parameter is omitted, a container is assumed to be essential.
> – https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.ContainerDefinitionOptions.html#essential

As I understand it - an essential container failing will cause the remaining containers in the task to fail.

However, we _do_ want this container failing to cause the rest to stop, so I can't remember why this is set to `false`; from memory it was causing the CFN deployment to fail when `true`, but will work this out properly and add an inline comment to explain it.

packages/cdk/lib/cloudquery.ts
```diff
@@ -266,5 +291,65 @@ export class CloudQuery extends GuStack {
...
+				{
+					name: 'All',
+					description: 'Data fetched across all accounts in the organisation.',
+					schedule: Schedule.rate(Duration.days(1)),
+					config: awsSourceConfigForOrganisation({
+						tables: ['*'],
+						skipTables: skipTables,
+					}),
+					managedPolicies: [
+						readonlyAccessManagedPolicy(this, 'fetch-all-managed-policy'),
```
akash1810, line 310: We're creating an instance of this managed policy for each task, which increases the chances of bumping against the [CFN template limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html). If we do see this in future, we can make instantiate just one, and pass it across.

packages/cdk/lib/ecs/cluster.ts
```diff
@@ -0,0 +1,125 @@
...
+				effect: Effect.ALLOW,
+				resources: [loggingStreamArn],
+			}),
+
+			new PolicyStatement({
+				effect: Effect.ALLOW,
+				resources: ['arn:aws:iam::*:role/cloudquery-access'],
+				actions: ['sts:AssumeRole'],
+			}),
+		];
```
akash1810, line 103: Worth keeping an eye on these, for similar reasons to [above](https://github.com/guardian/service-catalogue/pull/189/files#r1201982925).

packages/cdk/lib/ecs/task.ts
akash1810: 👍🏽  renamed to `cloudqueryTask`

PR #193, opened 2023-05-25
feat(cloudquery): Collect GitHub repository data 

akash1810:
This replaces #184 following the recent adoption of ECS in #189.

## What does this change?
In this change we add the [GitHub source](https://www.cloudquery.io/docs/plugins/sources/github/overview) to CloudQuery, initially collecting data only for repositories, once a day.

We authenticate with GitHub using a [GitHub App](https://docs.github.com/en/apps/creating-github-apps/setting-up-a-github-app/creating-a-github-app) that we manually created earlier with read-only access to the org.

CloudQuery requires three bits of GitHub config: private key, app ID, installation ID. These live in a manually created Secrets Manager resource on path `/INFRA/deploy/cloudquery/github-credentials`. The ARN of this resource has a random GUID at the end of it. The only way I could read this secret without passing in the ARN (which is [private information](https://github.com/guardian/recommendations/blob/main/github.md#repository-contents)) was to use [`Secret.fromSecretPartialArn`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_secretsmanager.Secret.html#static-fromwbrsecretwbrpartialwbrarnscope-id-secretpartialarn). Interestingly, this creates an IAM policy with a resource of:

```
"Fn::Join": [
  "",
  [
    "arn:",
    {
      "Ref": "AWS::Partition",
    },
    ":secretsmanager:",
    {
      "Ref": "AWS::Region",
    },
    ":",
    {
      "Ref": "AWS::AccountId",
    },
    ":secret:/TEST/deploy/cloudquery/github-credentials-??????",
  ],
],
```

I don't know _how_ the `--??????` works, but it does!

## Why?
Having GitHub data is integral to Service Catalogue; collecting it via CloudQuery means we can retire the [github-data-fetcher](https://github.com/guardian/service-catalogue/tree/main/packages/github-data-fetcher) and [github-lens-api](https://github.com/guardian/service-catalogue/tree/main/packages/github-lens-api) projects.

## How has it been verified?
I've manually [deployed](https://riffraff.gutools.co.uk/deployment/view/2c16c108-ec12-4ba8-a0ca-d7bc8d304826) this, and manually triggered the task. To confirm all is working as expected, the screenshot below shows when data was last collected (i.e. when the task was manually triggered):

image

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -14,6 +14,7 @@ import {
 	InstanceType,
 	Port,
 } from 'aws-cdk-lib/aws-ec2';
+import { Secret } from 'aws-cdk-lib/aws-ecs';
 import { Schedule } from 'aws-cdk-lib/aws-events';
 import type { DatabaseInstanceProps } from 'aws-cdk-lib/aws-rds';
 import { DatabaseInstance, DatabaseInstanceEngine } from 'aws-cdk-lib/aws-rds';
@@ -26,6 +27,7 @@ import { CloudqueryCluster } from './ecs/cluster';
 import {
 	awsSourceConfigForAccount,
 	awsSourceConfigForOrganisation,
+	githubSourceConfig,
 	skipTables,
 } from './ecs/config';
 import {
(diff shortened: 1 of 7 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -163,6 +177,31 @@ export class CloudQuery extends GuStack {
 						cloudqueryAccess(GuardianAwsAccounts.Security),
 					],
 				},
+				{
```
akash1810, line 180: The mechanism of passing in a load of sources to a `Cluster` is becoming a bit unwieldy, especially as we add more. Adopting a [builder pattern](https://refactoring.guru/design-patterns/builder) might make things easier to read. Suggest performing this refactor separately, and soonish.

packages/cdk/lib/cloudquery.ts
```diff
@@ -163,6 +177,31 @@ export class CloudQuery extends GuStack {
...
+							githubCredentials,
+							'app-id',
+						),
+						GITHUB_INSTALLATION_ID: Secret.fromSecretsManager(
+							githubCredentials,
+							'installation-id',
+						),
+					},
+					additionalCommands: [
+						'echo $GITHUB_PRIVATE_KEY | base64 -d > /github-private-key',
```
akash1810, line 200: I base64 encoded the key before placing it in Systems Manager to preserve the newlines. Here we base64 decode it, and write it to disk.

packages/cdk/lib/cloudquery.ts
nicl: I'm not sure how builder helps here. Typescript already supports optional parameters for types, e.g. which we use for the optional params here such as `managedPolicies?: IManagedPolicy[];` etc. So you can simply omit any that aren't relevant. The named arguments also helps here (rather than positional). Builder is meant to avoid complex constructors with many *positional* arguments, some or many of which are not used most of the time.

nicl: Ah, builder on the cluster! Makes more sense, but I still think this is overcomplicating things. Why not just move the github source definition into a named variable alongside `githubCredentials`?

akash1810: > Why not just move the github source definition into a named variable alongside `githubCredentials`?

Embarrassingly, this didn't occur to me 😬, that's better. Will refactor in a future PR, if that's OK?

nicl (approved):
⛵ 🌴 really nice!

akash1810:
Unfortunately the logs indicate that we continue to hit the GitHub API limits:

```log
GET https://api.github.com/repos/guardian/interactive-margin-fix/releases?per_page=1000: 403 API rate limit of 15000 still exceeded until 2023-05-26 17:57:28 +0000 UTC, not making remote request. [rate reset in 47m07s]
```

PR #194, opened 2023-05-25
feat(cloudquery): Add Fastly data source

akash1810:
## What does this change?
Collecting data for Fastly services once a day, this change adds [Fastly](https://www.cloudquery.io/docs/plugins/sources/fastly/overview) to the data collected using CloudQuery.

To enable this, I've manually created a Secrets Manager resource on path `/INFRA/deploy/cloudquery/fastly-credentials` with a [read-only automation token](https://docs.fastly.com/en/guides/using-api-tokens) provided by @davidfurey. The token has been set to expire on 22-Sep-2023, so we'd need to rotate it before then, should we deem this data useful.

## Why?
This might allow us to remove some of the Fastly crawling from https://github.com/guardian/aws-budgets.

## How has it been verified?
This has been [deployed](https://riffraff.gutools.co.uk/deployment/view/21fe8d9a-ef9a-461f-9655-0b52353e9f40). The ECS task has been manually run, and we can now see Fastly data in [Grafana](https://metrics.gutools.co.uk/goto/tPAO6cw4z?orgId=1).

### DIFF
--- a/packages/cdk/lib/ecs/config.ts
+++ b/packages/cdk/lib/ecs/config.ts
@@ -142,6 +142,40 @@ export function githubSourceConfig(
 	};
 }
 
+/**
+ * Configuration for the Fastly source plugin.
+ * @see https://www.cloudquery.io/docs/plugins/sources/fastly/overview#configuration
+ */
+export function fastlySourceConfig(
+	tableConfig: CloudqueryTableConfig,
+): CloudqueryConfig {
+	const { tables, skipTables } = tableConfig;
+
+	if (!tables && !skipTables) {
+		throw new Error('Must specify either tables or skipTables');
+	}
+
+	return {
+		kind: 'source',
+		spec: {
+			name: 'fastly',
+			path: 'cloudquery/fastly',
+			version: `v${Versions.CloudqueryFastly}`,
+			tables,
+			skip_tables: skipTables,
+			destinations: ['postgresql'],
+
+			// The Fastly API is rate limited to 1000 requests per hour.
+			// See https://docs.fastly.com/en/guides/resource-limits#rate-and-time-limits.
+			// TODO what's the ideal value here?!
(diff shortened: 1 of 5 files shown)


packages/cdk/lib/ecs/config.ts
```diff
@@ -142,6 +142,40 @@ export function githubSourceConfig(
...
+			path: 'cloudquery/fastly',
+			version: `v${Versions.CloudqueryFastly}`,
+			tables,
+			skip_tables: skipTables,
+			destinations: ['postgresql'],
+
+			// The Fastly API is rate limited to 1000 requests per hour.
+			// See https://docs.fastly.com/en/guides/resource-limits#rate-and-time-limits.
+			// TODO what's the ideal value here?!
+			concurrency: 1000,
```
nicl, line 171: Agree it's not very clear what the right value to use here is - trial and error is best. Some thoughts though:

* it defaults to 500k, which seems way too high(!) so 1k is better - in fact, I suspect that even 100 or less would be fine here as that can still add up to a lot of requests - e.g. 100 concurrent req each taking 2s to read and then process, still equals 3000 req/min.
* Fastly have a 1k req/hr limit on *write* requests but 'unlimited' read requests (see [here](https://developer.fastly.com/reference/api/#rate-limiting)) so hopefully we should not hit rate limiting in practice

nicl (approved):
LGTM!

packages/cdk/lib/ecs/config.ts
akash1810: > * Fastly have a 1k req/hr limit on _write_ requests but 'unlimited' read requests (see [here](https://developer.fastly.com/reference/api/#rate-limiting)) so hopefully we should not hit rate limiting in practice

Ah! Great spot. Let's update this in a future PR.

PR #198, opened 2023-05-31
Introduce Galaxies as Cloudquery source

nicl:
## What does this change?

Introduces Galaxies as a *source* for Cloudquery using an in-house plugin that lives here:

https://github.com/guardian/cq-source-galaxies

Galaxies does not (atm) support machine access to the API. Reading directly from the bucket is a 'pragmatic' solution here - it is easy to implement but vulnerable to changes in the model. The team do not expect these in the medium term though.

An initial rate of once per day is used as this data doesn't change very often.

## Why?

Galaxies is the Guardian's source of truth for people, teams, and streams. We want this data to be able to link services to teams and answer questions around **ownership**.

## How has it been verified?

TODO - test in INFRA and see how it goes. (The plugin has already been verified separately through local testing.)

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -1,10 +1,11 @@
 import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
-import { GuStack } from '@guardian/cdk/lib/constructs/core';
+import { GuStack, GuStringParameter } from '@guardian/cdk/lib/constructs/core';
 import {
 	GuSecurityGroup,
 	GuVpc,
 	SubnetType,
 } from '@guardian/cdk/lib/constructs/ec2';
+import { GuS3Bucket } from '@guardian/cdk/lib/constructs/s3';
 import { GuardianAwsAccounts } from '@guardian/private-infrastructure-config';
 import type { App } from 'aws-cdk-lib';
 import { ArnFormat, Duration } from 'aws-cdk-lib';
@@ -30,12 +31,14 @@ import {
 	awsSourceConfigForAccount,
 	awsSourceConfigForOrganisation,
 	fastlySourceConfig,
+	galaxiesSourceConfig,
 	githubSourceConfig,
 	skipTables,
 } from './ecs/config';
 import {
 	cloudqueryAccess,
 	listOrgsPolicy,
+	readBucketPolicy,
 	readonlyAccessManagedPolicy,
 	standardDenyPolicy,
 } from './ecs/policies';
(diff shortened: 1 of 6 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -268,12 +270,59 @@ export class CloudQuery extends GuStack {
...
+					readBucketPolicy(
+						actionsStaticSiteBucket.bucketArn,
+						'galaxies.gutools.co.uk/data',
+					),
+				],
+				config: {
+					kind: 'source',
+					spec: {
+						name: 'galaxies',
+						path: '"guardian/galaxies',
```
akash1810, line 304: Is the xtra `"` intentional?

packages/cdk/lib/cloudquery.ts
```diff
@@ -268,12 +270,59 @@ export class CloudQuery extends GuStack {
...
+				name: 'Galaxies',
+				description: 'Galaxies data',
+				schedule: Schedule.rate(Duration.days(1)),
+				policies: [
+					readBucketPolicy(
+						actionsStaticSiteBucket.bucketArn,
+						'galaxies.gutools.co.uk/data',
+					),
+				],
+				config: {
```
akash1810, line 300: Could we place this in the [config file](https://github.com/guardian/service-catalogue/blob/main/packages/cdk/lib/ecs/config.ts)?

packages/cdk/lib/cloudquery.ts
```diff
@@ -268,12 +270,59 @@ export class CloudQuery extends GuStack {
...
+					kind: 'source',
+					spec: {
+						name: 'galaxies',
+						path: '"guardian/galaxies',
+						version: 'v1.0.0',
```
akash1810, line 305: Could we push the version to [versions.ts](https://github.com/guardian/service-catalogue/blob/main/packages/cdk/lib/ecs/versions.ts)?

packages/cdk/lib/ecs/policies.ts
```diff
@@ -50,3 +50,11 @@ export function cloudqueryAccess(accountId: string) {
 		actions: ['sts:AssumeRole'],
 	});
 }
+
+export const readBucketPolicy = (bucketArn: string, prefix: string) => {
+	return new PolicyStatement({
+		effect: Effect.ALLOW,
+		resources: [`${bucketArn}/${prefix}/*`],
```
akash1810, line 57: The default addition of `/*` here feels excessive, and provides wider permissions than one might expect. Could the function caller pass this in for the prefix value instead, as they need it? For example:

```ts
readBucketPolicy(myBucketArn, "data/*");
```

Given the usage of this function is limited to this repository, this is a minor observation, and isn't a blocker.

packages/cdk/lib/cloudquery.ts
nicl: doh no!

packages/cdk/lib/ecs/policies.ts
nicl: No that's a good shout. It is a better API to allow the caller to limit the scope more precisely.

akash1810 (approved):
Exciting!

PR #202, opened 2023-06-01
Tidy repo now that Cloudquery has taken over

nicl:
TODO - re-add bits to main README on local dev and tidy the setup script.

## What does this change?

Previous packages:

- cdk
- cloudquery (dev)
- common
- github-data-fetcher
- github-lens-api
- repocop
- services-api

Now:

- ckd
- cloudquery
- repocop

There is some discussion to be had about whether Repocop still belongs here or should live elsewhere, in which case this repo would be solely focused on Cloudquery.

## Why?

To remove stuff we no longer plan to develop.

## How has it been verified?

CI tests only.


nicl:
@akash1810 stacks (and load balancers) deleted. The AMIs seem to be general ones so have left those for now.

akash1810:
I have removed the GitHub App named `Guardian GitHub Lens` now too.

PR #201, opened 2023-06-01
feat(cloudquery): Collect Snyk data in INFRA

akash1810:
## What does this change?
Running once a day, this change adds [Snyk](https://www.cloudquery.io/docs/plugins/sources/snyk/overview) to the ECS cluster.

Things to note, setting `tables: '*'` in the configuration was yielding errors:

```log
failed to load spec(s) from /source.yaml, /destination.yaml. Error: failed to strip yaml comments in file /source.yaml (section 1): yaml: line 7: did not find expected alphabetic or numeric character
```

Not entirely sure why, but listing the tables explicitly solves this. This isn't massively maintainable, as we'd have to be sure to update the list when we update the version. Thankfully, there aren't (currently) too many [tables](https://www.cloudquery.io/docs/plugins/sources/snyk/tables).

## Why?
This will help us answer some of this quarter's key questions:
- (Vulnerability monitoring) Does a service have vulnerability monitoring?
- (Vulnerability monitoring) Does a service have outstanding high or critical vulnerabilities?

## How has it been verified?
I've [deployed it](https://riffraff.gutools.co.uk/deployment/view/3d6dba98-5a99-4ef2-a14c-a141c36f2b75), manually run the task, and can see data in Grafana (see screenshot).

image

PR #203, opened 2023-06-01
fix(cloudquery): Attempt to get the all AWS task to complete

akash1810:
## What does this change?
The [release notes for the CloudQuery CLI v3.0.0](https://github.com/cloudquery/cloudquery/releases/tag/cli-v3.0.0) states:

> We're making it required as the previous default of ["*"] (all tables) is not suitable for production usage

This might hint to the error seen in #201:

```
failed to load spec(s) from /source.yaml, /destination.yaml. Error: failed to strip yaml comments in file /source.yaml (section 1): yaml: line 7: did not find expected alphabetic or numeric character
```

It looks like the catch-all AWS task is not working, presumably for the same reasons.

Listing all the [individual tables for AWS](https://www.cloudquery.io/docs/plugins/sources/aws/tables) is not maintainable, so use [wildcard matching](https://www.cloudquery.io/docs/advanced-topics/performance-tuning#use-wildcard-matching) instead.

Whilst this did resolve this issue, the task to collect all AWS data was getting terminated by AWS:

```log
Exit code: 1
OutOfMemoryError: Container killed due to memory usage
```

In an attempt to combat this, this change also collects EC2, Lambda, and S3 data in a different task.

We can use the same wildcard technique to simplify the Snyk source too.

## Why?
To ensure we collect all the data we're intending to.

## How has it been verified?
Each table has a `_cq_sync_time` column which we can use to determine when we last collected data for it. For S3 buckets, we can run:

```sql
SELECT max(_cq_sync_time) FROM aws_sqs_queues;
```

Prior to this change, this is `2023-05-22 14:02:42`, demonstrating that we've not collected data in a while (it should be updated daily).

After [deploying](https://riffraff.gutools.co.uk/deployment/view/a1365552-22e5-424a-8082-f33532c6d7f7) this branch, and manually running the all AWS task, this query returns `2023-06-01 22:09:05`, suggesting that we're collecting data again.

## Notes
In an attempt to summarise the state of the world on this branch, here's the configuration of the cluster:

| Tables | Frequency | Concurrency[^1] | Memory[^2] | Duration | Success |
|---|---|---|---|---|---|
| (AWS) Organisations | Daily | Default | Default | 2 | Yes |
| (AWS) Access Analyzer | Daily | Default | Default | 1 | Yes |
| (AWS) CloudFormation | Every 3 hours | Default | Default | 3 | Yes |
| (AWS) EC2 | Every 3 hours | 10000 | Default | 12 | Yes |
| (AWS) Lambda and Step Functions | Every 3 hours | 10000 | 2048 | 25 | No.[^3] |
| (AWS) S3 | Every 3 hours | 10000 | Default | 3 | Yes |
| (AWS) Everything else | Daily | 10000 | 2048 | 61 | No.[^4] |
| (GitHub) Repositories | Daily | Default | Default | 15 | Ish.[^5] |
| (GitHub) Teams (inc. members & repositories) | Weekly | Default | Default | 2 | Ish.[^5] |
| (Fastly) Services | Daily | Default | Default | 1 | Yes |
| (Galaxies) All | Daily | Default | Default | 1 | Yes |
| (Snyk) All | Daily | Default | Default | 9 | Yes |

[^1]: The default value is 500000. See https://www.cloudquery.io/docs/reference/source-spec#concurrency.
[^2]: The default value is 512MB.
[^3]: This is being terminated due to OOM issues.
[^4]: This isn't being terminated due to memory issues, I suspect there's a hard limit somewhere.
[^5]: We see GH API rate limiting.

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -115,18 +115,6 @@ export class CloudQuery extends GuStack {
 
 		const awsSources: CloudquerySource[] = [
 			{
-				name: 'All',
-				description: 'Data fetched across all accounts in the organisation.',
-				schedule: Schedule.rate(Duration.days(1)),
-				config: awsSourceConfigForOrganisation({
-					tables: ['aws_*'],
-					skipTables: skipTables,
-				}),
-				managedPolicies: [readonlyPolicy],
-				policies: [standardDenyPolicy, cloudqueryAccess('*')],
-				memoryLimitMiB: 2048,
-			},
-			{
 				name: 'DeployToolsListOrgs',
 				description:
 					'Data fetched from the Deploy Tools account (delegated from Root).',
@@ -205,6 +193,23 @@ export class CloudQuery extends GuStack {
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
 		];
+
+		const selectedAwsTables = awsSources.flatMap(
+			(_) => _.config.spec.tables,
+		) as string[];
+
+		const allAws: CloudquerySource = {
+			name: 'All',
+			description: 'Data fetched across all accounts in the organisation.',
+			schedule: Schedule.rate(Duration.days(1)),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_*'],
+				skipTables: [...skipTables, ...selectedAwsTables],
+			}),
+			managedPolicies: [readonlyPolicy],
+			policies: [standardDenyPolicy, cloudqueryAccess('*')],
+			memoryLimitMiB: 2048,
+		};
 
 		const githubCredentials = SecretsManager.fromSecretPartialArn(
 			this,
@@ -355,6 +360,7 @@ export class CloudQuery extends GuStack {
 			db,
 			dbAccess: applicationToPostgresSecurityGroup,
 			sources: [
+				allAws,
 				...awsSources,
 				...githubSources,
 				...fastlySources,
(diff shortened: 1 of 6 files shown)


akash1810:
This is sort of working, in that the task was force stopped with the reason:

```log
Exit code: 1
OutOfMemoryError: Container killed due to memory usage
```

packages/cdk/lib/cloudquery.ts
```diff
@@ -182,8 +171,61 @@ export class CloudQuery extends GuStack {
...
+			description: 'Data fetched across all accounts in the organisation.',
+			schedule: Schedule.rate(Duration.days(1)),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_*'],
+				skipTables: [...skipTables, ...selectedAwsTables],
+				concurrency: 10000,
+			}),
+			managedPolicies: [readonlyPolicy],
+			policies: [standardDenyPolicy, cloudqueryAccess('*')],
+			memoryLimitMiB: 2048,
```
akash1810, line 226: A somewhat random choice. The default is 512, so providing 4x memory feel quite generous. Unsure how to determine how much of this ends up being used.

packages/cdk/lib/cloudquery.ts
```diff
@@ -182,8 +171,61 @@ export class CloudQuery extends GuStack {
...
+					concurrency: 10000,
+				}),
+				managedPolicies: [readonlyPolicy],
+				policies: [standardDenyPolicy, cloudqueryAccess('*')],
+			},
 		];
 
+		const selectedAwsTables = awsSources.flatMap(
+			(_) => _.config.spec.tables,
+		) as string[];
```
akash1810, line 213: These tables are being collected especially, we do not need to collect them again in the all task; we add `selectedAwsTables` to the all task's `skipTables`.

akash1810:
Closing in favour of https://github.com/guardian/service-catalogue/pull/220.

PR #205, opened 2023-06-05
feat: Collect load balancers, and CW Alarms on their own

akash1810:
## What does this change?
This PR creates a discrete task to collect data on load balancers and CloudWatch Alarms (tables `aws_elbv1_*`, `aws_elbv2_*` and `aws_cloudwatch_alarms`).

## Why?
Currently, the "all AWS" task is failing (see #203), and as a result, these tables are empty/stale. Collecting the data in their own tasks should mean the data is kept more up-to-date, and therefore is more reliable. This is particularly important as these tables are used by the @guardian/devx-reliability team.

## How has it been verified?
I've [deployed this](https://riffraff.gutools.co.uk/deployment/view/0141d7df-0d32-408a-8acf-9acf3c1ee163) and manually triggered the task, and can see the tables filling up.

### Before
image

### After
image

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -192,6 +192,17 @@ export class CloudQuery extends GuStack {
 				managedPolicies: [readonlyPolicy],
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
+			{
+				name: 'OrgWideCloudwatchAlarms',
+				description:
+					'Collecting CloudWatch Alarm data across the organisation.',
+				schedule: Schedule.rate(Duration.days(1)),
+				config: awsSourceConfigForOrganisation({
+					tables: ['aws_cloudwatch_alarms'],
+				}),
+				managedPolicies: [readonlyPolicy],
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -182,6 +182,27 @@ export class CloudQuery extends GuStack {
 				managedPolicies: [readonlyPolicy],
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
+			{
+				name: 'OrgWideLoadBalancers',
+				description: 'Collecting load balancer data across the organisation.',
+				schedule: Schedule.rate(Duration.days(1)),
+				config: awsSourceConfigForOrganisation({
+					tables: ['aws_elbv1_*', 'aws_elbv2_*'],
```
akash1810, line 190: Rather than list the tables individually, use [wildcard matching](https://www.cloudquery.io/docs/advanced-topics/performance-tuning#use-wildcard-matching). This way any new tables in this "namespace" are picked up automatically.

jacobwinch (approved):
Thanks for doing this!

PR #209, opened 2023-06-07
fix: Collect Security Hub data once

akash1810:
## What does this change?
We've setup Security Hub (SH) as a delegated service in the Security account. This means the Security account can see SH results for all accounts, and we do not need to collect the data again from each member account.

## Why?
Avoids duplicated data, which inflates numbers.

## How has it been verified?
I've manually deployed, and run this task. Using Grafana, we can observe the database is being updated, however the task is also being terminated by AWS after ~61 minutes. This behaviour is similar to that observed in #203. I cannot find anything to suggest this duration is configurable, will raise a support ticket with AWS.

The data we do have indicates there might be some optimisation we can do regarding `ARCHIVED` findings. If we remove these, would it take less time to collect the data?

```sql
select      ac.name
            , shf.record_state
            , count(shf.id)
from        aws_securityhub_findings shf
            left join   aws_accounts ac
            on          ac.id = shf.aws_account_id
group by    ac.name
            , shf.record_state
order by    count(shf.id) desc;
```

| name | record_state | count |
|---|---|---|
| Frontend | ARCHIVED | 134181 |
| Guardian Mobile Team | ARCHIVED | 43245 |
| Guardian Developer Playground | ARCHIVED | 43109 |
| Deploy Tools | ARCHIVED | 43099 |
| Aws-Investigations | ARCHIVED | 24380 |
| Composer | ARCHIVED | 23167 |
| Guardian AWS Membership | ARCHIVED | 20540 |
| Media-API-GNM | ARCHIVED | 19056 |
| Guardian Content API | ARCHIVED | 11405 |
| gnm-multimedia | ARCHIVED | 9912 |
| Guardian Identity | ARCHIVED | 7720 |
| Composer | ACTIVE | 6022 |
| CMS Workflow | ARCHIVED | 4260 |
| Guardian AWS Membership | ACTIVE | 3765 |
| guardian discussion team | ARCHIVED | 3485 |
| Guardian Content API | ACTIVE | 3465 |
| Frontend | ACTIVE | 3424 |
| Aws-Investigations | ACTIVE | 3303 |
| Media-API-GNM | ACTIVE | 2866 |
| Guardian Mobile Team | ACTIVE | 2809 |
| Deploy Tools | ACTIVE | 2804 |
| CMS Fronts | ARCHIVED | 2543 |
| Guardian Developer Playground | ACTIVE | 2475 |
| aws-baton | ARCHIVED | 1574 |
| guardian discussion team | ACTIVE | 1565 |
| gnm-multimedia | ACTIVE | 1538 |
| Guardian Identity | ACTIVE | 1530 |
| Guardian Security | ARCHIVED | 1412 |
| CMS Workflow | ACTIVE | 1139 |
| Infrastructure AWS Services | ACTIVE | 1095 |
| Guardian Security | ACTIVE | 985 |
| GNM InfoSec | ACTIVE | 888 |
| CMS Fronts | ACTIVE | 770 |
| Ophan | ACTIVE | 699 |
| aws-targeting | ACTIVE | 571 |
| GNM InfoSec | ARCHIVED | 497 |
| Guardian Websys General PROD | ACTIVE | 486 |
| aws-baton | ACTIVE | 444 |
| aws-data-tech | ACTIVE | 396 |
| aws-domains | ACTIVE | 360 |
| aws-domains | ARCHIVED | 347 |
| Guardian Websys General DEV | ACTIVE | 312 |
| Infrastructure AWS Services | ARCHIVED | 233 |
| aws-targeting | ARCHIVED | 94 |
| Guardian Websys General PROD | ARCHIVED | 60 |
| aws-data-tech | ARCHIVED | 56 |
| Guardian Websys General DEV | ARCHIVED | 53 |
| Guardian Digital Interactives PROD | ACTIVE | 42 |
| Ophan | ARCHIVED | 36 |
| Secure Collaboration | ACTIVE | 18 |
| editorial-feeds | ACTIVE | 17 |
| Personalisation | ACTIVE | 17 |
| Guardian US Interactives | ACTIVE | 2 |
| Editorial Systems Development | ACTIVE | 2 |
| Data Science Team | ACTIVE | 2 |
| Guardian ESD | ACTIVE | 1 |
| Guardian Network Team | ACTIVE | 1 |
| Guardian Print Sites | ACTIVE | 1 |
| Guardian GLABS | ACTIVE | 1 |

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -167,16 +167,12 @@ export class CloudQuery extends GuStack {
 				],
 			},
 			{
-				name: 'SecurityAccessAnalyser',
+				name: 'DelegatedToSecurityAccount',
 				description:
-					'Data fetched from the Security account. Note, Access Analyzer collects data from our entire organisation so we only need to query it in one place.',
+					'Collecting data across the organisation from services delegated to the Security account.',
 				schedule: Schedule.rate(Duration.days(1)),
 				config: awsSourceConfigForAccount(GuardianAwsAccounts.Security, {
-					tables: [
-						'aws_accessanalyzer_analyzers',
-						'aws_accessanalyzer_analyzer_archive_rules',
-						'aws_accessanalyzer_analyzer_findings',
-					],
+					tables: ['aws_accessanalyzer_*', 'aws_securityhub_*'],
 				}),
 				managedPolicies: [readonlyPolicy],
 				policies: [
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -629,6 +629,567 @@ spec:
       },
       "Type": "AWS::IAM::Policy",
     },
+    "CloudquerySourceDelegatedToSecurityAccountScheduledEventRuleC7320B4E": {
+      "Properties": {
+        "ScheduleExpression": "rate(1 day)",
+        "State": "ENABLED",
+        "Targets": [
+          {
+            "Arn": {
+              "Fn::GetAtt": [
+                "cloudqueryCluster5370C11B",
+                "Arn",
+              ],
+            },
+            "EcsParameters": {
+              "LaunchType": "FARGATE",
+              "NetworkConfiguration": {
+                "AwsVpcConfiguration": {
+                  "AssignPublicIp": "DISABLED",
+                  "SecurityGroups": [
+                    {
+                      "Fn::GetAtt": [
+                        "PostgresAccessSecurityGroupCloudqueryE959A23F",
+                        "GroupId",
+                      ],
+                    },
+                  ],
+                  "Subnets": {
+                    "Ref": "PrivateSubnets",
+                  },
+                },
+              },
+              "TaskCount": 1,
+              "TaskDefinitionArn": {
+                "Ref": "CloudquerySourceDelegatedToSecurityAccountTaskDefinitionD407788D",
+              },
+            },
+            "Id": "Target0",
+            "Input": "{}",
+            "RoleArn": {
+              "Fn::GetAtt": [
+                "CloudquerySourceDelegatedToSecurityAccountTaskDefinitionEventsRoleA78E8779",
+                "Arn",
+              ],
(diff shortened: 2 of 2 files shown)


packages/cdk/lib/cloudquery.ts
```diff
@@ -149,16 +149,12 @@ export class CloudQuery extends GuStack {
...
+				name: 'DelegatedToSecurityAccount',
 				description:
-					'Data fetched from the Security account. Note, Access Analyzer collects data from our entire organisation so we only need to query it in one place.',
+					'Collecting data across the organisation from services delegated to the Security account.',
 				schedule: Schedule.rate(Duration.days(1)),
 				config: awsSourceConfigForAccount(GuardianAwsAccounts.Security, {
-					tables: [
-						'aws_accessanalyzer_analyzers',
-						'aws_accessanalyzer_analyzer_archive_rules',
-						'aws_accessanalyzer_analyzer_findings',
```
akash1810, line 160: [Wildcard matching](https://www.cloudquery.io/docs/advanced-topics/performance-tuning#use-wildcard-matching) makes this a little easier. There are only [three `aws_accessanalyzer` tables](https://www.cloudquery.io/docs/plugins/sources/aws/tables), so this change can be viewed as a no-op.

nicl: In this case, given there are only three tables, I wonder if wildcard matching is not so helpful? Is it better to be explicit whenever we can be?

Something else here, is this really no-op? `aws_securityhub_*` looks like an addition and the snapshot suggests other changes too that I'm not sure I understand.

nicl (commented):
(Agree in principle, just some questions about the table changes.)

packages/cdk/lib/cloudquery.ts
akash1810: Ah, apologies my wording wasn't especially clear here. The no-op is the change of the three explicit tables to the `aws_accessanalyzer_*` wildcard, rather than the entire change.

I think wildcards are better for the single reason as there are fewer LOC, but happy to be persuaded otherwise.

akash1810:
Closing in favour of https://github.com/guardian/service-catalogue/pull/221.

PR #212, opened 2023-06-07
chore: Remove RepoCop 😢

akash1810:
## What does this change?
RepoCop as a service uses the [GitHub Lens API as its source of data](https://github.com/guardian/service-catalogue/blob/3d646a56b2b44e3734c96c8474bd6212f9ce9e4c/packages/repocop/src/main/scala/GHLensAPI.scala#L16-L20). In https://github.com/guardian/service-catalogue/pull/202, we deleted the GitHub Lens API. This means in its current form, RepoCop is failing to run.

Remove the service for a later day.

## How has it been verified?
Erm... does `dig` count? Using `dig` we can confirm that the GitHub Lens API no longer exists, and by extension assert RepoCop is failing to run.

```sh
dig CNAME +short github-lens.gutools.co.uk
```

## TODO after merge
- [ ] Delete the CloudFormation stack in AWS


ADR/01-x01-repocop-and-vulnerability-api-arch.md
akash1810, line 1: A new file, as it's an edit. Naming it `01-x01` to have it appear nicely in file trees.

nicl (approved):

PR #213, opened 2023-06-07
fix(cloudquery): Remove `aws_inspector_findings` table

tjsilver:
## What does this change?

Removes the `aws_inspector_findings` from the `OrgWideInspector` in Cloudquery. We found in testing that the `aws_inspector2_findings` were sufficient. This avoids duplication.


akash1810 (approved):

tjsilver:
Closing as we might need the inspector 1 findings after all, while debugging inspector-lambda.

PR #215, opened 2023-06-12
Use custom plugin to fetch snyk projects

NovemberTang:
## What does this change?

Add [custom snyk plugin](https://github.com/guardian/cq-source-snyk-full-project) as a cloudquery source. This means we are now able to pull project tags and the aggregated issues, grouped by severity.

## Why?

This PR means we are more easily able to join Snyk data with other sources of information from GitHub or AWS. We are also able to see more about the state of a project at a glance, without needing to join with other tables.

## How has it been verified?

Running the changes locally using docker, we found that all the new information was collected. The change has been deployed and runs successfully

### DIFF
--- a/packages/cdk/lib/ecs/versions.ts
+++ b/packages/cdk/lib/ecs/versions.ts
@@ -48,8 +48,8 @@ export const Versions = {
 	 */
 	CloudquerySnyk: '3.0.0',
 
-	/** 
-  	 * The version of the CloudQuery Snyk source plugin to install.
+	/**
+	 * The version of the CloudQuery Snyk source plugin to install.
 	 *
 	 * @see https://github.com/guardian/cq-source-snyk-full-project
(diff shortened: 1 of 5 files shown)


packages/cdk/lib/ecs/versions.ts
```diff
@@ -47,4 +47,6 @@ export const Versions = {
 	 * @see https://github.com/cloudquery/cloudquery/releases?q=plugins-source-snyk
 	 */
 	CloudquerySnyk: '3.0.0',
+
+	CloudquerySnykGuardian: '0.1.1',
```
akash1810, line 51: Worth adding a `@see https://github.com/guardian/cq-source-snyk-full-project` doc string here to make it easier to find the repository in future?

akash1810 (approved):

PR #217, opened 2023-06-14
docs: Add docs for CloudQuery adoption

akash1810:
## What does this change?
Adds an ADR to the repository to document the decision to adopt CloudQuery. This is largely an adaptation of the [original proposal document](https://docs.google.com/document/d/1AJGK4HjHf6znEXRGqbbF5gYaq2PmmcrJjsX0feKXd0I/edit).

Also adds an "implementation" document, primarily as a way to have the SQL we've manually executed tracked in VCS.

This is part 1 of a series of changes to document this repository. Other documentation should cover:
- How to manually invoke an ECS task
- How to debug configuration
- How to query the database locally
- Anything else...

## Why?
ADRs help provide context for our future selves.

## How has it been verified?
By reading the documents, and checking it's an accurate reflection of events.

---

Closes https://github.com/guardian/service-catalogue/pull/182.

### DIFF
--- a/ADR/02-cloudquery.md
+++ b/ADR/02-cloudquery.md
@@ -1,0 +1,71 @@
+# CloudQuery Adoption
+
+## Status
+Accepted, and implemented.
+
+## Context
+The Service Catalogue project aims to "allow easy exploration of Guardian services", allowing one to answer questions such as:
+- What services to I (and my team) own?
+- Which repository do services come from?
+- Which services follow DevX best practice/use tooling?
+
+Initially, the Service Catalogue was architected in two parts:
+1. Data collection
+   ```mermaid
+   flowchart LR
+   A["Source(s)"] --> B[Data collector]
+   B --> C[(DataStore)]
+   ```
+2. Data presentation
+   ```mermaid
+   flowchart LR
+   A[(DataStore)] --> B[Lens API]
+   B --> C[Service Catalogue API]
+   C --> D((User))
+   ```
+
+That is, for each data source, we'd build:
+- A data collector
+- A data aggregation API (informally known as a Lens API)
+- Public-facing APIs (informally known as the Service Catalogue API)
+
+To present a piece of data in the Service Catalogue API, we would:
+- Create a data collector, integrating with third-party APIs, and writing results to a data store
+- Create a Lens API that aggregates the data in the store
+- Service Catalogue API endpoints to expose the Lens API to the user
+
+Whilst this pattern works, it results in complex[^1] code that needs to be maintained over time.
+
+## Proposal
+[CloudQuery](https://www.cloudquery.io/) is an open-source project to:
+
+> Sync any source to any destination, transform and visualize.
+
+CloudQuery can collect data from AWS, GitHub, Snyk, and more. 
+It can also store the data in a variety of destinations, including Postgres.
+
+Using CloudQuery would allow us to replace the first diagram above, 
+with the data collection being handled by CloudQuery, and the data store being Postgres.
+Using Postgres would allow us to use Grafana to visualise the data.
+
+## Consequences
+At a high-level, CloudQuery is a data collection tool.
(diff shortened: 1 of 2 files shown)


ADR/02-cloudquery.md
```diff
@@ -0,0 +1,80 @@
...
+Whilst this pattern works, it results in complex[^1] code that needs to be maintained over time.
+
+## Proposal
+[CloudQuery](https://www.cloudquery.io/) is an open-source project that can:
+
+> Sync any source to any destination, transform and visualize.
+
+Adopting CloudQuery would replace the first diagram above, and with its ability to store data in Postgres, we can replace the second diagram with Grafana dashboards.
+
+## Implementation
```
akash1810, line 46: Not convinced this implementation section belongs in the ADR.

JuliaBrigitte: Thank you for documenting this and thinking of our future selves.
The only thing that confused me was
"Sync any source to any destination, transform and visualize"
I did google and I think you have to connect to a tools of you choice like grafana to visualize, might it be worth to spell that out?

JuliaBrigitte (dismissed):

ADR/02-cloudquery.md
```diff
@@ -0,0 +1,71 @@
...
+- Public-facing APIs (informally known as the Service Catalogue API)
+
+To present a piece of data in the Service Catalogue API, we would:
+- Create a data collector, integrating with third-party APIs, and writing results to a data store
+- Create a Lens API that aggregates the data in the store
+- Service Catalogue API endpoints to expose the Lens API to the user
+
+Whilst this pattern works, it results in complex[^1] code that needs to be maintained over time.
+
+## Proposal
```
JuliaBrigitte, line 39: Since it is implemented should this read Original Proposal?

docs/cloudquery.md
```diff
@@ -0,0 +1,64 @@
...
+Each ECS task authenticates with Postgres using [IAM Authentication](https://repost.aws/knowledge-center/rds-postgresql-connect-using-iam).
+
+The root (sudo) user is not used. Instead, we manually created a user:
+
+```sql
+-- Create user for CloudQuery, and grant RDS IAM authentication.
+-- See https://repost.aws/knowledge-center/rds-postgresql-connect-using-iam
+CREATE USER cloudquery;
+GRANT rds_iam TO cloudquery;
+```
```
JuliaBrigitte, line 39: I think the cloudquery user is read only and I think this text means you can query the db from your db with your janus credentials, but to be quite honest, I would not have known from reading this if you had not recently set this up for me.

JuliaBrigitte (approved):

ADR/02-cloudquery.md
akash1810: Ah interesting, I've always seen ADRs as being future-tense, with the `status` section describing where we're at.

docs/cloudquery.md
akash1810: The `cloudquery` user has read and write permissions to the database.

I think we should include some information about how to connect to the database locally, however I'm not too sure where this should live. Is it in this document? Or in its own?

ADR/02-cloudquery.md
```diff
@@ -0,0 +1,71 @@
...
+[CloudQuery](https://www.cloudquery.io/) is an open-source project to:
+
+> Sync any source to any destination, transform and visualize.
+
+CloudQuery can collect data from AWS, GitHub, Snyk, and more. 
+It can also store the data in a variety of destinations, including Postgres.
+
+Using CloudQuery would allow us to replace the first diagram above, 
+with the data collection being handled by CloudQuery, and the data store being Postgres.
+Using Postgres would allow us to use Grafana to visualise the data.
```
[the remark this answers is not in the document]
akash1810: Similar to above, I wonder if the `status` section covers this?

PR #218, opened 2023-06-15
chore(deps): Bump the version of CloudQuery CLI, sources, and destination

akash1810:
## What does this change?
Bumps the versions of CloudQuery things being used.

The AWS and GitHub source plugins see a major version update, i.e. a breaking change. Looking at the notes for the [AWS plugin](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-aws-v18.0.0) and the [GitHub plugin](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-github-v6.0.0), the kind of breaking change would not impact us.


tjsilver (approved):
👍

PR #220, opened 2023-06-20
fix: Allow long running tasks to complete

akash1810:
## What does this change?
IAM authentication for RDS has the following [behaviour](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.html):
  - The generated password expires after 15 minutes
  - A connection established within 15 minutes is valid for as long as the connection is kept
  - Attempts to establish a connection after 15 minutes are rejected

Most of the CloudQuery tasks finish well before 15 minutes[^1], however there are some that take over 2 hours (see https://github.com/guardian/service-catalogue/pull/209).

We're seeing these tasks fail, with a log line of:

```log
failed to connect to `host=REDACTED user=cloudquery database=postgres`: server error (FATAL: PAM authentication failed for user \"cloudquery\" (SQLSTATE 28000))
```

It appears that CloudQuery does not hold a single connection, but instead establishes multiple, and the latter ones are made after 15 minutes.

Temporarily revert to static (root) credentials to get these long running tasks functional again.

Additionally, I observed the default of 500MB memory and 0.25 vCPU quickly resulted in the task experiencing an OOM and terminating. To correct this, the "all AWS" task has:
  - A [`concurrency` setting](https://www.cloudquery.io/docs/reference/source-spec#concurrency) of 2000
  - 2GB memory
  - 1 vCPU

These were selected entirely at random. I'd imagine there are some observability metrics we can use in the future to tune this.

## How has it been verified?
I've [deployed](https://riffraff.gutools.co.uk/deployment/view/2a929c91-8368-4378-8451-606753eec403) this branch, manually started a long running job (the "all AWS" one), and observed it run to completion in 3h 6m 25s[^2].

## TODO (async)
- [ ] Raise an issue/PR within CloudQuery

[^1]: See https://github.com/guardian/service-catalogue/pull/203 for a duration breakdown
[^2]: Between 21/06/2023, 2:55:31 UTC and 21/06/2023, 6:01:56 UTC

### DIFF
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -254,6 +254,10 @@ export class CloudQuery extends GuStack {
 			}),
 			managedPolicies: [readonlyPolicy],
 			policies: [standardDenyPolicy, cloudqueryAccess('*')],
+
+			// This task is quite expensive, and requires more power than the default (500MB memory, 0.25 vCPU).
+			memoryLimitMiB: 2048,
+			cpu: 1024,
 		};
 
 		const githubCredentials = SecretsManager.fromSecretPartialArn(
--- a/packages/cdk/lib/cloudquery.ts
+++ b/packages/cdk/lib/cloudquery.ts
@@ -131,18 +131,7 @@ export class CloudQuery extends GuStack {
 			'readonly-managed-policy',
 		);
 
-		const awsSources: CloudquerySource[] = [
-			{
-				name: 'All',
-				description: 'Data fetched across all accounts in the organisation.',
-				schedule: Schedule.rate(Duration.days(1)),
-				config: awsSourceConfigForOrganisation({
-					tables: ['aws_*'],
-					skipTables: skipTables,
-				}),
-				managedPolicies: [readonlyPolicy],
-				policies: [standardDenyPolicy, cloudqueryAccess('*')],
-			},
+		const individualAwsSources: CloudquerySource[] = [
 			{
 				name: 'DeployToolsListOrgs',
 				description:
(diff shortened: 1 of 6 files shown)


packages/cdk/lib/ecs/task.ts
```diff
@@ -201,7 +164,8 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 
 		managedPolicies.forEach((policy) => task.taskRole.addManagedPolicy(policy));
 		policies.forEach((policy) => task.addToTaskRolePolicy(policy));
-		db.grantConnect(task.taskRole, dbUser);
+
+		db.grantConnect(task.taskRole);
```
akash1810, line 168: The `dbUser` is removed as the default is to use the root user.

packages/cdk/lib/ecs/task.ts
```diff
@@ -107,36 +102,23 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
...
-		});
-
-		const cloudqueryTask = task.addContainer(`${id}Container`, {
+		task.addContainer(`${id}Container`, {
 			image: cloudqueryImage,
 			entryPoint: [''],
-			secrets,
+			secrets: {
+				...secrets,
+				DB_USERNAME: Secret.fromSecretsManager(db.secret, 'username'),
```
nicl, line 118: Is this secret (and format) automatically created by the DB construct? (I assume yes but the TS docs aren't very clear on this.)

And do we need to remove the `iamAuthentication: true` bit from our `DatabaseInstanceProps` in `cloudquery.ts` temporarily too?

nicl (approved):
Looking forward to fresh data :) 🥳

packages/cdk/lib/ecs/task.ts
akash1810: > Is this secret (and format) automatically created by the DB construct?

Yes, the secret is [provided by default](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#credentials).

> And do we need to remove the `iamAuthentication: true` bit from our DatabaseInstanceProps in `cloudquery.ts` temporarily too?

I'd like to keep it as the mechanism for connecting to the database locally. I wonder if https://github.com/guardian/service-catalogue/pull/217 (or similar) should document this process?

packages/cdk/lib/cloudquery.ts
```diff
@@ -243,6 +232,30 @@ export class CloudQuery extends GuStack {
...
+					// casting because `config.spec.tables` could be empty, though in reality it never is
+					...(individualAwsSources.flatMap(
+						(_) => _.config.spec.tables,
+					) as string[]),
+				],
+
+				// Defaulted to 500000 by CloudQuery, concurrency controls the maximum number of Go routines to use.
+				// The amount of memory used is a function of this value.
+				// See https://www.cloudquery.io/docs/reference/source-spec#concurrency.
+				concurrency: 2000,
```
akash1810, line 253: A somewhat random number! Using the default of 500000 here quickly results in an OOM error.

akash1810:
Looks like CloudQuery are working on a single write destination plugin (https://github.com/cloudquery/cloudquery/releases/tag/plugins-destination-postgresql-v5.1.0-single-write) which might help here.

PR #221, opened 2023-06-21
fix: Collect Security Hub data once

akash1810:
This is a rework of https://github.com/guardian/service-catalogue/pull/209.

## What does this change?
We've setup Security Hub as a delegated service in the Security account. This means that account can see Security Hub results for all accounts in the organisation. Collecting the data from each member account is redundant, as it is duplication.

This change collects Access Analyzer and Security Hub data _once_, from the Security account.

This should also reduce the running time of the "all AWS" task (see #220).

## How has it been verified?
I've [deployed](https://riffraff.gutools.co.uk/deployment/view/d4771128-9c33-4714-978b-8032f39957ca) this branch, manually run the task, and observed it run to completion in ~50 mins[^1].

[^1]: Between 21/06/2023, 8:55:56 UTC and 21/06/2023, 9:47:04 UTC

### DIFF
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -139,9 +139,8 @@ spec:
     - aws_organizations_organizational_units
     - aws_organizations_policies
     - aws_organizations_roots
-    - aws_accessanalyzer_analyzers
-    - aws_accessanalyzer_analyzer_archive_rules
-    - aws_accessanalyzer_analyzer_findings
+    - aws_accessanalyzer_*
+    - aws_securityhub_*
     - aws_cloudformation_stacks
     - aws_cloudformation_stack_resources
     - aws_cloudformation_stack_templates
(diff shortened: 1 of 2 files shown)


nicl (commented):
Do we need to remove the tables from the AWS all accounts task too as part of this? At the moment, the only changes I can see are to refactor the task name/description, introduce aws_securityhub_* collection from the security account, and add some resource limits (cpu/mem).

akash1810:
> Do we need to remove the tables from the AWS all accounts task too as part of this?

This should happen automatically due to a combination of the ["all AWS" task's configuration](https://github.com/guardian/service-catalogue/blob/c84abb43bca4ac282cf5c3593de90e68a3062b70/packages/cdk/lib/cloudquery.ts#L244-L248), and the [`write_mode`](https://www.cloudquery.io/docs/reference/destination-spec#write_mode).

> Also do you have any thoughts on why the snapshot diff is so large - is it just the result of the name change?

The rename is the cause of this, yes. Not ideal for a review 😞.

nicl:
@akash1810 ah awesome, hadn't seen the updated skip tables logic 👍 .

nicl (approved):
LGTM!

PR #222, opened 2023-06-21
Add task to ingest github_issues

nicl:
https://www.cloudquery.io/docs/plugins/sources/github/tables/github_issues

Used to understand PR activity, and especially bot activity (such as Dependabot and Scala Steward).

Initially run daily as PRs and Issues change frequently.


akash1810 (approved):
Nice! I feel like this'll be a lot of data, and might put us close to/over the API quota. Only way to find out is to merge 😄 .

nicl:
@akash1810 thanks. Will keep an eye - agree hard to tell impact until we run it!

PR #223, opened 2023-07-04
Add docs for people using cloudquery for the first time

NovemberTang:
## What does this change?

Adds a getting started guide for people who are unfamiliar with cloudquery.
Adds some information about what data people can provide to make cloudquery better/more complete

## Why?

All of the information people need to answer our OKR questions now exists, but it's not clear how they should access it or start using it. Hopefully this will provide some clarity, and give people enough information that they are able to start using cloudquery independently.

This information should be accessible to a minimally technical audience, such as a brand new engineer or product manager.


jorgeazevedo:
This is really great! May I ask how you made that gif? It's a great touch

docs/getting-started.md
```diff
@@ -0,0 +1,93 @@
...
+
+CloudQuery is only as good as the data we have, and the links between them. If you are following DevX best practice
+consistently, this shouldn't be too much of a problem. However, if you are not, you may find that some of the data you
+are expecting to see is missing, as we rely heavily on particular fields for joins.
+
+To make sure your team's CloudQuery information is as complete as it can be, please complete as many of these steps as
+is practical:
+
+- Your team email and github team slug
+  are [up-to-date in galaxies](https://github.com/guardian/galaxies/blob/main/shared/data/teams.ts)
```
akash1810, line 24: nit: `Galaxies` should be capitalised.

docs/getting-started.md
```diff
@@ -0,0 +1,93 @@
...
+
+CloudQuery generally collects metadata about these resources. That means that we have information like the name of a
+repository, and when it was created, but not the contents of the files in it.
+
+CloudQuery is only as good as the data we have, and the links between them. If you are following DevX best practice
+consistently, this shouldn't be too much of a problem. However, if you are not, you may find that some of the data you
+are expecting to see is missing, as we rely heavily on particular fields for joins.
+
+To make sure your team's CloudQuery information is as complete as it can be, please complete as many of these steps as
+is practical:
```
akash1810, line 21: Worth linking to https://github.com/guardian/recommendations/blob/main/best-practices.md at the moment, even though it's not complete? My thinking is we can say something like:

> If you're following the best practice, then you'll have the easiest time.

(Worded more eloquently though!)

akash1810 (commented):
Couple of non-blocking comments.

docs/getting-started.md
NovemberTang: yes, definitely!! we should also add the galaxies stuff to the table, as i don't think it's been written down anywhere yet

akash1810 (approved):

PR #224, opened 2023-07-05
Alert if a cluster task fails

NovemberTang:
## What does this change?

If a scheduled CloudQuery task fails, we are not notified. This change introduces an EventBridge rule, where if the container stops with an error code, we get a message telling us which, when, and why. Currently, this message is a rather nasty JSON blob, but it does contain all the information required to figure out what is going on.

Further work should be done in a subsequent PR to turn the alert message into something a bit nicer, maybe we could send the message to a lambda that reformats it and passes it along to Anghammarad. But this is a significant amount of scope creep and needs to be prioritised against our other work. For now (aka the next couple of days), it meets the minimum requirements

## Why?

We need to know if our data is stale, as we rely on it for OKR purposes. We also need to provide up to date information to other teams.

## How has it been verified?

I swapped around some of the DB credentials to force a failure, and attached my internal email to the alert topic. Since deployment, every attempted run of a container has failed and sent me a message

### DIFF
--- a/packages/cdk/lib/ecs/task.ts
+++ b/packages/cdk/lib/ecs/task.ts
@@ -12,9 +12,12 @@ import {
 } from 'aws-cdk-lib/aws-ecs';
 import type { ScheduledFargateTaskProps } from 'aws-cdk-lib/aws-ecs-patterns';
 import { ScheduledFargateTask } from 'aws-cdk-lib/aws-ecs-patterns';
+import { Rule } from 'aws-cdk-lib/aws-events';
+import { SnsTopic } from 'aws-cdk-lib/aws-events-targets';
 import type { IManagedPolicy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
 import { RetentionDays } from 'aws-cdk-lib/aws-logs';
 import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
+import type { Topic } from 'aws-cdk-lib/aws-sns';
 import { dump } from 'js-yaml';
 import type { CloudqueryConfig } from './config';
 import { postgresDestinationConfig } from './config';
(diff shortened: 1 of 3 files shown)


akash1810 (commented):
cc @guardian/devx-reliability in case they have opinions on alerting

packages/cdk/lib/ecs/task.ts
```diff
@@ -66,6 +69,11 @@ export interface ScheduledCloudqueryTaskProps
 	 */
 	sourceConfig: CloudqueryConfig;
 
+	/**
+	 * The topic used to notify someone if a task f
```
jacobwinch, line 73: Minor:

```suggestion
	 * The topic used to notify someone if a task fails
```

jacobwinch:
> cc @guardian/devx-reliability in case they have opinions on alerting

I'm not sure that anyone in the team has much experience configuring alerting with ECS (feel free to correct me if I'm wrong @michaelwmcnamara @jorgeazevedo!). The only other approach that I'm aware of is [using a Step Function to wrap the task and then alerting based on Step Function failures](https://github.com/guardian/cdk/blob/1b026c30a54ed3483bf5a2b3bf5a770c768ade69/src/patterns/scheduled-ecs-task.ts#L21-L31).

However, it looks like the infrastructure that you're using is a bit different from this and it sounds like this has been tested thoroughly so 👍 from my perspective.

packages/cdk/lib/ecs/task.ts
```diff
@@ -173,6 +182,29 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
...
 		});
 
+		new Rule(scope, `${id}-TaskErrorRule`, {
+			description:
+				'Rule for events indicating an ECS task exited due to an error.',
+			eventPattern: {
+				detail: {
+					clusterArn: [cluster.clusterArn],
+					containers: {
+						exitCode: [
```
akash1810, line 192: I think in the ideal, we'd be able to say "non 0" to catch all scenarios, but that doesn't seem to be supported here 😢 .

akash1810 (approved):
img

PR #227, opened 2023-07-10
fix(github): Stop collecting `github_repository_dependabot_alerts`

akash1810:
## What does this change?
We're not (yet) interested in this table, and it takes a long time to crawl.

## Why?
The `GitHubRepositories` task runs once a day, and with `github_repository_dependabot_alerts` being collected, the task takes over 24 hours to run!

image

## How has it been verified?
N/A


akash1810:
The `GitHubRepositories` task now takes <4 minutes to complete[^1]. The total number of repositories is also correct now too!

```sql
select count(*) from github_repositories;
-- 3300 previously
-- 3389 now
```

image

[^1]: Started 10/07/2023, 9:23:25 UTC. Completed 10/07/2023, 9:26:53 UTC

PR #229, opened 2023-07-10
feat: Initialise `github-recently-updated` lambda

akash1810:
## What does this change?
Adds the initial CI/CD structure for a new AWS lambda, called `github-recently-updated`. It takes inspiration from [tracker](https://github.com/guardian/tracker/tree/main/packages/example-lambda).

This is not a full implementation of the lambda's logic, but just the infrastructure. This is to keep the change small, and easier to review.

This _could_ be [built as a container](https://www.docker.com/blog/9-tips-for-containerizing-your-node-js-application/) and run in ECS too. The benefits would be that it could simply be added as a task before the CloudQuery step. The setup for this is a little more involved, so opted against this for now. Opinions wanted!

## Why?
Our default CloudQuery configuration is to collect data for all repositories in the GitHub organisation.

There are some cases where this leads to a long task execution. For example, the [`github_repository_branches`](https://www.cloudquery.io/docs/plugins/sources/github/tables/github_repository_branches) table 
will collect data for all branches, in all repositories. That's a lot of data!

This lambda takes inspiration from https://github.com/guardian/service-catalogue/pull/131, and will ultimately write a list of recently updated repositories to AWS, for further use by CloudQuery tasks.

It'll run:

```sql
select      full_name
            , pushed_at
            , updated_at
from        github_repositories
where       greatest(pushed_at, updated_at) > NOW() - INTERVAL '1 days';
```

This will return a list of repositories that have been updated in the last 24 hours[^1], a subset of all repositories in the organisation[^2]. This list can then be used in CloudQuery tasks to limit the scope of data collection.

## How has it been verified?
I've deployed this branch, manually ran the lambda, and observed the logs in Central ELK[^3].

image

[^1]: The query will run at the same cadence that the `github_repositories` table is updated. Currently, once a day.
[^2]: Today, this is 100 repositories. A _lot_ smaller than the 3000+ repositories in the organisation.
[^3]: Strangely https://github.com/guardian/cloudwatch-logs-management doesn't seem to be setting an expiry on the log group, or adding the desired tags to the log lines. This is a separate issue, which we should fix soon!


.nvmrc
```diff
@@ -1 +1 @@
-16.14.2
+18.16.1
```
akash1810, line 1: The lambda's runtime is 18, so also use 18 locally.

akash1810:
Closing in favour of https://github.com/guardian/service-catalogue/pull/234.

PR #230, opened 2023-07-11
feat: Logic for `github-recently-updated` lambda

akash1810:
## What does this change?
Builds on #229, adding the following logic:
1. Find the recently updated repositories
2. Write the list of repositories to Secrets Manager

This secret can then be used by the ECS tasks (https://github.com/guardian/service-catalogue/pull/231). Note, we're using a secret here because it can easily be added to an ECS task.

## How has it been verified?
I've manually deployed and executed the lambda.

## TODO
There is still some outstanding work, which will be completed in follow-up PRs:
- Trigger the lambda when the task to collect GitHub repositories completes
- Alerting


akash1810:
Closing in favour of https://github.com/guardian/service-catalogue/pull/234.

PR #232, opened 2023-07-11
fix: Collect GitHub branch data

akash1810:
## What does this change?
Collecting data for all [branches](https://www.cloudquery.io/docs/plugins/sources/github/tables/github_repository_branches), in all repositories.

Given the amount of data being collected, this task might take a while to complete. I'm using this PR as a benchmark, to understand if #231 is needed.

## Why?
We're specifically interested in the commit sha, as we can use this to compare to Snyk tags to determine if a repository is reliably integrated with Snyk.

## How has it been verified?
I've deployed this and manually triggered the task. 

The stats suggest #231 is not needed:
- Run time of ~7 minutes[^1]
- There are 3393 repositories in the organisation
- There are 25677 branches
- From the branch table, there are 3317 unique repositories[^2]. That is, we have data for 97.7% of the organisation's repositories
- From the branch table, we have data from 3312 default branches. That is, we know the last commit from the default branch of 97.6% of the organisation's repositories

The logs do not point to any errors being encountered, so we'd need to analyse the data to understand the cause of the missing data - presumably the GitHub API isn't returning anything. In any case, I think this is enough to avoid adding the lambda in #229, and the related PRs.

[^1]: Started 11/07/2023, 12:15:42 UTC. Stopped 11/07/2023, 12:21:32 UTC.
[^2]: SELECT COUNT(DISTINCT repository_id) FROM github_repository_branches;
[^3]: SELECT COUNT(DISTINCTfull_name) FROM github_repositories;

### DIFF
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -3298,6 +3298,692 @@ spec:
       },
       "Type": "AWS::Events::Rule",
     },
+    "CloudquerySourceGitHubBranchesScheduledEventRuleD1D634B8": {
+      "Properties": {
+        "ScheduleExpression": "rate(1 day)",
+        "State": "ENABLED",
+        "Targets": [
+          {
+            "Arn": {
+              "Fn::GetAtt": [
+                "cloudqueryCluster5370C11B",
+                "Arn",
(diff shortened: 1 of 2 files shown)


akash1810:
Closing in favour of https://github.com/guardian/service-catalogue/pull/234, which offers a much simpler implementation.

PR #233, opened 2023-07-11
Add SQL views script to repo

NovemberTang:
## What does this change?

Some of our dashboards are relying on Cloudquery views. This PR adds a SQL script to create/recreate those views

## Why?

Views are useful as there are some patterns/joins that we are performing repeatedly, such as connecting a stack to a Galaxies team.

We need an auditable record of our views so that if the database is destroyed, we can recreate this data. We also need to know if and when views have changed, and keeping track of them in GitHub allows us to do this.

At some point we could use something like flyway to manage this, but just having the queries available on GitHub at all is a marked improvement from where we were before

## How has it been verified?

These views are currently live and are viewable in Grafana

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -36,22 +36,38 @@ where (
     )
   and ec2.state ->> 'Name' = 'running';
 
+-- General information about running EC2 instances
+-- | account_name |  app |  image_id  | instance_id | built_by_amigo | launch_time|
+-- |--------------|------|------------|-------------|----------------|------------|
+-- |    account1  | grid | ami-123456 |  i-1234556  |      true      | 2023-06-23 |
 create or replace view view_running_instances as
-select distinct
-    accts.name,
+with id_and_tags as (select image_id, tags ->> 'BuiltBy' as built_by from aws_ec2_images where tags is not null order by image_id),
+
+aggregated_images as (select image_id,
+       CASE
+           WHEN 'amigo' = ANY(array_agg(built_by)) THEN true
+           ELSE false
+           END as built_by_amigo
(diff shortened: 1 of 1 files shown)


akash1810 (approved):

akash1810:
Is it worth mentioning in https://github.com/guardian/service-catalogue/blob/main/docs/getting-started.md that these views exist?

NovemberTang:
> Is it worth mentioning in https://github.com/guardian/service-catalogue/blob/main/docs/getting-started.md that these views exist?

Yes, for sure. We should also document these views so they're easier to understand before referring to them directly though

jorgeazevedo:
I've changed `aws_accounts` from

```sql
CREATE OR REPLACE VIEW aws_accounts AS
SELECT DISTINCT aws_organizations_accounts.id,
                aws_organizations_accounts.name,
                aws_organizations_accounts.email,
                aws_organizations_accounts.status,
                aws_organizations_accounts.joined_timestamp
FROM aws_organizations_accounts;
```

to

```sql
CREATE OR REPLACE VIEW aws_accounts AS
SELECT DISTINCT acc.id,
                acc.name,
                acc.email,
                acc.status,
                acc.joined_timestamp,
                COALESCE(ou.name, 'ROOT') AS "organizational_unit"
FROM aws_organizations_accounts acc
         LEFT JOIN aws_organizations_account_parents par  ON acc.id = par.id
         LEFT JOIN aws_organizations_organizational_units ou on par.parent_id = ou.id
```

This works in grafana and when using user `postgres` to query, but now when using user `cloudquery` as that user is missing permissions to read on `aws_organizations_account_parents`. I'm not super sure how to solve that!

PR #237, opened 2023-07-13
fix: Collect all `aws_organization` tables

akash1810:
## What does this change?
Collect all AWS Org data, as we've started to become more interested in it. In particular the `aws_organizations_account_parents` table tells us which OU an account is in.

Uses [wildcard](https://www.cloudquery.io/docs/advanced-topics/performance-tuning#use-wildcard-matching) matching as there are a lot of [tables](https://www.cloudquery.io/docs/plugins/sources/aws/tables):

image

## Why?
As we organise OUs by department, knowing which account is in which OU helps us determine how each department is performing.

## How has it been verified?
I've collected `aws_organization*` locally, and can confirm the data is expected.

### DIFF
--- a/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/cloudquery.test.ts.snap
@@ -191,13 +191,7 @@ spec:
     - aws_xray_groups
     - aws_stepfunctions_map_runs
     - aws_stepfunctions_map_run_executions
-    - aws_organizations
-    - aws_organizations_accounts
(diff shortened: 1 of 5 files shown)


akash1810:
Ah, the tables of interest are added in a version we're not yet using - https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-aws-v19.0.0. Will update that too. Converting to draft for now.

NovemberTang (approved):

JuliaBrigitte (approved):

PR #240, opened 2023-07-20
refactor: Use the same CloudQuery versions locally and in PROD

akash1810:
## What does this change?
Aimed at improving the story of version upgrades, this PR refactors how we define the versions of CloudQuery being used. Currently, we track the PROD and DEV versions independently, and we're not in sync. In this change, we use the root `.env` file to track the versions for all environments. This should make it easier to verify version bumps.

## Why?
Keeping two files in sync isn't very fun!

## How has it been verified?
- Followed the README for running CloudQuery locally, and it still works as before
- The CDK snapshot tests passing _and_ not being updated demonstrates this is a no-op for our AWS infrastructure

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,9 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires -- can't use import here
+const dotenv = require('dotenv');
+
+// Load environment variables from .env file at the root of the repository
+dotenv.config({ path: `${__dirname}/.env` });
+
 module.exports = {
 	verbose: true,
 	testEnvironment: 'node',
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,25 +1,16 @@
-const generateProject = (name) => {
-	return {
-		displayName: name,
-		transform: {
-			'^.+\\.tsx?$': 'ts-jest',
-		},
-		transformIgnorePatterns: [
-			'node_modules/(?!@guardian/private-infrastructure-config)',
-		],
-		testMatch: [`<rootDir>/packages/${name}/**/*.test.ts`],
-		setupFilesAfterEnv: [`./packages/${name}/jest.setup.js`],
-		moduleNameMapper: {
-			'^common$': '<rootDir>/packages/common/src',
-			'^common/(.*)$': '<rootDir>/packages/common/src/$1',
-		},
-	};
-};
-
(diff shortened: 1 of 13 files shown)


jest.config.js
```diff
@@ -1,25 +1,22 @@
...
+
+// Load environment variables from .env file at the root of the repository
+dotenv.config({ path: `${__dirname}/.env` });
 
 module.exports = {
 	verbose: true,
 	testEnvironment: 'node',
 	projects: [
-		'cdk',
-	].map(generateProject),
```
akash1810, line 24: We only have one project now, so we can remove the abstraction offered by `generateProject`.

package.json
```diff
@@ -17,7 +17,6 @@
     "@guardian/eslint-config-typescript": "^2.0.0",
     "@guardian/prettier": "^2.1.5",
     "@guardian/tsconfig": "^0.2.0",
-    "aws-sdk-client-mock": "^2.0.0",
```
akash1810, line 20: Unused dependency gets removed.

packages/cdk/package.json
```diff
@@ -4,8 +4,7 @@
   "scripts": {
     "test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects cdk",
     "typecheck": "tsc -noEmit",
-    "cdk": "cdk",
-    "synth": "cdk synth --path-metadata false --version-reporting false"
+    "synth": "DOTENV_CONFIG_PATH=../../.env cdk synth --path-metadata false --version-reporting false"
```
akash1810, line 7: This is some [dotenv](https://www.npmjs.com/package/dotenv) preload magic ✨.

packages/cloudquery/package.json
```diff
@@ -2,10 +2,6 @@
   "name": "cloudquery",
   "version": "0.0.0",
   "scripts": {
-    "start": "./script/start",
-    "test": "echo not implmented",
-    "build": "echo not implmented",
-    "lint": "echo not implmented",
-    "typecheck": "echo not implmented"
```
akash1810, line 9: The `package.json` at the root now uses the [`--if-present`](https://docs.npmjs.com/cli/v9/commands/npm-run-script#if-present), making these redundant.

jorgeazevedo (approved):

PR #248, opened 2023-08-17
Initial python implementation of RepoCop data collector lambda

NovemberTang:
## What does this change?

This change adds code that is capable of assessing GitHub repositories against our [best practice recommendations](https://github.com/guardian/recommendations/blob/main/best-practices.md), and creating a table to hold that information.

This code can be run on AWS, as a lambda, or locally, by running the `main.py` file or as a jupyter notebook, for a more interactive and visual way of interacting with the data.

It generates a table that looks something like this

| full_name      | repository_01 | timestamp                  |
|----------------|---------------|----------------------------|
| guardian/repo1 | true          | 2023-08-17 10:18:21.000000 |
| guardian/repo2 | true          | 2023-08-17 10:18:21.000000 |
| guardian/repo3 | false         | 2023-08-17 10:18:21.000000 |

## Why?

This is the first stage of implementing RepoCop, which will notify teams if they are not following best practices, and prompt them to fix this. We've chosen to do this in code, rather than in SQL as unit testing frameworks allow us to verify the correctness of our queries.

Python has been chosen as

- We have considerable Python experience on the team, and already have some tools written in the language
- Its compatibility with jupyter notebooks and simple data visualisation story makes prototyping and data exploration easy
- The pandas library means that transforming and testing data in a tabular format is simple

## How has it been verified?

The repocop lambda is running successfully in PROD, the data transformation logic has been tested, and the table currently exists in the cloudquery database.

## Any additional comments?

This PR does not represent a feature complete version of the repocop data collector. It's a minimal proof of concept that needs further work, namely:

- The local execution story is not great, it requires a connection to the production database. If we choose this implementation, we should find a way to run a local DB.
- We need to find another way to connect to the database, we are currently using the root user to read and write from the DB


packages/repocop/database_interactions.py
```diff
@@ -0,0 +1,50 @@
...
+def connect_to_db(session: boto3.Session):
+    """Gets the credentials for the database from AWS Secrets Manager"""
+    secret_client = session.client('secretsmanager', 'eu-west-1')
+
+    if "SECRET_ARN" in os.environ:
+        print('AWS Lambda environment detected')
+        secret_arn = os.environ.get('SECRET_ARN')
+    else:
+        print("Local environment detected: fetching secret ARN")
+        secret_arn = _get_secret_arn(secret_client)
```
akash1810, line 33: Not blocking, but I think it would be simpler to require an env var for local development too. It'll allow us to remove some code, and also keeps config in one place (the `.env` file).

akash1810: Although moving to IAM authentication would mean the majority of this file goes away, and wouldn't require an env var, so that's probably the better change.

packages/repocop/rules.py
```diff
@@ -0,0 +1,29 @@
...
+    #make sure that the required columns exist
+    assert 'full_name' in default_branch_df.columns, 'full_name column is missing'
+    assert 'default_branch' in default_branch_df.columns, 'default_branch column is missing'
+
+    default_branch_df['repository_01'] = default_branch_df['default_branch'] == 'main'
+    return default_branch_df[['full_name', 'repository_01']]
+
+#owner_df requires columns 'repo_name', 'slug'
+#topics_df requires columns 'full_name', 'topics'
+def repository_06(owner_df, topics_df, production_topics, non_pe_teams)-> DataFrame:
```
akash1810, line 17: This function isn't used by the lambda. Is that intentional?

packages/repocop/script/ci.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+MAJOR_PYTHON_VERSION="3.11"
```
akash1810, line 5: I think this variable is unused now.

akash1810 (approved):
A couple of comments, none of which are blocking.

I'd be keen to follow up on this PR with two changes too:
  - Avoid using the PROD database when running locally
  - Adopting IAM authentication to the database from the lambda

EDIT: Just spotted you've already made this same points to improve in the PR description. Apologies.

packages/repocop/rules.py
NovemberTang: yeah, i think one rule is enough to demonstrate that this is doable within a lambda. We've already demonstrated that we can merge multiple repo rule dataframes elsewhere, and I don't want to clog the lambda up with repository_06 stuff until we've renamed the `view_repository_owners` table columns

packages/repocop/database_interactions.py
NovemberTang: yeah exactly, i don't want to sink too much time into making this neat if we're just gonna chuck most of it away

packages/repocop/test_rules.py
```diff
@@ -0,0 +1,55 @@
...
+    ], columns=['full_name', 'default_branch'])
+
+    repository_01_df = repository_01(branch_df)
+    # check only one cell in the repository_01 column is True
+    repo1_follows_rule = repository_01_df.loc[repository_01_df['full_name'] == 'repo1', 'repository_01'].isin([True]).bool()
+    true_count= (repository_01_df['repository_01']).sum()
+    false_count = (~repository_01_df['repository_01']).sum()
+    assert repo1_follows_rule is True
+    assert true_count == 1
+    assert false_count == 2
```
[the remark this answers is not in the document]
NovemberTang: good point! though pytest is currently being grumpy so I'll merge now and address this in the next PR.

PR #251, opened 2023-08-24
Rebuild and rename stack

JuliaBrigitte:
## What does this change?
Changed the stack name to service-catalogue from cloudquery. 

## Why?
To better reflect the fact that it now includes repocop and cloudquery.
We had to rebuild the stack because the database from the old stack had been corrupted. We took the opportunity to change the stack name.

## How has it been verified?
We successfully deployed with riffraff.

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -52,7 +52,7 @@ import {
 	standardDenyPolicy,
 } from './ecs/policies';
 
-export class CloudQuery extends GuStack {
+export class ServiceCatalogue extends GuStack {
 	constructor(scope: App, id: string, props: GuStackProps) {
 		super(scope, id, props);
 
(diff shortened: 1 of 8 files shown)


NovemberTang (approved):

packages/cdk/lib/service-catalogue.ts
```diff
@@ -248,7 +248,7 @@ export class CloudQuery extends GuStack {
 					) as string[]),
 				],
 
-				// Defaulted to 500000 by CloudQuery, concurrency controls the maximum number of Go routines to use.
+				// Defaulted to 500000 by ServiceCatalogue, concurrency controls the maximum number of Go routines to use.
```
akash1810, line 251: Not sure all the changes in the comments are correct. For example, this comment indeed relates to the CloudQuery application, and not our Service Catalogue.

PR #252, opened 2023-08-24
Create a code environment that runs less often.

NovemberTang:
## What does this change?

- Creates separate CODE and PROD stacks. They supersede the single INFRA stack, which can then be deleted
- Set the cadence of the all ECS jobs in the code stack to be monthly, to minimise cost, while still keeping data vaguely up to date (we will add a mechanism to kick off all these jobs manually in a subsequent PR)
- Create GitHub and Snyk secrets for API keys (these will need to be populated if the secret is recreated)

## Why?

The last time we attempted a database migration on INFRA, we got ourselves into a bit of a state. A CODE stack means we can test out changes like this in a cloud environment.

## How has it been verified?

The stacks are running, and we can connect to the CODE database to verify the data exists

### DIFF
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -1,8 +1,8 @@
 import 'source-map-support/register';
 import { GuRootExperimental } from '@guardian/cdk/lib/experimental/constructs/root';
-import { ServiceCatalogue } from '../lib/service-catalogue';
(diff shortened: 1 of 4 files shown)


akash1810 (approved):
Looks great.

For the benefit of future travellers, I wonder if its worth describing the impact of these changes too? That is, two new stacks will be created, and the `INFRA` one will need to be deleted manually.

PR #259, opened 2023-08-25
feat: Collect GitHub Workflow data

akash1810:
The [`github_workflows` table](https://www.cloudquery.io/docs/plugins/sources/github/tables/github_workflows) has a `content` column, allow the workflow file's content to be queried.

With adoption of GitHub Actions for CI is growing, it'll be interesting to see how we're using GHA. Potential use-cases include:
  - A vuln. has been detected in a third-party action. Which repositories are impacted?
  - How is a repository integrated with Snyk? Via the departmental action? Or other means?

## How has it been verified?
Ran locally.


tjsilver (approved):
Presumably we'll need to grant grafana permissions again?

JuliaBrigitte (approved):

PR #260, opened 2023-08-25
Initial Prisma Implementation

NovemberTang:
## What does this change?

Introduces Prisma to the service catalogue. This project allows us to use typescript to run queries against the cloudquery data, and write back the results.

We have created a scheduled lambda with access to the CloudQuery database, that performs RepoCop evaluations daily based on [these rules](https://github.com/guardian/recommendations/blob/main/best-practices.md).

Currently, it runs against one repo, and the operation to write the data back to a new table does not work. This will be addressed in subsequent PRs. I am raising this one now as there is a lot of divergent work that needs to happen (cleaning up CDK, and getting the migrations running correctly can run in parallel, and merging now gives us an adequate baseline for both of those to happen.

## Why?

The reasoning is almost identical to that in #248, with the added benefit of Prisma being able to create types based on the database schema, for extra security. It also allows us to create REST APIs over the database if we want to, and handles database migrations, meaning we could reduce our reliance on the currently fragile way we have of creating our custom tables and views (in sql/views.sql)

## How has it been verified?

The lambda runs. It does not create the repocop table, but the logs show that repositories are being correctly evaluated.


akash1810:
Could I suggest we stage these changes into smaller chunks? For example:
  - Upgrade to Node 18
  - Introduce Prisma, with a simple `SELECT` query on one table (to prove the wiring works)
  - Set up Prisma to migrate on CD
  - Implement RepoCop logic

Small PRs reduces risk, and allows us to progressively build on a working solution.

NovemberTang:
> * Upgrade to Node 18

AFAICT there isn't a node upgrade as part of this PR. Is there a reason it couldn't wait until after the other work is complete?

akash1810:
> > * Upgrade to Node 18
> 
> AFAICT there isn't a node upgrade as part of this PR. Is there a reason it couldn't wait until after the other work is complete?

Ah, sorry, you're right. We're using v3 of the AWS JS SDK, which is included in the [Node 18 lambda runtime](https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/).

Using Node 16 and SDK v3 means our artifact is slightly bigger, as we have to bundle the SDK in.

You're right though, it's not a blocker.

tjsilver:
The .idea folder seems to have been committed even though it was in .gitignore.

akash1810:
We've broken this down into a set of smaller PRs now:
- https://github.com/guardian/service-catalogue/pull/277
- https://github.com/guardian/service-catalogue/pull/278
- https://github.com/guardian/service-catalogue/pull/280
- https://github.com/guardian/service-catalogue/pull/291
- https://github.com/guardian/service-catalogue/pull/292
- https://github.com/guardian/service-catalogue/pull/294

There are still a few questions around Prisma, for example, what's the best way to automatically perform migrations? (they're currently being performed manually). However, that was out of scope for this PR, so I think we can close this now.

PR #254, opened 2023-08-25
feat: Add a CLI to list and run ECS tasks

akash1810:
## What does this change?
Running a task on demand via the AWS web console is a bit of a faff, as you have to fill in a lot of fields, and it is easy to get wrong.

This change adds a CLI to make it easier to list and run tasks. There are three commands:
- `list-tasks` to  list all task definitions for a given set of `Stack`, `Stage`, and `App` tags
- `run-task` to run a single task using the task's `Stack`, `Stage`, `App` and `Name` tags (requires #253)
- `run-all-tasks` to run all tasks for a given set of `Stack`, `Stage`, and `App` tags

## Why?
Reduce human error in the AWS web console, and improve DX.

## How has it been verified?
Run the command(s) locally (from the root). For example:

```sh
npm run cli --workspace cloudquery list-tasks -- \
  --stack deploy \
  --stage CODE \
  --app service-catalogue
```

```sh
npm run cli --workspace cloudquery run-task -- \
  --stack deploy \
  --stage CODE \
  --app service-catalogue \
  --name DeployToolsListOrgs
```

PR #272, opened 2023-08-30
Rollback to CQ_AWS 22.6.0

jorgeazevedo:


michaelwmcnamara (approved):

akash1810:
> We've been able to rollback production to this version and re-run the load balancer ecs task, and the data is now coming through correctly.

ICYMI since https://github.com/guardian/service-catalogue/pull/254 we have a CLI to make it easier to run an ECS task.

PR #280, opened 2023-09-01
Add prisma

JuliaBrigitte:
## What does this change?
Add prisma and generate schema in schema.prisma, make a query in table… repository with the prisma client
We had to create the db user repocop manually (dbuser.sql)

## Why?
First step to use prisma with cloudquery

## How has it been verified?
Deploy to code and run lambda
[riffraff deploy](https://riffraff.gutools.co.uk/deployment/view/1c43e970-6119-4187-a940-a6c253324dac)

### DIFF
--- a/packages/repocop/index.ts
+++ b/packages/repocop/index.ts
@@ -36,7 +36,7 @@ export async function main() {
 			db: {
 				url: await getDatabaseUrl(config),
 			},
-		},
+		}
 	});
 	console.log('Query prisma');
 	const queryResult = await prisma.github_repositories.findFirst({where: {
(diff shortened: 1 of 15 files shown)


packages/repocop/index.ts
```diff
@@ -1,3 +1,67 @@
...
+	console.log('Query prisma');
+	const queryResult = await prisma.github_repositories.findFirst({where: {
+		archived: false,
+	}})
+
+	if (queryResult)
+		console.log(queryResult.name || "not found")
+}
+
+async function getDatabaseUrl(config: Config){
```
akash1810, line 50: I don't think we'd want this when running locally. Locally we should connect to the local database running in Docker, which wouldn't support IAM authentication.

Not a blocker; we can address this in a follow-up PR.

akash1810 (approved):

PR #285, opened 2023-09-05
Streamline docs

NovemberTang:
## What does this change?

This change improves the developer experience. Following the instructions now means that this works first time. I've also added some more information that will be useful for first time contributors.

I've also made the output from the script more readable, simplifying the paths displayed in the output, and showing error messages in a different colour

## Why?

Having pulled the repo fresh and trying to follow the setup instructions, I'd often have to go back and repeat steps. The steps are now in the correct order.

Often, when setting up the project, users did not understand that the setup script was failing. Showing the error message in a different colour will hopefully make that clearer

## How has it been verified?

Tested locally. Now a bad output looks like this. Much more error-y

Screenshot 2023-09-05 at 10 30 17

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,9 @@ The following packages form part of Service Catalogue:
 
 * [cloudquery](packages/cloudquery/README.md): A set of Cloudquery (ECS) tasks to collect the data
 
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
Followed the instructions and got up and running flawlessly! 👍

PR #286, opened 2023-09-05
refactor(repocop): Use local database when running locally

akash1810:
## What does this change?
Addresses [this comment](https://github.com/guardian/service-catalogue/pull/280#discussion_r1315026786) and refactors Repocop to use a locally running database when running locally.

This change also introduces a `src` directory to house all the TypeScript for Repocop.

## Why?
Developing against a local database provides a faster feedback cycle, and also reduces any destructive impact should we make a breaking change to the database.

## How has it been verified?
Run Repocop locally via `npm start -w repocop`.

### DIFF
--- a/packages/repocop/src/run-locally.ts
+++ b/packages/repocop/src/run-locally.ts
@@ -1,5 +1,12 @@
 import { main } from './index';
 
 if (require.main === module) {
+    // TODO set these values from a .env file.
+    //  The .env file should also be used within `docker-compose.yml` to stay DRY.
+    process.env.STAGE = 'DEV';
+    process.env.DATABASE_HOSTNAME = 'localhost';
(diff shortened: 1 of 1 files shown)


packages/repocop/src/config.ts
```diff
@@ -0,0 +1,86 @@
...
+    username: user,
+    region: 'eu-west-1'
+  });
+
+  return await signer.getAuthToken();
+}
+
+async function getDatabaseConnectionString(config: DatabaseConfig) {
+  const { user, password, hostname, port  } = config;
+  const dbPassword = password ?? await getRdsToken(config);
```
akash1810, line 83: If we need to connect to the CODE database locally, we can unset the `DATABASE_PASSWORD` env var; we'll fall to the right hand side of this expression.

PR #287, opened 2023-09-06
Fix: update repocop README for running locally

tjsilver:
## What does this change?

Updates instructions in README for running `repocop` locally. `ts-node` does not run from a local dependency without `npx`, but instead can but run from a script in `package.json`. The current command would only work if the user had ts-node installed globally.

## Why?

Users should be able to run code in the repo using packages provided in the repo, without relying on them to be installed globally.


packages/repocop/README.md
```diff
@@ -1,8 +1,5 @@
...
-The .env file contains the DATABASE_URL needed for prisma
 
-To run locally you can run
+The .env file contains the DATABASE_URL needed for prisma.
 
-    ts-node run-locally.ts
-
-which is also available as start in the package.json
\ No newline at end of file
+To run locally you can run `npm run start` in the `repocop` directory.
```
akash1810, line 5: Minor: `start` is a [predefined command](https://docs.npmjs.com/cli/v7/commands/npm-start), so calling `run` isn't needed.

akash1810 (approved):

packages/repocop/README.md
tjsilver: Oh yes, forgot 🤦

PR #289, opened 2023-09-07
ci: Remove `actions/setup-python`

akash1810:
## What does this change?
Since #267 we no longer have Python in the repository, so can remove it from the CI environment.


adamnfish (approved):
Nice one

PR #291, opened 2023-09-07
Configure intial db migration

JuliaBrigitte:
## What does this change?
Add directory prisma/migration and a file with the initial data layout
The migration process is manual for now.

## Why?
We want to be able to use prisma migrations to make changes to the database and need to set is up according to the
[prisma doc](https://www.prisma.io/docs/concepts/components/prisma-migrate/db-push#choosing-db-push-or-prisma-migrate)

## How has it been verified?
We applied the migration first locally and to code and it worked
```
npx prisma migrate resolve --applied 0_init
```

### DIFF
--- a/packages/repocop/prisma/migrations/0_init/migration.sql
+++ b/packages/repocop/prisma/migrations/0_init/migration.sql
@@ -1,0 +1,17855 @@
+-- CreateTable
+CREATE TABLE "aws_accessanalyzer_analyzer_archive_rules" (
+    "_cq_sync_time" TIMESTAMP(6),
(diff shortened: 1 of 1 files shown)


akash1810 (approved):

akash1810:
I think we followed https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate/add-prisma-migrate-to-a-project#baseline-your-production-environment here, and not https://www.prisma.io/docs/concepts/components/prisma-migrate/db-push#choosing-db-push-or-prisma-migrate.

PR #292, opened 2023-09-07
Add table repocop_github_repository_rules

JuliaBrigitte:
## What does this change?
Adds table repocop_github_repository_rules with prisma migrate

## Why?
We want to have a change history of our db changes so are use prisma migrations

## How has it been verified?
Manual applied on local machine
```
npx prisma migrate dev --name create-table-repocop_github_repository_rules
```


akash1810:
I think we want to use `prisma migrate deploy` for these and future changes.

`prisma migrate dev` seems to want to keep the schema file and the database in sync; any additional tables in the database will get removed:

```console
➜ npx -w repocop prisma migrate dev --name create-table-repocop_github_repository_rules
Environment variables loaded from .env
Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "postgres", schema "public" at "localhost:5432"

Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

It should be understood as the set of changes to get from the expected schema to the actual schema.

[+] Added tables
  - aws_rds_cluster_parameters
  - aws_rds_engine_versions
  - repocop_github_repository_rules

[*] Changed the `aws_rds_engine_versions` table
  [+] Added unique index on columns (_cq_id)

[*] Changed the `repocop_github_repository_rules` table
  [+] Added unique index on columns (full_name)

✔ We need to reset the "public" schema at "localhost:5432"
Do you want to continue? All data will be lost. … no

Reset cancelled.
```

With `prisma migrate deploy`, only the `.sql` files in the migration directory are applied; any tables in the database but not in the schema file are left untouched:

```console
➜ npx -w repocop prisma migrate deploy
Environment variables loaded from .env
Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "postgres", schema "public" at "localhost:5432"

2 migrations found in prisma/migrations

Applying migration `20230907105726_create_table_repocop_github_repository_rules`

The following migration have been applied:

migrations/
  └─ 20230907105726_create_table_repocop_github_repository_rules/
    └─ migration.sql
      
All migrations have been successfully applied.
```

akash1810:
Successfully applied to CODE and PROD now via `npx -w repocop prisma migrate deploy`. Merging.

akash1810 (approved):

PR #294, opened 2023-09-07
feat(repocop): Add logic to evaluate rules repository_01 and repository_02

akash1810:
## What does this change?
Heavily inspired by #260, this change provides an initial implementation applying RepoCop logic:
  1. Get one repo
  2. Get all branches
  3. Evaluate [rules repository_01 (default branch should be main), and repository_02 (branch protection on default branch)](https://github.com/guardian/recommendations/blob/main/best-practices.md)
  4. Write results to database

AND TESTS!

Note: We've disabled the linter on the repocop project, so there might be some formatting issues!

## Why?
This change completes the end to end demonstration of using Prisma for RepoCop. The next stage would be to implement the evaluation of the remaining rules. This will be done in a separate PR.

## How has it been verified?
Run the branch locally via `npm -w repocop start`, observe the logs, and observe the `repocop_github_repository_rules` table being written to.

### DIFF
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -1,5 +1,6 @@
 import { PrismaClient } from "@prisma/client";
 import { getConfig } from "./config";
+import { repositoryRuleEvaluation } from "./rules/repository";
 
 export async function main() {
 	const config = await getConfig();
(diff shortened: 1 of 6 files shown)


packages/repocop/src/index.ts
```diff
@@ -10,11 +11,27 @@ export async function main() {
...
 
-	if (queryResult)
-		console.log(queryResult.name || "not found")
+	// TODO Process ALL repositories
+	const repo = await prisma.github_repositories.findFirst();
+
+	if (!repo) {
+		console.log('No repositories found');
+	} else {
+		const branches = await prisma.github_repository_branches.findMany();
```
akash1810, line 21: I think this is a `SELECT * FROM github_repository_branches` query; should we add a `WHERE full_name = 'X'` to improve performance?

NovemberTang (approved):

akash1810:
The lambda uses the database user `repocop`, which currently has `SELECT` permission on all tables. We need to grant it permissions to delete and insert rows into the `repocop_github_repository_rules` table before merging this.

akash1810:
> The lambda uses the database user `repocop`, which currently has `SELECT` permission on all tables. We need to grant it permissions to delete and insert rows into the `repocop_github_repository_rules` table before merging this.

I've manually run the following on CODE and PROD:

```sql
GRANT ALL ON public.repocop_github_repository_rules TO repocop;
```

PR #306, opened 2023-09-12
feat(repocop): Implement rule repository_04

akash1810:
## What does this change?
Implements rule [repository_04](https://github.com/guardian/recommendations/blob/main/best-practices.md):

> Grant at least one GitHub team Admin access - typically, the dev team that own the project.
> Repositories without one of the following topics are exempt: production, testing, documentation.

## How has it been verified?
Run Repocop locally. I've also [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/c236cb9b-db4c-4e4c-9e35-c326ade5e6c0), and successfully executed the lambda.

### DIFF
--- a/packages/repocop/src/query.ts
+++ b/packages/repocop/src/query.ts
@@ -1,7 +1,7 @@
 import type { github_repositories, Prisma, PrismaClient } from '@prisma/client';
 import type { GetFindResult } from '@prisma/client/runtime/library';
 
-export type RepositoryTeams = GetFindResult<
+export type RepositoryTeam = GetFindResult<
 	Prisma.$github_team_repositoriesPayload,
 	{
 		select: { role_name: boolean; id: boolean };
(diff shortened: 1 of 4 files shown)


packages/repocop/src/query.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function getRepositoryTeams(
+	client: PrismaClient,
+	repository: github_repositories,
+): Promise<RepositoryTeams[]> {
+	const data: RepositoryTeams[] =
+		await client.github_team_repositories.findMany({
+			select: {
+				id: true,
+				role_name: true,
+			},
```
akash1810, line 21: Repocop was timing out on CODE when selecting all the fields in this table. This change only pulls out the fields we actually use, resulting in less memory usage, and a successful lambda execution.

The lambda is configured with a 30 second timeout, so we could increase that also/instead.

akash1810: Worth noting the [docs](https://www.prisma.io/docs/guides/performance-and-optimization/query-optimization-performance#using-select-to-limit-number-of-columns-returned) suggests that this level of fine-tuning isn't typically necessary.

packages/repocop/src/query.ts
```diff
@@ -0,0 +1,36 @@
+import type { github_repositories, Prisma, PrismaClient } from '@prisma/client';
+import type { GetFindResult } from '@prisma/client/runtime/library';
+
+export type RepositoryTeam = GetFindResult<
+	Prisma.$github_team_repositoriesPayload,
+	{
+		select: { role_name: boolean; id: boolean };
+		where: { id: bigint };
+	}
+>;
```
akash1810, line 10: The IDE generated this type for me.

tjsilver (approved):
💯

PR #307, opened 2023-09-13
refactor: Move all query parts into `query.ts`

akash1810:
## What does this change?
Moves all query functions into the `query.ts` file.

## Why?
For better organisation - `index.ts` now simply imports from `query.ts` and passes the results into the rule evaluation function.

## How has it been verified?
N/A.


tjsilver (approved):
👍

PR #313, opened 2023-09-13
Get all unarchived branches up front

NovemberTang:
## What does this change?

Previously, we were getting branches for each repo by making a DB call for each one. Now, we are grabbing all the ones we need at the beginning

## Why?

Fewer requests mean fewer chances for requests to fail. Also places less strain on the DB

## How has it been verified?

~~Tested a before and after on CODE. Job ran successfully. We've gone from 7.5GB seconds to 9 GB seconds. Assuming one run a day for 30 days, this equates to an additional cost of $0.00075 per month (about 1/12th of a cent). Which seemed like an acceptable increase to me.~~

After following @akash1810 's advice on filtering archived repos, we've actually **dropped** from 7.5 GB seconds to 7.1.

### DIFF
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -2,9 +2,9 @@ import { PrismaClient } from '@prisma/client';
 import type { repocop_github_repository_rules } from '@prisma/client';
 import { getConfig } from './config';
 import {
-	getUnarchivedRepositories,
-	getRepositoryTeams,
 	getRepositoryBranches,
+	getRepositoryTeams,
+	getUnarchivedRepositories,
 } from './query';
 import { repositoryRuleEvaluation } from './rules/repository';
 
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -2,9 +2,9 @@ import { PrismaClient } from '@prisma/client';
 import type { repocop_github_repository_rules } from '@prisma/client';
 import { getConfig } from './config';
 import {
-	getRepositories,
+	getUnarchivedRepositories,
 	getRepositoryTeams,
-	getUnarchivedRepositoryBranches,
+	getRepositoryBranches,
 } from './query';
 import { repositoryRuleEvaluation } from './rules/repository';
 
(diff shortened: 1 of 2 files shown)


packages/repocop/src/query.ts
```diff
@@ -29,23 +29,19 @@ export async function getRepositories(
...
 }
 
-export async function getRepositoryBranches(
+// We only care about branches from unarchived repos, so lets only pull those to save us some time/memory
+export async function getUnarchivedRepositoryBranches(
 	client: PrismaClient,
-	repository: github_repositories,
+	repos: github_repositories[],
 ): Promise<github_repository_branches[]> {
+	const unarchivedRepos = repos.filter((repo) => !repo.archived);
```
akash1810, line 37: minor: I don't think filtering for archived is entirely necessary here as the incoming `repos` originates from [this call](https://github.com/guardian/service-catalogue/blob/0873c590f2a0917933fa3f279c10f025ba2144d6/packages/repocop/src/query.ts#L15-L16). That is, the filtering has been done previously.

akash1810 (approved):

packages/repocop/src/index.ts
```diff
@@ -1,10 +1,10 @@
-import type { repocop_github_repository_rules } from '@prisma/client';
 import { PrismaClient } from '@prisma/client';
+import type { repocop_github_repository_rules } from '@prisma/client';
```
akash1810, line 2: This change is curious, as it suggests we've our environments setup differently. Let's address this separately, as it'll be annoying to have these swap orders on every pull request.

adamnfish (approved):
Looks good, although there are two things to consider:

This will make the performance a bit more "spiky".

We'll have a big network and CPU spike up front while we request and parse this data, rather than having this be spread out while the work is done. No harm, I think this sounds like a good idea.

This will also raise the amx memory usage. It mean holding all the branch objects in memory at the same time rather than allowing us to only hold the relevant ones as they are used, so it might be worth doing an back-of-envelope on the numbers of branches / amount of memory we expect here - especially with the repository and branch count expected to continue to grow. I'm sure this is well within sensible, you have already tested it for now, and this sounds like a helpful change in general.

Mostly unrelated to this PR: Should there be tests of this behaviour? For example, making sure we extract only branches relating to the correct repository? This is slightly more important after this change, now that we're extracting the relevant information from all repositories. I can see this mixes remote calls with logic in the TS style, so it might be a bit of a challenge to test, but it looks like there are a few important bits of logic around here.

akash1810:
> Mostly unrelated to this PR: Should there be tests of this behaviour? For example, making sure we extract only branches relating to the correct repository? This is slightly more important after this change, now that we're extracting the relevant information from all repositories. I can see this mixes remote calls with logic in the TS style, so it might be a bit of a challenge to test, but it looks like there are a few important bits of logic around here.

I'd like to add integration tests to Repocop where we:
- Launch a local postgres
- Seed some data
- Run Repocop against it
- Assert against the rows of the `repocop_github_repository_rules` table

PR #314, opened 2023-09-14
feat(repocop): Implement rule repository_05

akash1810:
## What does this change?
Implements rule [repository_05](https://github.com/guardian/recommendations/blob/main/best-practices.md):

> Repositories that are no longer used should be archived.

Where "no longer used" is defined as:
- No commits added to repository in last year
- Repository topic is not `production`, `testing` or `documentation`
- Repository is not owned by a team (for a team to own a repository, they must be admins of that repository)
- TBD: Repository is not running in an AWS services

## Why?
Archiving unused repositories improves the overall health of our GitHub organisation. It allows us to be more focused in our efforts of, for example, updating dependencies.

## How has it been verified?
TBD.

### DIFF
--- a/packages/repocop/src/date.test.ts
+++ b/packages/repocop/src/date.test.ts
@@ -1,0 +1,26 @@
+import { daysDifference } from './date';
+
+describe('Date calculation', () => {
+	test('Calculating the difference between two dates, both at midnight, in the same month', () => {
+		const date1 = new Date(2023, 0, 20, 0, 0, 0);
+		const date2 = new Date(2023, 0, 0, 0, 0, 0);
+
+		expect(daysDifference(date1, date2)).toEqual(20);
+	});
+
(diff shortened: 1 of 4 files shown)


akash1810:
If we consider `repository_02`:

> Enable branch protection for the default branch, ensuring changes are reviewed before being deployed.

This rule is blocking; we'd recommend teams address it as soon as possible as it is crucial for our security posture.

After chatting with @adamnfish about rule `repository_05`, we concluded this one is not blocking. It doesn't change our security posture. It's aim is to reduce the number of repositories teams should monitor.

For this reason, it shouldn't be flagged by RepoCop, at least in the first phase. The first phase of RepoCop should be reporting on those blocking rules (not least because they are more clearly defined).

The `repository_05` rule is still valid, so we should still have it defined.

Closing. For now.

PR #315, opened 2023-09-15
Add evaluated_on migration for table repocop_github_repository_rules

JuliaBrigitte:
## What does this change?
Add a mandatory field evaluated_on to the table repocop_github_repository_rules 

## Why?
We want to have a timestamp when the entries were last generated

## How has it been verified?
Applied locally

## How to deploy to code
We do have a migration history for database changes but applying those changes is a manual process.
We need to apply 
```
npx prisma migrate deploy  
```

### DIFF
--- a/packages/repocop/prisma/migrations/20230915132202_add_field_evaluated_on_to_repocop_github_repository_rules/migration.sql
+++ b/packages/repocop/prisma/migrations/20230915132202_add_field_evaluated_on_to_repocop_github_repository_rules/migration.sql
@@ -1,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `evaluated_on` to the `repocop_github_repository_rules` table without a default value. This is not possible if the table is not empty.
+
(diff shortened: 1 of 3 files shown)


akash1810 (approved):

packages/repocop/prisma/migrations/20230915132202_add_field_evaluated_on_to_repocop_github_repository_rules/migration.sql
```diff
@@ -0,0 +1,10 @@
+/*
+  Warnings:
+
+  - Added the required column `evaluated_on` to the `repocop_github_repository_rules` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- DeleteEntries
+DELETE FROM repocop_github_repository_rules;
```
akash1810, line 8: minor: might be worth adding a comment here to explain why we're happy deleting the content of this table?

PR #326, opened 2023-09-20
Refactor: Read GH and Snyk tokens from user's local `.env`

tjsilver:
## What does this change?

- The Cloudquery setup script now creates a `.env` file in the user's `$HOME/.gu/service_catalogue/secrets/` directory if it's not there already, and populates it with the token variable names. The user then has to paste in their credentials. 
- Two env files are now passed to the `docker compose` command for running locally - the main repo `.env` containing the versions of all the dependencies and the user's local `.env` containing their tokens.

## Why?

This keep the user's tokens out of the repo altogether so they not accidentally committed in the repo's main `.env` file. It also simplifies the process in that a copy of `.env` as `.env.local` is no longer required in the repo. 

This is a step towards simplifying local development and making it more secure.

## How has it been verified?

Tested locally.


NovemberTang (approved):
smart!

PR #328, opened 2023-09-20
feat(prisma): Add Galaxies tables

akash1810:
Similar to #327, this change enables us to interrogate the Galaxies tables within Prisma.

Some minor changes to the setup script have been made.

- We now make the AWS credential check call earlier on, so we know whether or not it's safe to retrieve the galaxies bucket name parameter.
- We only prompt users to update their .env file if we detect that the values are empty.
- Text colouring has been changed to only highlight calls to action and links

Preview of new script output

## Rollout
We don't want to apply the migration to CODE and PROD as it'll result in data loss. Instead, we'll need to [resolve](https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-resolve) the migration:

```sh
npx -w repocop prisma migrate resolve --applied 20230920135403_galaxies_tables
```

---

Co-authored by @NovemberTang


packages/cloudquery/docker-compose.yaml
```diff
@@ -24,6 +24,7 @@ services:
       CQ_SNYK: ${CQ_SNYK}
       GITHUB_ACCESS_TOKEN: ${GITHUB_ACCESS_TOKEN}
       SNYK_TOKEN: ${SNYK_TOKEN}
+      GALAXIES_BUCKET: ${GALAXIES_BUCKET}
```
akash1810, line 27: This is a new secret. It'll be easier to document this once #326 is merged.

scripts/setup.sh
```diff
@@ -36,13 +36,26 @@ setup_node_env() {
...
 }
 
 setup_cloudquery() {
   red='\033[0;31m'
   clear='\033[0m'
-  green='\033[0;32m'
+  yellow='\033[1;33m'
+  cyan='\033[0;36m'
+
+  echo "Checking AWS credentials"
```
NovemberTang, line 48: This isn't new, it's just moved up, as we need to do the credential check before pulling the bucket name

akash1810: > we need to do the credential check before pulling the bucket name

Should we be checking the `deployTools` profile here then?

scripts/setup.sh
```diff
@@ -70,32 +87,35 @@ setup_cloudquery() {
...
-  STATUS=$(aws sts get-caller-identity --profile developerPlayground 2>&1 || true)
-  if [[ ${STATUS} =~ (ExpiredToken) ]]; then
-    echo -e "${red}Credentials for the developerPlayground profile have expired${clear}. Please fetch new credentials."
-    exit 1
-  elif [[ ${STATUS} =~ ("could not be found") ]]; then
-    echo -e "${red}Credentials for the developerPlayground profile are missing${clear}. Please fetch some."
-    exit 1
-  else
-    echo "AWS credentials are valid"
+  # Check if GitHub token is set
```
tjsilver, line 104: Nice!

scripts/setup.sh
NovemberTang: whoops! correct!

NovemberTang: fixed

scripts/setup.sh
NovemberTang: thanks i was quite proud of that one

PR #329, opened 2023-09-21
Use deployTools as dev account

NovemberTang:
## What does this change?

Previously, we were using developerPlayground and deployTools resources to test Cloudquery. This required us to fetch two sets of credentials. We are now using only deployTools resources for local testing.

## Why?

To improve the developer experience, and make setup faster

## How has it been verified?

The correct error messages show up at the expected points, and we are able to gather AWS data locally for the relevant account

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -2,6 +2,12 @@
 
 set -e
 
+# Set up text colours
+red='\033[0;31m'
+clear='\033[0m'
+yellow='\033[1;33m'
+cyan='\033[0;36m'
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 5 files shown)


scripts/setup.sh
```diff
@@ -44,24 +50,17 @@ check_credentials() {
...
 }
 
 setup_cloudquery() {
-  red='\033[0;31m'
-  clear='\033[0m'
-  yellow='\033[1;33m'
-  cyan='\033[0;36m'
-
-  check_credentials developerPlayground
-  check_credentials deployTools
```
NovemberTang, line 64: the deployTools check now runs before the `setup_cloudquery` function

tjsilver (approved):
👍

PR #336, opened 2023-09-26
feat(cloudquery): Collect S3 buckets in the own task

akash1810:
## What does this change?
We're seeing some S3 buckets not appearing in the database. The logs don't show anything obvious, however as S3 data collection is part of the `RemainingAwsData` task, it is likely that the logs are not easy to parse.

This change moves S3 bucket collection into its own task. Hopefully this makes it easier to identify issues.

## How has it been verified?
N/A

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -262,6 +262,16 @@ export class ServiceCatalogue extends GuStack {
 				managedPolicies: [readonlyPolicy],
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
+			{
+				name: 'OrgWideS3Buckets',
+				description: 'Collecting S3 bucket data across the organisation',
+				schedule: nonProdSchedule ?? Schedule.rate(Duration.days(1)),
+				config: awsSourceConfigForOrganisation({
+					tables: ['aws_s3_*'],
+				}),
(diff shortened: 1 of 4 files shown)


packages/cdk/lib/service-catalogue.ts
```diff
@@ -262,6 +262,16 @@ export class ServiceCatalogue extends GuStack {
 				managedPolicies: [readonlyPolicy],
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
+			{
+				name: 'OrgWideS3Buckets',
+				description: 'Collecting S3 bucket data across the organisation',
+				schedule: nonProdSchedule ?? Schedule.rate(Duration.days(1)),
+				config: awsSourceConfigForOrganisation({
+					tables: ['aws_s3_*'],
```
akash1810, line 270: This wildcard will match the following tables:

- aws_s3_access_points
- aws_s3_accounts
- aws_s3_buckets
- aws_s3_bucket_cors_rules
- aws_s3_bucket_encryption_rules
- aws_s3_bucket_grants
- aws_s3_bucket_lifecycles
- aws_s3_bucket_notification_configurations
- aws_s3_bucket_object_lock_configurations
- aws_s3_bucket_websites
- aws_s3_multi_region_access_points

akash1810:
This didn't improve reliability of S3 bucket data. Closing.

PR #338, opened 2023-09-28
fix(prisma): Bring some manually created tables into Prisma

akash1810:
## What does this change?
Adds a couple of manually created enumeration tables as Prisma migrations, to reduce the amount of manually invoked SQL.

## Why?
This is related to some work to bring our views into Prisma. Part of this involves [introspecting the views](https://www.prisma.io/docs/concepts/components/prisma-schema/views#use-introspection). I'm pulling this change out separately to keep the diff as small.


NovemberTang (approved):

akash1810:
I've applied these to CODE and PROD via:

```sh
npx -w repocop prisma migrate resolve --applied 20230928080512_guardian_production_status 
npx -w repocop prisma migrate resolve --applied 20230928080813_guardian_non_p_and_e_github_teams
```

PR #339, opened 2023-09-28
feat(repocop): Query the `view_repo_ownership` view

akash1810:
## What does this change?
Following https://github.com/guardian/service-catalogue/pull/338, this change brings the SQL view `view_repo_ownership` into Prisma[^1] so that it can be used within RepoCop.

## Why?
In order for RepoCop to know who to notify about a repository, we need to know who owns it, this information is within the `view_repo_ownership` view.

## How has it been verified?
I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/6a5f63dc-32ea-46b7-8984-03867cc0b209), run the lambda, and observed [the logs](https://logs.gutools.co.uk/s/devx/app/discover#/doc/626e7830-dd3a-11ea-88b6-37713dba5739/logstash-dev-tools-2023.09.28?id=B38O24oB1GrvOkJ_ak7i) matching the view.

### Log message
image

### SQL result
image

[^1]: SQL view support in Prisma is currently in "ver early Preview" - https://www.prisma.io/docs/concepts/components/prisma-schema/views

### DIFF
--- a/packages/repocop/prisma/migrations/20230928084104_view_repo_ownership/migration.sql
+++ b/packages/repocop/prisma/migrations/20230928084104_view_repo_ownership/migration.sql
@@ -1,0 +1,13 @@
+create or replace view view_repo_ownership as
+select ght.id        as "github_team_id"
+     , ght.name      as "github_team_name"
+     , tr.full_name  as "repo_name" --deprecated. use full_name below for consistency with other tables
+     , tr.full_name  as "full_name"
+     , tr.role_name
+     , tr.archived
+     , gtt.team_name as "galaxies_team"
+     , gtt.team_contact_email
+from github_team_repositories tr
+         join github_teams ght on tr.team_id = ght.id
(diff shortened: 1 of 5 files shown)


packages/repocop/prisma/migrations/20230928084104_view_repo_ownership/migration.sql
JuliaBrigitte, line 1: -> deprecated. use full_name below for consistency with other tables 
How will people know not to use this field unless they read the PR?

akash1810: Great question! I propose we do some work to migrate any uses and then drop the column. Worth noting that this the current definition of the view, so slightly out of scope for this change, so suggest we do that as a follow up PR.

NovemberTang (approved):

PR #340, opened 2023-10-02
Add hooks to protect .env file

NovemberTang:
## What does this change?

1. Sets up a precommit hook that verifies that a change to the original .env file is supposed to be committed
2. Sets up a pre-push hook that makes the original .env file read only

## Why?

There are legitimate reasons why someone may want to modify the original .env file. This includes version bumps for keeping cloudquery and its plugins up to date. However, we have had one too many instances of secrets being committed to this file. This PR adds a three layers of protection by

1. Requiring the developer to change the permissions of the file before editing it
2. Verifying that the changes were intentional
3. Removing write permissions at the end, so the next developer has to do the same

## How has it been verified?

Local testing. No further testing required as this is a local developer experince change


scripts/setup.sh
```diff
@@ -9,7 +9,7 @@ yellow='\033[1;33m'
 cyan='\033[0;36m'
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-ROOT_DIR=${DIR}/..
+ROOT_DIR=$(realpath "${DIR}/..")
```
NovemberTang, line 12: QOL change. Removes /.. from paths ie
`~/code/myscripts/../` -> `~/code/`

tjsilver (commented):
>> Currently, if the service catalogue detects a non-empty env.local file in the user's environment, it will not make any changes to it. This means that dependency updates are never propagated through to developers' laptops unless the file is deleted and recreated.

The `.env.local` on the user's machine was set up to only contains the user's tokens, it doesn't contain any of the dependencies, so this should not be an issue, IIRC.

scripts/setup.sh
```diff
@@ -102,6 +102,7 @@ GALAXIES_BUCKET=${GALAXIES_BUCKET}
     echo "No .env.local file found - creating it in $LOCAL_ENV_FILE_DIR and adding token names"
     mkdir -p "$HOME"/.gu/service_catalogue
     touch -a "$LOCAL_ENV_FILE_DIR"/.env.local
+    chmod +w "$LOCAL_ENV_FILE_DIR"/.env.local
```
tjsilver, line 105: Why is this required?

NovemberTang: thanks to your clarification, we know it's not needed, i'll remove it

scripts/setup.sh
```diff
@@ -122,7 +123,44 @@ Visit ${cyan}https://docs.snyk.io/snyk-api-info/authentication-for-api${clear},
...
+  PRE_COMMIT="${ROOT_DIR}/.git/hooks/pre-commit"
+  echo "$PRE_COMMIT_TEXT" > "$PRE_COMMIT"
+  chmod +x "$PRE_COMMIT"
+
+  echo "Setting up pre-push hook"
+  HOOK_TEXT="#!/usr/bin/env bash
+
+echo 'Running pre-push hook'
+if [ -w ${ROOT_DIR}/.env ]; then
+echo 'Removing write permissions from .env' file && chmod -w .env
```
akash1810, line 153: With this being performed in a pre-push, is the change in file mode land on remote, or does it need committing too?

scripts/setup.sh
```diff
@@ -122,7 +123,44 @@ Visit ${cyan}https://docs.snyk.io/snyk-api-info/authentication-for-api${clear},
   fi
 }
 
+setup_hooks() {
+  echo "Setting up pre-commit hook"
+  PRE_COMMIT_TEXT='#!/usr/bin/env bash
```
akash1810, line 128: The content of these hooks are a difficult to test/run on-demand. What if we made them files in their own right, and then `cat` them here to get their content? With them being their own files, the IDE will be able to provide syntax highlighting etc.

scripts/setup.sh
```diff
@@ -122,7 +123,44 @@ Visit ${cyan}https://docs.snyk.io/snyk-api-info/authentication-for-api${clear},
...
+PRE_PUSH="${ROOT_DIR}/.git/hooks/pre-push"
+echo "$HOOK_TEXT" > "$PRE_PUSH"
+chmod +x "$PRE_PUSH"
+}
+
 check_node_version
 install_dependencies
 check_credentials deployTools
 setup_cloudquery
+setup_hooks
```
akash1810, line 166: Might the [husky package](https://typicode.github.io/husky/) make configuring git hooks simpler?

akash1810 (commented):
The content of the hooks are no

scripts/setup.sh
NovemberTang: good point

akash1810 (approved):

NovemberTang (commented):
Turns out, git only stores the executable bit, not the read or write ones, so I'm going to close this PR

PR #349, opened 2023-10-06
pre-commit hook

NovemberTang:
## What does this change?

Creates a pre-commit hook. If staged changes to the `.env` file are detected, it asks the user to verify that this change was deliberate.

This change is based on the proposed changes in #340, but without the read/write permissions changes, as git does not store information about the `r` or `w` bits of files

## Why?

This makes it less likely a user will commit secret information

## How has it been verified?

Locally tested with an unmodified and modified `.env`


.hooks/pre-commit
```diff
@@ -0,0 +1,15 @@
...
+clear="\033[0m"
+yellow="\033[1;33m"
+#check for staged changes in .env and ask user to confirm
+if [[ $(git diff --cached --name-only | grep .env) ]]; then
+  echo -e "${red}You have staged changes in the original .env file. This might include secrets you do not intend to commit."
+  echo -e "${yellow}Are you sure you want to commit these changes?${clear} [y/N] "
+  exec < /dev/tty
+  read REPLY
+  if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+    echo -e "${red}Aborting commit.${clear} Please restore unintended changes before committing again."
```
AshCorr, line 12:
```suggestion
    echo -e "${red}Aborting commit.${clear} Please discard unintended changes with `git restore .env` before committing again."
```
Could give users the command to quickly discard the changes?

AshCorr (approved):

PR #351, opened 2023-10-06
make setup process clearer

NovemberTang:
## What does this change?

Makes setup process easier to follow. Use an incrementing number for `STEP_COUNT` so functions can be run in any order

## Why?

Make it clearer what is happening when

## How has it been verified?

Output now looks like this
New setup output with enumerated steps and clear end.

Previously it looked more like this
Old setup output

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -102,7 +102,7 @@ GALAXIES_BUCKET=${GALAXIES_BUCKET}
     # Check if file is empty - don't want to overwrite any existing tokens
     if [ "$FILE_SIZE" -gt "$SIZE_THRESHOLD" ]
     then
-      echo "Detected env.local. No changes made."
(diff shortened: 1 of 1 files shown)


scripts/setup.sh
```diff
@@ -93,7 +102,7 @@ GALAXIES_BUCKET=${GALAXIES_BUCKET}
     # Check if file is empty - don't want to overwrite any existing tokens
     if [ "$FILE_SIZE" -gt "$SIZE_THRESHOLD" ]
     then
-      echo "Non-empty .env.local file found in $LOCAL_ENV_FILE_DIR. No changes made"
+      echo "Detected env.local. No changes made."
```
tjsilver, line 105: Can we add 'non-empty' here to differentiate from finding an empty file in the else statement?

akash1810 (approved):
💎

img

PR #358, opened 2023-10-10
Auth via GithubApp

NovemberTang:
## What does this change?

We have created a secret to contain these credentials, and pull them down as part of the lambda. This also means that a local user will not have to store all of these sensitive values as environment variables.

## Why?

This code has only run locally, using a Personal Access Token. In preparation for running this on AWS as a lambda, we're moving to github app auth. 

## How has it been verified?

The secret has been populated, and running the code locally works
`npm -w branch-protector start`

Still to do:

- Create separate code and prod GitHub Apps
- Run code as a lambda


packages/branch-protector/src/config.ts
```diff
@@ -1,18 +1,72 @@
...
 	/**
-	 * Github Personal Access Token.
-	 *
-	 * If the `GITHUB_ACCESS_TOKENB` environment variable is not set, this won't work.
+	 * Auth configuration for the GitHub App.
 	 */
-	githubAccessToken: string;
+	githubAppConfig: {
+		strategyOptions: StrategyOptions;
+		installationId: string | number;
```
akash1810, line 16: This typing is interesting... Any reason we need `string`?

packages/branch-protector/src/config.ts
```diff
@@ -1,18 +1,72 @@
...
+	installationId: string;
+}
+
+async function getGithubAppSecretJson(): Promise<GithubAppSecret> {
+	const secretsManager = new SecretsManager();
+
+	const secret = await secretsManager.getSecretValue({
+		SecretId:
+			'/CODE/deploy/service-catalogue/branch-protector-github-app-secret',
+	});
```
akash1810, line 52: Should this be stage aware?

packages/branch-protector/src/run-locally.ts
```diff
@@ -2,6 +2,9 @@ import type { UpdateBranchProtectionEvent } from './model';
 import { main } from './index';
 
 if (require.main === module) {
+	process.env.AWS_PROFILE = 'deployTools';
+	process.env.AWS_REGION = 'eu-west-1';
```
akash1810, line 6: Are we setting these as an alternative to a [credential provider chain](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html)?

akash1810:
> ts-node packages/branch-protector/src/run-locally.ts

minor: we could shorten this via `npm -w branch-protector start`, as the `start` script is making the `ts-node` call.

packages/branch-protector/src/config.ts
NovemberTang: yes, but not at this stage, as there is no lambda yet.

packages/branch-protector/src/run-locally.ts
NovemberTang: yeah, this was the simplest way to get started. once we actually create the lambda on AWS we can migrate

akash1810 (approved):
A non-blocking comment on a typing. LGTM!

PR #359, opened 2023-10-11
Prisma migrate scripts

tjsilver:
## What does this change?
- Provides NPM scripts for Prisma migration to code and prod environments.
- The code and prod migration scripts require confirmation in order to continue, just to be extra safe.
- Provides an NPM script for local setup of Prisma.
- The scripts create the `DATABASE_URL` environment variable that Prisma requires. In the case of code and prod, these are created from secret retrieved from Secrets Manager.
- Updates the dev script for running locally, separating out the reset function from the adding a new migration function.
- Updates the README to reflect the changes.

## Why?
- Prisma migration requires the setting of the `DATABASE_URL` environment variable. This previously relied on users putting the correct values into their environment, which is far from ideal and risks committing secrets.
- This should make it easier for future travelers to get started with Repocop and Prisma.

## How has it been verified?
Tested locally and on code. I have run the scripts for prod without the VPN to ensure the scripts run correctly (but the migration was not attempted for real). The code and prod scripts are essentially the same and the migration completed successfully on code.

### DIFF
--- a/packages/repocop/script/migrate.sh
+++ b/packages/repocop/script/migrate.sh
@@ -13,6 +13,16 @@ migrateDEV() {
 }
 
 migrateCODE() {
+
+  allSecrets=$(aws secretsmanager list-secrets --profile deployTools --region eu-west-1 | jq '.SecretList')
+  secret_id=$(jq -r '.[] | select(.Tags[] as $t | any($t.Key=="Stack" and $t.Value=="deploy")) | select(.Tags[] as $t | any($t.Key=="Stage" and $t.Value=="CODE")) | select(.Tags[] as $t | any($t.Key=="App" and $t.Value=="service-catalogue")) | select(.Name | test("PostgresInstance")) | .Name' <<< "$allSecrets")
+  dbSecretJson=$(aws secretsmanager get-secret-value --secret-id "$secret_id" --profile deployTools --region eu-west-1 | jq -c '[.SecretString | fromjson]')
+  db_username=$(jq -r '.[] | .username' <<< "$dbSecretJson")
+  db_password=$(jq -r '.[] | .password' <<< "$dbSecretJson")
+  db_host=$(jq -r '.[] | .host' <<< "$dbSecretJson")
+  db_port=$(jq -r '.[] | .port' <<< "$dbSecretJson")
+
+  export DATABASE_URL="postgresql://$db_username:$db_password@$db_host:$db_port/postgres"
   echo "Migrating CODE"
 }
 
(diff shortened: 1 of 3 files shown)


packages/repocop/script/migrate.sh
```diff
@@ -2,36 +2,85 @@
...
+    else
+        echo "Exiting without migrating, goodbye."
+        return
+    fi
+}
+
 migrateCODE() {
-  echo "Migrating CODE"
+  echo -e "$note"
+  echo -e "${red}WARNING: Changes made on your local machine will affect Service Catalogue CODE data.${plain}"
```
NovemberTang, line 55: Minor point, but to me, red implies an error rather than a warning. Could we use yellow/orange instead?

packages/repocop/script/migrate.sh
```diff
@@ -2,36 +2,85 @@
...
+  db_port=$(jq -r '.[] | .port' <<< "$dbSecretJson")
+
+  export DATABASE_URL="postgresql://$db_usermigration_name:$db_password@$db_host:$db_port/postgres"
+}
+
+note="NOTE: You need to be on the VPN to migrate. \nMigration may fail if the table(s) already exist in the database.\nSee https://www.prisma.io/docs/guides/migrate/production-troubleshooting#failed-migration for how to fix."
+
+continueQuery() {
+  stage=$1
+  echo -e "${bold}Do you want to continue (y/n)?${plain}"
```
NovemberTang, line 41:
```suggestion
  echo -e "${bold}Do you want to continue (y/N)?${plain}"
```
make clearer that `No` is the default

packages/repocop/script/migrate.sh
```diff
@@ -2,36 +2,85 @@
...
+        npx prisma migrate deploy
+    else
+        echo "Exiting without migrating, goodbye."
+        return
+    fi
+}
+
 migrateCODE() {
-  echo "Migrating CODE"
+  echo -e "$note"
```
NovemberTang, line 54: migrateCODE and migratePROD look similar enough that they could probably be a single function that takes a stage parameter. Is there a reason why we've done this for `continueQuery` and `getDbUrl` but not `migrate`?

packages/repocop/script/migrate.sh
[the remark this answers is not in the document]
tjsilver: Yes good point.

packages/repocop/script/migrate.sh
tjsilver: I did think about this - and wasn't sure. The messaging is different (and the stage). I'll take another look at it!

NovemberTang (approved):

PR #360, opened 2023-10-11
Set up branch protection app as a lambda

NovemberTang:
## What does this change?

- Branch protector now reads off an SQS topic.
- The schedule of this lambda has initially been set to weekly, while we are still testing  but this will change as we find the optimum cadence.
- Functions have been pulled out to separate files based on what APIs they are calling.

This looks like a much larger PR than it is, due to bits of code being moved about, and a large snapshot. The main functional change is the introduction of the `getEvents` method, to read items in the queue

We are also pulling the anghammarad ARN as part of the setup script

Local execution does still work, but the developer experience has been degraded slightly as some environment variables need to be manually retrieved and set. This will be addressed in a subsequent PR

TODO: 
- Set QUEUE_URL as part of local setup
- Set GITHUB_APP_SECRET (the ARN of the secret, not the credentials themselves) as part of local setup
- Create Github App to authenticate on `PROD`

## How has it been verified?

Deployed to `CODE`

Submitting a message such as
```json
{
"fullName": "X"
"teamNameSlugs": ["Y", "Z"]
}
```
to the queue, and triggering the lambda, works.

### DIFF
--- a/packages/branch-protector/src/aws-requests.ts
+++ b/packages/branch-protector/src/aws-requests.ts
@@ -1,0 +1,50 @@
+import { ReceiveMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
+import { Anghammarad, RequestedChannel } from '@guardian/anghammarad';
+import type { Config } from './config';
+import type { UpdateBranchProtectionEvent } from './model';
+
+export async function getEvents(
+	msgCount: number,
+	config: Config,
+): Promise<UpdateBranchProtectionEvent[]> {
+	const command = new ReceiveMessageCommand({
+		QueueUrl: config.queueUrl,
+		MaxNumberOfMessages: msgCount,
+		WaitTimeSeconds: 20,
+	});
+
+	const sqsClient = new SQSClient({});
+
+	const result = await sqsClient.send(command);
+	if (result.Messages === undefined) {
+		console.log('No messages found');
+		return [];
+	} else {
+		const messages = result.Messages.map((msg) => msg.Body)
+			.filter((msg): msg is string => !!msg)
+			.map((msg) => JSON.parse(msg) as UpdateBranchProtectionEvent);
+		return messages;
+	}
+}
+
+export async function notify(
+	fullRepoName: string,
+	topicArn: string,
+	slug: string,
+) {
+	const client = new Anghammarad();
+	await client.notify({
+		subject: 'Hello',
+		message: `Branch protection has been applied to ${fullRepoName}`,
+		actions: [
+			{
+				cta: 'Check branch protections here',
+				url: `https://github.com/${fullRepoName}/settings/branches`,
+			},
+		],
+		target: { GithubTeamSlug: slug },
+		channel: RequestedChannel.PreferHangouts,
+		sourceSystem: 'branch-protector',
+		topicArn: topicArn,
+	});
+}
--- a/packages/branch-protector/src/github-requests.ts
+++ b/packages/branch-protector/src/github-requests.ts
@@ -1,0 +1,70 @@
+import { createAppAuth } from '@octokit/auth-app';
+import { Octokit } from 'octokit';
+import type { Config } from './config';
+import type { UpdateBranchProtectionParams } from './model';
+
+export async function getGithubClient(config: Config) {
+	const auth = createAppAuth(config.githubAppConfig.strategyOptions);
+
+	const installationAuthentication = await auth({
+		type: 'installation',
+		installationId: config.githubAppConfig.installationId,
+	});
+
+	const octokit: Octokit = new Octokit({
+		auth: installationAuthentication.token,
+	});
+	return octokit;
+}
+
+export async function updateBranchProtection(
+	octokit: Octokit,
+	owner: string,
+	repo: string,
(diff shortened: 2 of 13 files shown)


packages/branch-protector/src/github-requests.ts
```diff
@@ -0,0 +1,70 @@
+import { createAppAuth } from '@octokit/auth-app';
```
NovemberTang, line 1: none of the code in this file is new, it's all been moved from `index.ts`

packages/branch-protector/src/aws-requests.ts
```diff
@@ -0,0 +1,50 @@
...
+		return [];
+	} else {
+		const messages = result.Messages.map((msg) => msg.Body)
+			.filter((msg): msg is string => !!msg)
+			.map((msg) => JSON.parse(msg) as UpdateBranchProtectionEvent);
+		return messages;
+	}
+}
+
+export async function notify(
```
NovemberTang, line 30: this function is not new, it's been moved from `index.ts`. However, `getEvents` is new

packages/cdk/lib/service-catalogue.ts
```diff
@@ -525,5 +534,60 @@ export class ServiceCatalogue extends GuStack {
...
+			queueName: 'branch-protector-queue.fifo',
+			contentBasedDeduplication: true,
+			fifo: true, //defaults to false for some reason
+		});
+
+		const branchProtectorLambdaProps: GuScheduledLambdaProps = {
+			app: 'branch-protector',
+			fileName: 'branch-protector.zip',
+			handler: 'index.main',
+			monitoringConfiguration: {
```
NovemberTang, line 562: monitoring configuration should be stage aware in future

packages/cdk/lib/service-catalogue.ts
```diff
@@ -525,5 +534,60 @@ export class ServiceCatalogue extends GuStack {
...
+			{
+				secretName: `/${stage}/${stack}/${app}/branch-protector-github-app-secret`,
+			},
+		);
+
+		const anghammaradTopicParameter = StringParameter.fromStringParameterName(
+			this,
+			'anghammarad-topic',
+			'/account/services/anghammarad.topic.arn',
+		);
```
akash1810, line 550: There's some Anghammarad stuff in GuCDK which we could use:
  - `GuAnghammaradTopicParameter`
  - `GuAnghammaradSenderPolicy`

```ts
const anghammaradTopicParameter = GuAnghammaradTopicParameter.getInstance(scope);
```

packages/cdk/lib/service-catalogue.ts
```diff
@@ -525,5 +534,60 @@ export class ServiceCatalogue extends GuStack {
...
+			timeout: Duration.minutes(1),
+		};
+
+		const branchProtectorLambda = new GuScheduledLambda(
+			this,
+			'branch-protector',
+			branchProtectorLambdaProps,
+		);
+
+		branchProtectorQueue.grantConsumeMessages(branchProtectorLambda);
```
akash1810, line 585: Does RepoCop need write permissions to this queue too?

scripts/setup.sh
```diff
@@ -81,13 +81,19 @@ setup_cloudquery() {
 
   GALAXIES_BUCKET=$(aws ssm get-parameter --name /INFRA/deploy/services-api/galaxies-bucket-name --profile deployTools --region eu-west-1 | jq '.Parameter.Value' | tr -d '"')
 
+  ANGHAMMARAD=$(aws ssm get-parameter --name /account/services/anghammarad.topic.arn --profile deployTools --region eu-west-1 | jq '.Parameter.Value' | tr -d '"')
```
akash1810, line 84: minor: We could use pass the `-r` flag to `jq`, and remove piping to `tr`:

> `--raw-output` / `-r`:
> With this option, if the filter's result is a string then it will be written directly to standard output rather than being formatted as a JSON string with quotes. This can be useful for making jq filters talk to non-JSON-based systems.
> – https://jqlang.github.io/jq/manual/

packages/cdk/lib/service-catalogue.ts
```diff
@@ -525,5 +534,60 @@ export class ServiceCatalogue extends GuStack {
...
+				toleratedErrorPercentage: 50,
+				lengthOfEvaluationPeriod: Duration.minutes(1),
+				numberOfEvaluationPeriodsAboveThresholdBeforeAlarm: 1,
+				snsTopicName: 'devx-alerts',
+			},
+			rules: [{ schedule: Schedule.rate(Duration.days(7)) }],
+			runtime: Runtime.NODEJS_18_X,
+			environment: {
+				GITHUB_APP_SECRET: branchProtectorGithubCredentials.secretName,
+				ANGHAMMARAD: anghammaradTopicParameter.stringValue,
```
akash1810, line 572: minor: `ANGHAMMARAD` is a little ambiguous (is it the repo, the SNS topic name, the SNS topic ARN?). `ANGHAMMARAD_SNS_ARN` might be more intention revealing?

tjsilver (approved):
Pair reviewed with @NovemberTang 👍

PR #361, opened 2023-10-12
chore(lint): Address some lint warnings

akash1810:
## What does this change?
I spotted some lint warnings in the build of some PRs, this change aims to resolve them to reduce noise.


tjsilver (approved):
Nice!

PR #365, opened 2023-10-13
Find owners for repos that fail REPOSITORY_02

NovemberTang:
## What does this change?

This PR introduces a feature that selects 5 random `production` or `documentation` repositories[^1] that fail REPOSITORY_02, and finds the github slugs of their owners. We need the slugs as anghammarad uses this information to figure out where to send notifications.

All this new code was crowding `packages/repocop/src/index.ts`, so a lot of it has been moved, or put into functions to keep `main()` clear and understandable.

Tests have also been refactored so not all the functions in `repository.ts` need to be exposed

## Why?

We need a mechanism to notify teams that we are going to change their branch protections.

## How has it been verified?

Lambda has been deployed to CODE, and repos/owners show up in the logs, as intended

[^1]: we are looking only at production and documentation repos at the moment as they are the most important. This restriction will loosen as we progress through the list of noncompliant repos


packages/repocop/src/remediations/repository-02.test.ts
```diff
@@ -0,0 +1,86 @@
...
+	github_team_id: BigInt(0),
+	github_team_name: '',
+	repo_name: '',
+	role_name: '',
+	archived: false,
+	galaxies_team: null,
+	team_contact_email: null,
+};
+
+const nullTeam: github_teams = {
```
NovemberTang, line 15: these kinds of template objects are now duplicated across multiple test files. In another PR we should import them from one location, keeping to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#:~:text=%22Don't%20repeat%20yourself%22,redundancy%20in%20the%20first%20place.) principle

packages/repocop/src/index.ts
```diff
@@ -1,31 +1,61 @@
...
+		productionOrDocs.includes(repo.full_name),
 	);
+
+	const msgs = createRepository02Messages(relevantRepos, repoOwners, teams);
+	await notifyAndAddToQueue(msgs);
+}
+/*
+ *TODO - implement this
+ * For CODE, we should add a message to the queue but should never send a notification.
+ * Similarly, the branch-protector on CODE should log something,but should never update
```
NovemberTang, line 49: the branch protector part of this has been handled in #366

tjsilver (approved):
👍

PR #366, opened 2023-10-13
make branch protector stage aware

NovemberTang:
## What does this change?

Prevents non-prod environments from making changes to repos.

## Why?

1. We don't want a lambda intended for testing to send real messages to teams or make real changes to repos
2. The github app for CODE shouldn't have permission to make changes to repos, but having this control adds an extra layer of safety
3. If the code github app doesn't have permission to make changes to repos, it's pretty likely `updateBranchProtection` will throw an error, so we should try to avoid calling it entirely.

## How has it been verified?

TODO: add a message to the queue and invoke the lambda on CODE or DEV

### DIFF
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -22,21 +22,29 @@ async function protectBranch(
 	}
 
 	const defaultBranchName = await getDefaultBranchName(owner, repo, octokit);
-	const isProtected = await isBranchProtected(
+	const branchIsProtected = await isBranchProtected(
 		octokit,
 		owner,
 		repo,
 		defaultBranchName,
 	);
 
-	if (isProtected) {
-		console.log(`${repo}'s default branch is protected. No action required`);
-	} else {
+	console.log(`${repo} branch protection: ${branchIsProtected.toString()}`);
+
+	const stageIsProd = config.stage === 'PROD';
+
+	if (branchIsProtected && stageIsProd) {
+		console.log(`No action required`);
+	} else if (!branchIsProtected && stageIsProd) {
 		await updateBranchProtection(octokit, owner, repo, defaultBranchName);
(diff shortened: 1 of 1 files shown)


packages/branch-protector/src/index.ts
```diff
@@ -22,21 +22,29 @@ async function protectBranch(
...
-	if (isProtected) {
-		console.log(`${repo}'s default branch is protected. No action required`);
-	} else {
+	console.log(`${repo} branch protection: ${branchIsProtected.toString()}`);
+
+	const stageIsProd = config.stage === 'PROD';
+
+	if (branchIsProtected && stageIsProd) {
+		console.log(`No action required`);
+	} else if (!branchIsProtected && stageIsProd) {
```
akash1810, line 38: minor: I think we can refactor this to be a bit simpler, with single `if` conditions:

```ts
if(!stageIsProd) {
  console.log(`Detected stage: ${config.stage}. No action taken.`);
}
else {
  if(branchIsProtected) {
    console.log(`No action required`);
  }
  else {
    await updateBranchProtection();
    ...
  }
}
```

akash1810 (approved):

packages/branch-protector/src/index.ts
NovemberTang: Initially it was set up that way, but I think avoiding nesting, and having all the conditions laid out on one line makes for slightly easier reading, and lower cognitive complexity

akash1810: Yeah, the nesting is a bit horrible; early returns help here:

```ts
// Only PROD affects things IRL
if(!stageIsProd) { 
  console.log(`Detected stage: ${config.stage}. No action taken.`); 
  return;
}

// Nothing to do
if(branchIsProtected) { 
  console.log(`No action required`); 
  return;
}

// Something to do
await updateBranchProtection();
...
```

My comments are not blocking though.

PR #367, opened 2023-10-13
fix: Apply Tags to running tasks

akash1810:
Apply tags to running tasks to increase service discovery ability.

See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-tags.

## How has it been verified?
TBD.


akash1810:
This isn't working 😢

image

PR #368, opened 2023-10-13
chore(deps): Use logging solution from devx-logs

akash1810:
https://github.com/guardian/devx-logs/pull/32 should be merged first, to ensure log lines have the correct markers in Central ELK.

## What does this change?
Replaces the logging container with the one from [devx-logs](https://github.com/guardian/devx-logs/tree/main/ecs).

## Why?
The devx-logs solution is more "official", and with this change we can archive https://github.com/guardian/hackday-firelens.

## How has it been verified?
https://github.com/guardian/service-catalogue/pull/369 tests https://github.com/guardian/devx-logs/pull/32. Once deployed to CODE, the logs in Central ELK look like this:

## Before
image

## After
image

### DIFF
--- a/packages/cdk/lib/ecs/task.ts
+++ b/packages/cdk/lib/ecs/task.ts
@@ -29,7 +29,7 @@ const cloudqueryImage = ContainerImage.fromRegistry(
 );
 
 const firelensImage = ContainerImage.fromRegistry(
-	'ghcr.io/guardian/hackday-firelens:main',
+	'ghcr.io/guardian/devx-logs:main',
 );
 
 export interface ScheduledCloudqueryTaskProps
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/ecs/task.ts
```diff
@@ -29,7 +29,7 @@ const cloudqueryImage = ContainerImage.fromRegistry(
 );
 
 const firelensImage = ContainerImage.fromRegistry(
-	'ghcr.io/guardian/hackday-firelens:main',
+	'ghcr.io/guardian/devx-logs:main',
```
AshCorr, line 32: Not for this PR, but perhaps in future we could consider using a specific version of `devx-logs` instead of always using `main`? We'd have issues if we ever needed to release a breaking version of devx-logs!

AshCorr (approved):

PR #370, opened 2023-10-16
Update Cloudquery AWS plugin to 22.15.0

AshCorr:
## What does this change?

Bumps the Cloudquery AWS plugin to 22.15.0

## Why?

We fixed the bug in this version of CQ AWS which was causing some old buckets to not appear in our S3 buckets table.

https://github.com/cloudquery/cloudquery/pull/14476


tjsilver (approved):

tjsilver:
I think the snapshots might need updating as they're still referencing the previous version of CQ AWS plugin.

akash1810 (approved):

PR #371, opened 2023-10-16
feat: add best practices markdown and generator script

tjsilver:
## What does this change?
- Adds the best practices markdown file to this repo and converts its Deno generator script to a Node script. It is currently copied over from `guardian/recommendations` but eventually will be removed from there and linked to.
- Creates a new package called `best-practices` to house the files.
- Converts the project from Deno to a NPM workspace that uses `ts-node` (ESM - this was required because the NPM package `markdown-table` on which the generator relies is ESM-only).
- Updates the documentation to reflect the changes.

## Why?
- We want to keep the best practices up-to-date and have a single source of truth. Currently they are copied into the [Repocop Grafana dashboard](https://metrics.gutools.co.uk/d/2uaV8PiIz/repocop?orgId=1) and there is a risk of drift and repetition of work. The plan is to generate both the markdown and the table in the database for Grafana from the same source. This is the first step towards that.
- It was converted from Deno because Snyk does not support vulnerability scanning in Deno. The rest of the project is NPM workspaces and it also makes sense to keep it in line with that.

## How has it been verified?
The best practices table was deleted in the markdown file, then the generator script was run locally using `npm -w best-practices run generate `. This successfully re-generated the table in the markdown file.


packages/best-practices/generator/generate.sh
akash1810, line 1: With these files being added as a `package` (i.e. an NPM workspace), do we want to use Deno?

packages/best-practices/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "module": "es6",
+    "allowImportingTsExtensions": true
```
akash1810, line 5: https://www.youtube.com/watch?v=8ORIzvgNWhU is a good watch here.

akash1810 (approved):

PR #380, opened 2023-10-17
Repocop add repository_02 messages to queue

tjsilver:
Co-authored-by: @NovemberTang 

## What does this change?

The `repository_02` rule says that 'all production and documentation repos must have a protected default branch'.

- Randomly selects 2 P&E  repos with topics of `production` or `documentation` that fail the `repository_02` test (this number is configurable)
- Creates a message for each one containing the repo name and the team slugs of the repo admin(s)
- Adds the message to the `branch-protector` queue
- Creates a notification to teams that branch protection will be applied via Anghammarad

## Why?

We want to automate the remediation of repositories that fail the rule, and let teams know that we're doing it.  The recent UX research indicated that teams would generally be happy for us to remediate this issue without their direct involvement.

## How has it been verified?

Tested on CODE.

## TODO

- Choose how many repos to act on in one go for CODE, and PROD, by using a stage aware env var in CDK
- Choose a schedule for the repocop and branch protector lambdas that doesn't overwhelm teams, and leaves enough time before warning and resolution (ideally about a day)


packages/repocop/src/index.ts
```diff
@@ -40,20 +47,22 @@ async function writeRepo02Messages(
...
+	events: UpdateBranchProtectionEvent[],
+	config: Config,
+	anghammaradClient: Anghammarad,
+) {
+	await addMessagesToQueue(events, config);
+	if (config.stage === 'PROD') {
+		await sendNotifications(anghammaradClient, events, config);
+	} else {
+		console.log('Messages added to queue but notifications NOT sent');
+	}
```
akash1810, line 65: Is this flow correct? IIUC RepoCop is not the one that makes changes to GitHub, that is handled asynchronously by branch protector. I'd suggest branch protector be the one to send a notification via Anghammarad, as that's the only time we'll know if the operation was successful.

This division provides clearer roles:
- RepoCop identifies things that need action
- Branch protector performs the action, and once complete, notifies the owning team

packages/repocop/src/remediations/repository-02.ts
```diff
@@ -55,7 +61,73 @@ export function createRepository02Messages( //TODO: test this function
...
+}
+
+async function notifyOneRepo(
+	anghammaradClient: Anghammarad,
+	event: UpdateBranchProtectionEvent,
+	config: Config,
+) {
+	for (const slug of event.teamNameSlugs) {
+		await notifyOneTeam(anghammaradClient, event.fullName, slug, config);
+	}
```
akash1810, line 120: I think we'll need a [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) here so that we only return once all calls to `notifyOneTeam` have finished:

```ts
await Promise.all(
  event.teamNameSlugs.map(slug => notifyOneTeam(anghammaradClient, event.fullName, slug, config))
);
```

packages/repocop/src/remediations/repository-02.ts
```diff
@@ -55,7 +61,73 @@ export function createRepository02Messages( //TODO: test this function
...
+}
+
+export async function sendNotifications(
+	anghammaradClient: Anghammarad,
+	events: UpdateBranchProtectionEvent[],
+	config: Config,
+): Promise<void> {
+	for (const event of events) {
+		await notifyOneRepo(anghammaradClient, event, config);
+	}
```
akash1810, line 132: Think we'll need a `Promise.all` here too.

packages/repocop/src/index.ts
NovemberTang: That division already exists, they just both send a notification. Repocop sends a message informing teams that we will be taking action on their behalf. Branch protector sends a second message to the same teams when the action is complete. We've done it this way as teams have said that they would like to know about changes we're making on their behalf before action, so they have time to understand, or object.

NovemberTang: #382 explains this better

AshCorr (approved):

PR #381, opened 2023-10-17
feat(cloudquery): Collect remaining GitHub tables

akash1810:
## What does this change?
In https://github.com/guardian/service-catalogue/pull/372 we bumped the GitHub source plugin, noting that it provides more data.

This change creates a `RemainingGitHubData` task, which will collect those tables that are not part of a discrete job. This is similar to the `RemainingAwsData` task.

## Why?
See https://github.com/guardian/service-catalogue/pull/372 for some possible questions this can help answer.

## How has it been verified?
I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/1867ab04-f2dc-44f8-8374-655382451b38), ran the task, and can see the [data in Grafana](https://metrics.code.dev-gutools.co.uk/goto/2s2ThinIz?orgId=1).

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -10529,6 +10529,678 @@ spec:
       },
       "Type": "AWS::Events::Rule",
     },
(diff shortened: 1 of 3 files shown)


NovemberTang (approved):
Do we need to make prisma aware of these new tables?

PR #383, opened 2023-10-18
docs: update repocop README

tjsilver:
## What does this change?
- Edits the repocop README to make information about the `DATABASE_URL` (hopefully!) clearer.

## Why?
- It was a little confusing mentioning the format of the url (implying that the user will have to set it) when it is set automatically by a script now.

## How has it been verified?
Is it easier to understand?


JuliaBrigitte (approved):
Much less confusing, thank you.

PR #387, opened 2023-10-19
burn after reading

NovemberTang:
## What does this change?

Previously, we were reading the messages off of SQS, and then letting them sit there. This meant that we were reading the same messages over and over again. Now we delete messages after branch protection has been successfully applied.

## Why?

This prevents a backlog accumulating, as we can now read new messages off of the queue.

## How has it been verified?

Tested on DEV. NB that even for local invocations, we are interacting with real AWS resources, namely reading/deleting messages from the CODE queue.

### DIFF
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -1,5 +1,7 @@
+import type { Message } from '@aws-sdk/client-sqs';
+import { SQSClient } from '@aws-sdk/client-sqs';
 import type { Octokit } from 'octokit';
-import { getEvents, notify } from './aws-requests';
+import { deleteFromQueue, notify, readFromQueue } from './aws-requests';
 import { getConfig } from './config';
 import type { Config } from './config';
 import {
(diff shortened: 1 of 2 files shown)


packages/branch-protector/src/index.ts
```diff
@@ -48,12 +50,29 @@ async function protectBranch(
 	}
 }
 
+function createEvents(
```
NovemberTang, line 53: this function used to be a part of getEvents, but now it's not calling AWS directly, it's no longer in that file. I'm not sure where the best place for it is, so leave it in index.ts for now

packages/branch-protector/src/index.ts
```diff
@@ -48,12 +50,26 @@ async function protectBranch(
...
+	const sqsClient = new SQSClient({});
 
-	const events = await getEvents(1, config);
-	for (const event of events) {
-		await protectBranch(octokit, config, event);
-	}
+	const messages = await readFromQueue(config, 1, sqsClient);
+	const events = createEvents(messages);
+	await Promise.all(
+		events.map(async (event) => await protectBranch(octokit, config, event)),
```
AshCorr, line 70:
```suggestion
		events.map((event) => protectBranch(octokit, config, event)),
```

If any call to protectBranch fails (Github API outage seems pretty common these days!) would this mean that events that were handled successfully won't be removed from the queue?

NovemberTang: yes. good point!

tjsilver (approved):

PR #388, opened 2023-10-19
Refactor: Change branch protector schedule

tjsilver:
## What does this change?

- Updates the schedule for the `branch-protector` lambda to run every morning at 9am Tue-Fri only. 
- Makes a small change to the way the `GITHUB_APP_SECRET` env var is retrieved.

## Why?
- We don't want people being notified about their branch protection being enabled on a non-working day. Starting the lambda on a Tuesday allows for pre-messaging to go out on the Monday also.


louishather (approved):
👍 Looks good!

PR #391, opened 2023-10-23
Fixed running repocop locally and improved logic on what to run in which setting

JuliaBrigitte:
## What does this change?
We are able to run repocop locally again
We changed the logic and what to run in which environment to make it easier to understand.
PROD talks to branch_protector and the anghammarad
CODE talks to branch_protector
DEV does not talk to anything by default, but you can set ENABLE_MESSAGING=true for testing purposes

## Why?
To be able to test locally

## How has it been verified?
It worked locally.

### DIFF
--- a/.env
+++ b/.env
@@ -26,11 +26,14 @@ CQ_SNYK=3.1.5
 # See https://github.com/guardian/cq-source-snyk-full-project
 CQ_GUARDIAN_SNYK=0.1.1
 
-# For local development only
+
+# --- FOR LOCAL DEVELOPMENT ONLY ---
+
 LOCAL_STAGE=DEV
 LOCAL_DB_USER=postgres
 LOCAL_DB_PASSWORD=not_at_all_secret
 LOCAL_DB_HOSTNAME=localhost
 LOCAL_DB_PORT=5432
 
+# Enables messaging from Repocop to Anghammarad for local testing
 ENABLE_MESSAGING=false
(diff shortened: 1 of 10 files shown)


packages/cloudquery/README.md
```diff
@@ -52,6 +53,8 @@ It includes:
 > **Note**
 > You can also use other Postgres clients, such as `psql` to query the data, or even your IDE!
 
+To configure your db connection, go the RDS in the deploytool account and get the DB Identifier for code or prod, as 
+authentification for the postgres db you can use SecretsmanagerAuth with the arn for the secret.
```
akash1810, line 57: This doesn't feel related to this change. Is it needed? I'd advocate against connecting to RDS locally too, so unsure when this advice would be needed/followed?

.env
```diff
@@ -32,3 +32,5 @@ LOCAL_DB_USER=postgres
 LOCAL_DB_PASSWORD=not_at_all_secret
 LOCAL_DB_HOSTNAME=localhost
 LOCAL_DB_PORT=5432
+
+ENABLE_MESSAGING=false
```
akash1810, line 36: Worth adding an inline comment here to explain what this flag controls?

akash1810 (approved):
LGTM.

Let's deploy to CODE before merging to ensure the changes to the AWS credentials still work?

tjsilver:
> LGTM.
> 
> Let's deploy to CODE before merging to ensure the changes to the AWS credentials still work?

I deployed to CODE, and ran the repocop lambda. It successfully ran and added two messages to the CODE fifo queue for branch protector.  I also ran the branch protector lambda, and it picked one message off the queue and didn't take any further action because it was CODE. So I think we're good to go.

PR #398, opened 2023-10-24
refactor(repocop): Read straight from `.env` file to reduce duplication

akash1810:
## What does this change?
Removes some env var aliasing when running RepoCop locally, in favour of reading values directly from the `.env` file.

## Why?
It's simpler.

## How has it been verified?
N/A.


AshCorr (approved):

tjsilver:
This has broken running Cloudquery locally as the db vars are also used in the docker-compose.yml file. I will create a fix now.

PR #402, opened 2023-10-24
refactor: Obtain home directory via `os`

akash1810:
## What does this change?
Obtain the home directory from the `os` package. This allows us to remove an `eslint` disablement.


tjsilver (approved):
Great!

PR #405, opened 2023-10-25
fix: update env vars to enable running cloudquery and migrate locally

tjsilver:
## What does this change?
Updates the local development environment variables in the scripts that use them, following changes in #398.

## Why?
Running Cloudquery locally and the local migrate script were broken.

## How has it been verified?
Tested locally.


AshCorr (approved):
Good catch!

PR #408, opened 2023-10-26
revert: Stop collecting all GitHub data as it's too expensive

akash1810:
## What does this change?
Reverts #381 (and #385) as the `RemainingGitHubData` task is taking a long time - multiple days! This is related to the API quota limits being reached, and CloudQuery sleeping until they reset.

The task was scheduled to run every 4 hours too, meaning there are overlapping invocations, making the issue worse.

For heavily rate limited APIs, we should collect exactly what we need to reduce the chance of exceeding the rate limit.

## How has it been verified?
N/A - this is a reversion to a previous state.


tjsilver (approved):
Looks good! 👍

PR #412, opened 2023-10-26
Use a v1 of the devx-logs Image instead of latest

AshCorr:
## What does this change?

Use v1 of devx-logs

## Why?

I will be making some breaking changes in https://github.com/guardian/devx-logs/pull/34 which will cause `main` to break logging for service-catalogue . Also good practice to not always use the latest version of a docker image.


akash1810 (approved):
Looks like the snapshots need updating too.

AshCorr:
> Looks like the snapshots need updating too.

Ach I'm always forgetting!

PR #413, opened 2023-10-26
Use devx-logs v2

AshCorr:
## What does this change?

Consume v2 of our devx-logs image.

## Why?

We've switched to a different version of the aws-for-fluent-bit base image in this release of devx-logs. This new version allows us to add ECS metadata to our logs such as Task ARN, Task Definition, and ECS Cluster.

## How has it been verified?

From our logs

image


packages/cdk/lib/ecs/task.ts
```diff
@@ -187,11 +186,6 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 			},
 			firelensConfig: {
 				type: FirelensLogRouterType.FLUENTBIT,
-				options: {
-					enableECSLogMetadata: true,
-					configFileType: FirelensConfigFileType.FILE,
-					configFileValue: '/custom.conf',
-				},
```
AshCorr, line 194: `enableECSLogMetadata` is `true` by default

configFileType and configFileValue are no longer required as aws-for-fluent-bit relies on environment variables which are baked into our image now

https://github.com/aws/aws-for-fluent-bit/blob/mainline/use_cases/init-process-for-fluent-bit/README.md#how-to-use-multi-config-feature
https://github.com/guardian/devx-logs/blob/main/ecs/Dockerfile#L3-L13

akash1810 (approved):
Excellent work!

PR #414, opened 2023-10-26
Prisma migrations for five views already present on prod

JuliaBrigitte:
## What does this change?
This is bringing five views that have been applied to prod by running sql into the prisma migration history.

TODO There are two views on prod that are not yet in prisma, since there was no sql to create them in version control: elk_index_to_space and grafana_cloudwatch_to_internal_infra

The migrations have to be applied to prod after this PR is merged

## Why?
When running repocop locally the views need to be defined as a migration to appear. We also can only use the views in typescript if they are defined in schema.prisma.

## How has it been verified?
The migrations have been successfully applied to CODE


sql/views_info.txt
tjsilver, line 1: I can see why you went with separating out the views into separate files, but will this comment will have to be updated manually when views are added? There is a risk it could fall out of date if so.

tjsilver (approved):
This looks good and I have seen the migrations in CODE. I'm not clear on the history of why these tables were added via raw sql - but I wonder if we need to improve our processes/documentation so we don't have to keep doing these 'mop-up' jobs?

akash1810 (commented):
I'm not sure the changes to the `sql` directory are necessary, as each file is commented out. Do we need these files? Could we rely on the revision history instead?

sql/views_info.txt
JuliaBrigitte: Good point, I will delete them and put only a comment about the two missing ones there.

JuliaBrigitte:
I will clean up the sql directory explaining about the two missing views and the attempt a follow up PR to explain how to add view to the migration since I agree with @tjsilver that this is how the views should have been added in the first place.

PR #417, opened 2023-10-30
fix(cdk): Run the RemainingAwsData task each Monday at 10AM 

akash1810:
## What does this change?
Follows #416, and #410 to set the `RemainingAwsData` task to run at 10AM on Monday's. It was previously every minute, of every hour on Monday!

---

Co-authored-by: @tjsilver


tjsilver (approved):
Paired on this solution 👍

PR #418, opened 2023-10-30
refactor: friendly names for repocop rules

tjsilver:
## What does this change?

- Renames the Repocop best practice rules in the `repocop_github_repository_rules` from their IDs to their Names.
- Renames the rule evaluation functions to indicate what they do using the friendly names, and updates the function comments to be more accurate.

## Why?

It will be easier for users to visit the [Repocop Compliance](https://metrics.gutools.co.uk/d/EOPnljWIz/repocop-compliance?orgId=1) dashboard and understand which rules their team's repos are breaking, reducing the number of times they will have to access the [rule definitions](https://github.com/guardian/service-catalogue/blob/main/packages/best-practices/best-practices.md). This is supported by the results of the recent UX research.

## How has it been verified?
The migration has been applied on CODE and the renamed columns can be seen on this [CODE dashboard](https://metrics.code.dev-gutools.co.uk/d/CqXjemVSz/repocop-compliance-code?orgId=1).

*Note*
Once this PR is merged, the migration will need to be carried out manually on PROD.

### DIFF
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -12,18 +12,18 @@ import {
 } from '../query';
 
 /**
- * Apply the following rule to a GitHub repository:
+ * Evaluate the following rule for a Github repository:
  *   > The default branch name should be "main".
  */
-function repository01(repo: github_repositories): boolean {
+function hasDefaultBranchNameMain(repo: github_repositories): boolean {
 	return repo.default_branch === 'main';
 }
 
 /**
- * Apply the following rule to a GitHub repository:
+ * Evaluate the following rule for a Github repository:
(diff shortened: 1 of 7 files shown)


packages/repocop/src/rules/repository.ts
```diff
@@ -96,15 +96,14 @@ export function repositoryRuleEvaluation(
 
 	return {
 		full_name: fullName,
-		repository_01: repository01(repo),
-		repository_02: repository02(repo, allBranches),
-
+		default_branch_name: repository01(repo),
```
akash1810, line 99: Should we rename these functions too? From `repository01` to `checkDefaultBranch`?

tjsilver: Yes, good spot!

akash1810 (approved):

PR #424, opened 2023-10-31
Source deployment data from RiffRaff

AshCorr:
## What does this change?

Source riff-raff data such as deployment history into Cloudquery by using the riff-raff. This is done using the [Postgresql source plugin](https://www.cloudquery.io/docs/plugins/sources/postgresql/overview) to pull data from some [custom views](https://github.com/guardian/riff-raff/pull/1273) created in the riff-raff DB.

New tables:

- `riffraff_deploys` - List of all deploys
- `riffraff_deploy_logs` - History of each deployment
- `riffraff_authorized_history` - Users with access to riff-raff

Requires https://github.com/guardian/riff-raff/pull/1273 & https://github.com/guardian/deploy-tools-platform/pull/731 to be merged first.

## Why?

This lets us answer questions such as "Whats the latest version of an application running in production?" or "Is an application being deployed everyday?" or "What users have access to riff-raff but aren't part of a P&E team?"

## How has it been verified?

 - Deployed to CODE along with https://github.com/guardian/riff-raff/pull/1273 & https://github.com/guardian/deploy-tools-platform/pull/731

image

### DIFF
--- a/packages/cdk/lib/ecs/config.ts
+++ b/packages/cdk/lib/ecs/config.ts
@@ -25,7 +25,7 @@ export function postgresDestinationConfig(): CloudqueryConfig {
 			name: 'postgresql',
 			registry: 'github',
 			path: 'cloudquery/postgresql',
-			version: `v${Versions.CloudqueryPostgres}`,
+			version: `v${Versions.CloudqueryPostgresDestination}`,
 			migrate_mode: 'forced',
 			spec: {
 				connection_string: [
@@ -221,7 +221,7 @@ export function riffraffSourcesConfig(
 		spec: {
 			name: 'postgresql',
 			path: 'cloudquery/postgresql',
-			version: `v3.0.7`,
+			version: `v${Versions.CloudqueryPostgresSource}`,
 			destinations: ['postgresql'],
 			tables: ['riffraff_*'],
 			spec: {
@@ -231,8 +231,8 @@ export function riffraffSourcesConfig(
 					`host=${riffRaffDBAddress}`,
 					`port=${riffRaffDBPort}`,
 					'dbname=riffraff',
-					// Our riff-raff DB instances use an old SSL certificate that don't work with the version of Go that we use for service-catalogue.
-					// We cannot enable sslmode until we switch to using a different CA authority for riff-raff.
+					// Ideally we'd use sslmode=verify-full however the certificates used by riff-raffs DB are quite old and don't have any SANs set.
+					// In order to upgrade to verify-full we need to change the CA used by the riff-raff DB.
 					// See https://github.com/golang/go/issues/39568
 					'sslmode=verify-ca',
 				].join(' '),
--- a/packages/cdk/lib/ecs/config.ts
+++ b/packages/cdk/lib/ecs/config.ts
@@ -212,10 +212,7 @@ export function galaxiesSourceConfig(bucketName: string): CloudqueryConfig {
 	};
 }
 
-export function riffraffSourcesConfig(
-	riffRaffDBAddress: string,
-	riffRaffDBPort: string,
-): CloudqueryConfig {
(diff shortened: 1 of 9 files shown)


packages/cdk/lib/ecs/config.ts
```diff
@@ -212,6 +212,35 @@ export function galaxiesSourceConfig(bucketName: string): CloudqueryConfig {
...
+				connection_string: [
+					'user=${RIFFRAFF_DB_USERNAME}',
+					'password=${RIFFRAFF_DB_PASSWORD}',
+					`host=${riffRaffDBAddress}`,
+					`port=${riffRaffDBPort}`,
+					'dbname=riffraff',
+					// Ideally we'd use sslmode=verify-full however the certificates used by riff-raffs DB are quite old and don't have any SANs set.
+					// In order to upgrade to verify-full we need to change the CA used by the riff-raff DB.
+					// See https://github.com/golang/go/issues/39568
+					'sslmode=verify-ca',
```
akash1810, line 237: https://eu-west-1.console.aws.amazon.com/rds/home?region=eu-west-1#ca-cert-update: is flagging that we need to rotate the CA on a number of our running databases as they're due to expire in August 2024.

Don't think it's a blocker for this PR, but we should do this work soon.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.RegionCertificateAuthorities suggests all the newer ones support automatic rotation.

packages/cdk/lib/service-catalogue.ts
```diff
@@ -158,6 +159,19 @@ export class ServiceCatalogue extends GuStack {
 			dataType: ParameterDataType.TEXT,
 		});
 
+		const riffRaffDatabaseAccessSecurityGroupParam =
+			StringParameter.valueForStringParameter(
+				this,
+				`/${stage}/deploy/riff-raff/external-database-access-security-group`,
+			);
```
akash1810, line 166: Worth adding an inline comment to explain which stack provisioned this parameter?

packages/cdk/lib/service-catalogue.ts
```diff
@@ -512,6 +526,45 @@ export class ServiceCatalogue extends GuStack {
...
+
+		const riffRaffDatabaseAddress = StringParameter.valueForStringParameter(
+			this,
+			`/${stage}/deploy/riff-raff/external-database-access-address`,
+		);
+
+		const riffRaffDatabasePort = StringParameter.valueForStringParameter(
+			this,
+			`/${stage}/deploy/riff-raff/external-database-access-port`,
+		);
```
akash1810, line 545: Any reason to split this across three resources vs. a JSON blob in Secrets Manager?

akash1810 (commented):
This is exciting!

cc @guardian/devx-reliability who might be interested in adding this information to their SLO dashboards.

packages/cdk/lib/ecs/config.ts
AshCorr: Yep, I was considering rotating them as part of this PR but it was already getting big enough and I'm not sure how the riff-raff app will react to using a new CA so decided to leave it out for the time being!

packages/cdk/lib/service-catalogue.ts
akash1810: Ah, is the SSM Params are being created in Riff-Raff's DB stack, and the Secrets Manager resource created by hand?

AshCorr: Yep thats right, the Secrets Manager resource does get provisioned by service catalogue but all its credentials have to be manually editted!

AshCorr: See https://github.com/guardian/deploy-tools-platform/pull/731#discussion_r1378761791

PR #425, opened 2023-11-02
feat(messaging): Send messages to a thread, rather than individually 

akash1810:
## What does this change?
This change adopts the recently added message threading in Anghammarad (see https://github.com/guardian/anghammarad/pull/138) to group messages together.

Messages from RepoCop and Branch Protector will appear on the _same_ thread, for all time.

❓ Should there be a different thread per day?

## Why?
The intention here is to reduce noise in Chat. Grouping messages together also helps to keep a running narrative of events.

## How has it been verified?
I've run RepoCop and observed the messages in Chat appearing on a single thread:

image

### DIFF
--- a/packages/repocop/package.json
+++ b/packages/repocop/package.json
@@ -23,6 +23,6 @@
 		"@aws-sdk/rds-signer": "^3.429.0",
 		"@aws-sdk/credential-providers": "^3.429.0",
 		"@prisma/client": "^5.5.0",
-		"@guardian/anghammarad": "^1.8.0"
+		"@guardian/anghammarad": "^1.8.1"
 	}
 }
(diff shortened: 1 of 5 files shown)


packages/repocop/package.json
```diff
@@ -22,6 +22,7 @@
 		"@aws-sdk/client-sqs": "^3.438.0",
 		"@aws-sdk/rds-signer": "^3.429.0",
 		"@aws-sdk/credential-providers": "^3.429.0",
-		"@prisma/client": "^5.5.0"
+		"@prisma/client": "^5.5.0",
+		"@guardian/anghammarad": "^1.8.1"
```
akash1810, line 26: RepoCop is using this package, so it should have it declared as a dependency. Things were working previously as `@guardian/anghammarad` is being installed in branch-protector.

tjsilver (approved):
Great work! So, once branch protector runs on those repos, the messages confirming that it's been applied will appear in the same thread as the repocop warnings?

akash1810:
> So, once branch protector runs on those repos, the messages confirming that it's been applied will appear in the same thread as the repocop warnings?

Correct.

PR #426, opened 2023-11-02
refactor: Tweak the message metadata

akash1810:
## What does this change?
We have two systems sending messages to teams via Anghammarad.

This change tweaks the sent message to allow one to better identify the origin of the message:
- The message's `sourceSystem` is unique, and is comprised of the app name and stage
- The message's `subject` includes the GitHub team slug

## Why?
We currently message every administering GitHub team of a repository. These GitHub teams may belong to the same P&E team, and therefore land in a single Chat thread.

For example:
image

These changes help to understand why the same message is appearing multiple times as the subjects will read:
- Repocop branch protection (for GitHub team A)
- Repocop branch protection (for GitHub team B)
- Repocop branch protection (for GitHub team C)

Rather than simply "Repocop branch protection" x3.

## How has it been verified?
N/A.

### DIFF
--- a/packages/branch-protector/package.json
+++ b/packages/branch-protector/package.json
@@ -6,7 +6,7 @@
 	"scripts": {
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node18 --outdir=dist --external:@aws-sdk",
 		"test": "echo \"No test specified\" && exit 0",
-		"start": "ts-node src/run-locally.ts"
+		"start": "APP=branch-protector ts-node src/run-locally.ts"
 	},
 	"author": "guardian",
 	"dependencies": {
--- a/packages/branch-protector/src/aws-requests.ts
+++ b/packages/branch-protector/src/aws-requests.ts
@@ -47,11 +47,13 @@ export async function deleteFromQueue(
 
 export async function notify(
 	fullRepoName: string,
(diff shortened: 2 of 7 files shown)


packages/branch-protector/package.json
```diff
@@ -6,7 +6,7 @@
 	"scripts": {
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node18 --outdir=dist --external:@aws-sdk",
 		"test": "echo \"No test specified\" && exit 0",
-		"start": "ts-node src/run-locally.ts"
+		"start": "APP=branch-protector ts-node src/run-locally.ts"
```
akash1810, line 9: Set the value here, rather than the root `.env` file so that it's unique across branch-protector and RepoCop.

JuliaBrigitte (commented):
How has it been verified?
N/A.
This baffles me a bit.

akash1810:
> How has it been verified? N/A. This baffles me a bit.

Ah, how come? I use N/A here as the changes aren't significant enough to need a CODE deployment. I have confidence the change will work as it is small, and only to copy - there is no change to logic, etc. Happy to expand this section if you think it's needed though.

tjsilver (approved):
This will be useful, especially for our team(s)!

PR #428, opened 2023-11-02
Add dedicated CloudQuery source for DynamoDB tables

jacobwinch:
## What does this change?

This PR allows us to collect DynamoDB data once per day. 

## Why?

Currently this data is [collected once per week](https://github.com/guardian/service-catalogue/blob/6ead55e82f342f37011195dba1e95400a9f13456/packages/cdk/lib/service-catalogue.ts#L302-L308), but DevX Reliability are tracking one of their KRs based on this data, so it'd be nice to have more regular updates.

## How has it been verified?

As we are copying an existing pattern, I think the snapshot testing is sufficient here, but please let me know if you'd normally test anything else before merging a PR like this.

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -297,6 +297,17 @@ export class ServiceCatalogue extends GuStack {
 				managedPolicies: [readonlyPolicy],
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
+			{
+				name: 'OrgWideDynamoDB',
+				description:
+					'Collecting DynamoDB data across the organisation. Uses include auditing backup configuration.',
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/service-catalogue.ts
```diff
@@ -297,6 +297,17 @@ export class ServiceCatalogue extends GuStack {
 				managedPolicies: [readonlyPolicy],
 				policies: [listOrgsPolicy, standardDenyPolicy, cloudqueryAccess('*')],
 			},
+			{
+				name: 'OrgWideDynamoDB',
+				description:
+					'Collecting DynamoDB data across the organisation. Uses include auditing backup configuration.',
+				schedule: nonProdSchedule ?? Schedule.cron({ minute: '0', hour: '5' }),
```
jacobwinch, line 304: It looks like we are deliberately spacing these tasks out, so I have scheduled this one for an hour later than the previous source (S3).

PR #436, opened 2023-11-07
Group threads by repository

NovemberTang:
Using a slash in the thread key will confuse google, so replace with dash

## What does this change?

When threading messages from repocop/branch protector, make each repository a separate thread

## Why?

Having dozens of messages in a thread makes it difficult at first glance to figure out which ones are today's, and which ones are important. It also takes up a lot of visual space on google chat. Grouping them by repository means that we are still batching messages in the case that multiple teams resolve to the same chat, but keeps the thread length manageable.

## How has it been verified?

TODO: verify on code.


AshCorr (approved):
Looks good!

akash1810:
> figure out which ones are today's

Wonder if threading by date might be better?

PR #437, opened 2023-11-07
fix: Update RDS certificate authority

akash1810:
The current certificate authority (CA) (rds-ca-2019) is due to expire in August 2024. Update to the newer rds-ca-rsa2048-g1. This CA supports automatic rotation too! See https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.RegionCertificateAuthorities.

According to the [CloudFormation spec](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-cacertificateidentifier), this change should not replace the current database, however it will require a database reboot. This might cause running CloudQuery tasks to error, however we can simply rerun them.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -197,7 +197,7 @@ exports[`The ServiceCatalogue stack matches the snapshot 1`] = `
             "Command": [
               "/bin/sh",
               "-c",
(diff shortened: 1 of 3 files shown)


akash1810:
This took less than 5 minutes to apply on CODE.

tjsilver (approved):
Will we be setting up auto rotation too?

akash1810:
> Will we be setting up auto rotation too?

AFAICT this is directly managed by AWS.

PR #439, opened 2023-11-07
Change scheduling

NovemberTang:
## What does this change?

1. Changes schedule of remaining AWS tasks from 10am on Monday to 4pm on Saturday
2. Lower timeout of repocop

## Why?

1. The job running during the week means a lot of DB connections are used up when things like repocop are running, resulting in timeouts. Moving it to the weekend makes this issue less prelevant.
2. Repocop typically takes between 10 and 70 seconds to complete. A hanging DB request here means that the lambda will wait for 15 minutes before failing, accumulating cost, and wasting resources. Lowering the timeout to 5 minutes reduces that waste, while also giving us plenty of headroom.


AshCorr (approved):

akash1810:
I wonder if enabling [multi AZ](https://aws.amazon.com/rds/features/multi-az/) would be a more reliable fix for this issue?

PR #440, opened 2023-11-07
Skip `riffraff_deploy_logs`.

AshCorr:
## What does this change?

Skip `riffraff_deploy_logs` table

## Why?

riff-raff does not appear to have any retention policy on deployment logs so this table is HUUUUUGE (36M rows last checked) and takes hours to complete.

If we do want this info we should edit our riff-raff views to include a limit to restrict our view recent logs only (2 weeks?)


akash1810:
> 36M rows last checked

img

PR #441, opened 2023-11-07
refactor: give branch protector queue url env var unique name

tjsilver:
## What does this change?

Changes the name of the `QUEUE_URL` env var that is used by `branch-protector` to be specific to that app.

## Why?

This makes it clear that this is a branch protector and allows for another queue url to be added for other apps in future.

## How has it been verified?
Tested locally by running ./scripts/setup and then cloudquery and then repocop.

### DIFF
--- a/packages/branch-protector/src/config.ts
+++ b/packages/branch-protector/src/config.ts
@@ -72,7 +72,7 @@ export async function getConfig() {
 			installationId: secretJson.installationId,
 		},
 		anghammaradSnsTopic: getEnvOrThrow('ANGHAMMARAD_SNS_ARN'),
-		queueUrl: getEnvOrThrow('BRANCH_PROT_QUEUE_URL'),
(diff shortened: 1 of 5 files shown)


packages/branch-protector/src/config.ts
```diff
@@ -72,7 +72,7 @@ export async function getConfig() {
 			installationId: secretJson.installationId,
 		},
 		anghammaradSnsTopic: getEnvOrThrow('ANGHAMMARAD_SNS_ARN'),
-		queueUrl: getEnvOrThrow('QUEUE_URL'),
+		queueUrl: getEnvOrThrow('BRANCH_PROT_QUEUE_URL'),
```
JuliaBrigitte, line 75: BRANCH_PROT_QUEUE_URL what does PROT mean?

tjsilver: It's short for branch protector (for brevity) - I can make it the full word if that's useful?

NovemberTang (approved):

PR #443, opened 2023-11-07
collect dependabot alerts

NovemberTang:
## Why?

In order for PNPM to become a recommended package manager across the department, we need to be able to track its vulnerabilities. Let's start gathering this information

## How has it been verified?

TODO: deploy to code

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -3389,7 +3389,6 @@ spec:
   skip_tables:
     - github_releases
     - github_release_assets
-    - github_repository_dependabot_alerts
     - github_repository_dependabot_secrets
   destinations:
     - postgresql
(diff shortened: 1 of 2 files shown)


akash1810:
FYI there are quite a few - https://github.com/orgs/guardian/security/alerts/dependabot. Hopefully the API quota is sufficient!

NovemberTang:
> FYI there are quite a few - https://github.com/orgs/guardian/security/alerts/dependabot. Hopefully the API quota is sufficient!

We are definitely going to get stuck, but hopefully no more than once as [our limit should be 12500](https://docs.github.com/en/rest/overview/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github-app-installations)

NovemberTang:
This PR has been closed as collecting the dependabot alerts table hits so many rate limits, I gave up and cancelled the job after two days. If we need to alert on this information, we should collect it ad-hoc from the Github API, from repositories we care about, reducing the volume of requests.

PR #444, opened 2023-11-07
fix: throw error if a repo doesn't have any branches

tjsilver:
## What does this change?

Throws an error in `branch-protector` if the a repo doesn't have any branches. Also adds some console errors if branch protection fails, and slightly refactors the protect branch function to by DRYer.

## Why?

We recently had problems when `branch-protector` came across a repo that was empty and thus didn't have any branches to protect  (`guardian/braze-diff-publisher`).

## How has it been verified?
Tested on CODE with `braze-diff-publisher`. There were 5 other messages on the queue, and `branch-protector` ran successfully, didn't apply branch protections to the other repos (as it's CODE, expected), threw an error for `braze-diff-publisher`, and deleted all the messages off the queue including the failing one.

### DIFF
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -69,8 +69,8 @@ async function handleMessage(
 		try {
 			await protectBranch(octokit, config, event);
 		} catch (error) {
-			console.error(`Error: branch protection failed for  ${event.fullName}`);
-			console.error(error);
+			console.warn(`Branch protection failed for ${event.fullName}`);
+			console.warn(error);
 		}
 	}
 	await deleteFromQueue(config, message, sqs);
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -23,7 +23,13 @@ async function protectBranch(
 		throw new Error(`Invalid repo name: ${event.fullName}`);
 	}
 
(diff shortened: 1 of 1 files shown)


packages/branch-protector/src/index.ts
```diff
@@ -23,7 +23,13 @@ async function protectBranch(
 		throw new Error(`Invalid repo name: ${event.fullName}`);
 	}
 
-	const defaultBranchName = await getDefaultBranchName(owner, repo, octokit);
+	let defaultBranchName = undefined;
```
NovemberTang, line 26: throwing an error would mean that the lambda fail before deleting the item off the queue, and would re-read it next time, meaning the lambda would basically fail forever until we came along and manually deleted the message. Could we handle this a different way? Maybe just skip if we find the repo to be empty or have no branches?

NovemberTang (changes requested):

packages/branch-protector/src/index.ts
tjsilver: I did test this on CODE and it worked - see description. Happy to discuss. I agree that we could be fixing earlier in repocop.

NovemberTang (approved):

AshCorr:
Should this not be a warning instead of an error? Just thinking if you ever wanted to setup alerting if these lambdas failed for any reason you probably wouldn't want it to get a false positive because a repo has no branches!

PR #448, opened 2023-11-09
create common functions for duplicated repocop logic

NovemberTang:
## What does this change?

Creates and tests functions for duplicated anghammarad logic

## Why?

Adherence to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle, and more reliable code with better test coverage

## How has it been verified?

Tests pass

### DIFF
--- a/packages/common/src/functions.test.ts
+++ b/packages/common/src/functions.test.ts
@@ -1,0 +1,39 @@
+import { anghammaradThreadKey, branchProtectionCtas } from './functions';
+
+function isValidUrl(str: string) {
+	try {
+		new URL(str);
+		return true;
+	} catch (err) {
+		return false;
+	}
+}
+
+describe('branchProtectionCtas', () => {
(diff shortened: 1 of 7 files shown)


tjsilver (approved):
neat!

packages/common/src/functions.test.ts
```diff
@@ -0,0 +1,39 @@
+import { anghammaradThreadKey, branchProtectionCtas } from './functions';
+
+function isValidUrl(str: string) {
+	try {
+		new URL(str);
+		return true;
+	} catch (err) {
+		return false;
+	}
+}
```
akash1810, line 10: We should update the types used in anghammarad-client, [`url`](https://github.com/guardian/anghammarad/blob/08eed43102a35b2a71e7b2d426e683f40f24960d/anghammarad-client-node/src/interfaces.ts#L5) should be of type `URL`.

PR #450, opened 2023-11-09
feat: Enable backups in PROD

akash1810:
## What does this change?
Uses a [recent change in GuCDK](https://github.com/guardian/cdk/releases/tag/v52.2.0) to enable backups of the RDS database in PROD (the test is updated so we can see the snapshot changes).

## Why?
Having database backups is good practice!

## How has it been verified?
Not sure how TBH, as true success is if the backup is made. @guardian/devx-reliability any advice?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3698,15 +3698,15 @@
       "integrity": "sha512-aKEgbdG4+LoQXcMbxf2AH6epRPefLl/k5JTHLNuUkgtyQTZB02LwQviGhOxDU5W4oPo0EHScBVw+cf7/PyLV7w=="
     },
     "node_modules/@guardian/cdk": {
-      "version": "52.1.0",
-      "resolved": "https://registry.npmjs.org/@guardian/cdk/-/cdk-52.1.0.tgz",
-      "integrity": "sha512-2Lzd+13n7F4/jmtGmklEeDVcl2A/UcdRvcIZ/OtopqJcBI6+igHOrn2OTBIGo72kSRFZSqhbJ1RomW1ZQzP1Sw==",
+      "version": "52.2.0",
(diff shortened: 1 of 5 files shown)


jacobwinch:
I found the RDS/AWS Backup integration confusing, so just wanted to clarify:

Currently this RDS instance has a feature called automated backups enabled (at the RDS instance level)[^1]:

image

This allows you to restore to any point in time within the backup retention period:

image

In order to support this functionality, AWS takes an automated snapshot of your RDS instance during the specified backup window once per day:

image

By opting-in to this new feature via https://github.com/guardian/aws-backup, a nightly snapshot will be taken by AWS Backup (at ~00:00 UTC). It will be stored in a 'locked' vault, which means that it is better protected (read: harder to delete) than the backups that you currently have configured at the RDS instance level.

In the future, we might look at start managing the automated backups feature (currently configured on individual RDS instances) via AWS Backup, but for now this PR won't modify any of those settings.

[^1]: This RDS feature gets [enabled by default](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-backupretentionperiod)

jacobwinch:
> ## How has it been verified?
> Not sure how TBH, as true success is if the backup is made. @guardian/devx-reliability any advice?

Yes, I think we'll just need to check that the new type of backup has been created tomorrow morning.

NovemberTang (approved):
🎉

akash1810:
Closing this, as (extra) backups don't make too much sense for this application (yet):
- Our configuration of CloudQuery is such that entire database tables are rewritten each run
- Typically, we run CloudQuery daily. That is, any backup is going to be stale.
- In the event of needing to restore service, we can re-run all the CloudQuery tasks

If I'm honest, I mainly raised this PR to test the GuCDK update. Will update a different app for this purpose instead.

PR #452, opened 2023-11-10
docs: Add ADR for checking our data accuracy

akash1810:
## What does this change?
Adds an ADR for how we can verify our data accuracy. The proposal is to directly use the SDKs, rather than using AWS Resource Explorer, or AWS Config.

## Why?
See https://github.com/guardian/recommendations/blob/main/architecture-decision-records.md.

## How has it been verified?
I've looked at [implementing Resource Explorer](https://github.com/guardian/aws-account-setup/pull/133). That is, the opinion to not use it for this purpose is well informed.

### DIFF
--- a/ADR/03-data-accuracy.md
+++ b/ADR/03-data-accuracy.md
@@ -19,6 +19,9 @@ We plan to do this by asking AWS to count the number of resources, and compare t
 
 This ADR outlines options for implementing the second item.
 
+It is also worth noting that this is only part of the data accuracy checks we'll perform.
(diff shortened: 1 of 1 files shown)


AshCorr (approved):
Is it worth also mentioning that we want to look at data age & error count during sync as an additional way of confirming that our data is accurate?

akash1810:
> Is it worth also mentioning that we want to look at data age & error count during sync as an additional way of confirming that our data is accurate?

Good shout. Done in https://github.com/guardian/service-catalogue/pull/452/commits/bed69a49d2a6ff1f581f814090c8322a21ed6790.

PR #456, opened 2023-11-10
Add aws_ec2_images to 5 minutely ec2 cloudquery task

louishather:
## What does this change?

This change adds the aws_ec2_images table to the task which syncs ec2 data every 5 minutes.

## Why?

AWS EC2 AMI data was fairly out of date, and caused some dashboards/queries to report misleading information.


akash1810 (approved):
Unrelated gif:

img

PR #457, opened 2023-11-10
bump versions

NovemberTang:
## What does this change?

- bumps versions
- deletes most snapshot tests

## Why?

- they need bumping
- inline snapshots are not automatically updated when using this version of jest. We've removed some of them for now as two is enough to verify that the functionality works, while being less annoying to developers

## How has it been verified?
TODO: test on code

### DIFF
--- a/.env
+++ b/.env
@@ -3,10 +3,10 @@
 # Environment variables shared between ci and DEV.
 
 # See https://github.com/cloudquery/cloudquery/releases?q=cli
-CQ_CLI=3.14.4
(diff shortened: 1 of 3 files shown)


akash1810 (approved):
img

akash1810:
https://jestjs.io/docs/configuration/#prettierpath-string looks like we can add `prettierPath: null` to the [Jest config](https://github.com/guardian/service-catalogue/blob/main/jest.config.js) to get around the inline snapshot woes.

PR #460, opened 2023-11-10
CI concurrency

NovemberTang:
## What does this change?

Not all CI steps are dependent on the previous one. Run them concurrently to save time/increase resource efficiency

## Why?

Speed things up a bit

## How has it been verified?

A successful CI run on GH now takes about two minutes rather than four

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -4,13 +4,6 @@ set -e
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 ROOT_DIR=$(realpath "${DIR}/..")
 
-npm ci
-npm run typecheck
-npm run lint
-npm run test
-npm run synth
-npm run build
-
 createRepocopZip() {
   echo "Creating repocop package"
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
Not a blocker, but I wonder if crafting the GitHub workflow with parallel jobs might be more performant? https://medium.com/tradeling/how-to-achieve-parallel-execution-using-github-actions-d534404702fb looks helpful here.

NovemberTang:
> Not a blocker, but I wonder if crafting the GitHub workflow with parallel jobs might be more performant? https://medium.com/tradeling/how-to-achieve-parallel-execution-using-github-actions-d534404702fb looks helpful here.

potentially! I didn't have a lot of time to look into this so slapped together something quickly, but this is definitely something to investigate in the future

AshCorr:
> Not a blocker, but I wonder if crafting the GitHub workflow with parallel jobs might be more performant?

Would also make the actions logs easier to interpret, I think right now all the logs from parallel tasks are being jumbled together!

PR #464, opened 2023-11-13
Verify a repository has been generated from an interactive template

NovemberTang:
## What does this change?

- Extract github credential logic into common project
- Add logic that checks whether a repo has been generated from an interactive template
- Repocop sends a list of potential interactives to the interactive monitor

## Why?

- To avoid duplicating logic
- A repo being generated from one of these templates is one of two reliable signals that something is an interactive
- So it knows where to look

## How has it been verified?

Deployed to CODE. We are able to consume a message from the queue, and check the repo in the message was generated from one of the interactive templates.


packages/common/src/functions.ts
```diff
@@ -1,4 +1,63 @@
...
+	return githubAppConfig;
+}
+
+export async function getGitHubAppConfig(
+	envVar = 'GITHUB_APP_SECRET',
+): Promise<GitHubAppConfig> {
+	const secretString = await getGithubAppSecret(envVar);
+	const githubAppConfig = parseSecretJson(secretString);
+	return githubAppConfig;
+}
```
akash1810, line 60: I think we can simplify here by removing the `envVar` argument, as it always appears to be the default of `GITHUB_APP_SECRET`.

packages/interactive-monitor/package.json
```diff
@@ -7,5 +7,9 @@
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node18 --outdir=dist --external:@aws-sdk",
 		"test": "echo \"No test specified\" && exit 0"
 	},
-	"author": "guardian"
+	"author": "guardian",
+	"dependencies": {
+		"@octokit/types": "^12.0.0",
```
akash1810, line 12: Should the types be `devDependencies`?

packages/interactive-monitor/package.json
```diff
@@ -7,5 +7,9 @@
...
+		"octokit": "^3.1.1"
```
akash1810, line 13: I think these dependencies should be moved to the `common` package, as there's a function there that imports from them?

akash1810: Or even push them to the root `package.json`; `esbuild` will tree-shake them away anyway.

packages/cdk/lib/interactive-monitor.ts
```diff
@@ -1,32 +1,38 @@
 import { type GuStack } from '@guardian/cdk/lib/constructs/core';
 import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
 import { Runtime } from 'aws-cdk-lib/aws-lambda';
+import { Secret } from 'aws-cdk-lib/aws-secretsmanager';
 import { Topic } from 'aws-cdk-lib/aws-sns';
 import { LambdaSubscription } from 'aws-cdk-lib/aws-sns-subscriptions';
 
-const app = 'interactive-monitor';
-
+const service = 'interactive-monitor';
```
akash1810, line 8: Any reason to not make this the `app`?

side note: I wonder if we've been using the correct values thus far? Wonder if it should be `Stack=service-catalogue`, `Stage=PROD`, `App=cloud-query | repocop | interactive-monitor`.

packages/interactive-monitor/README.md
```diff
@@ -1 +1,10 @@
 ## Interactive Monitor
+
+The interactive Monitor attempts to spot and label new/unknown interactives. It is currently under active development.
+
+There are a few ways of spotting an interactive
+
+1. It has been generated from an interactive template, typically these repos will start with `interactive-atom-template`. Currently this part of the GitHub API is not exposed via cloudquery, so we are using the GitHub API directly until we can commit an upstream change to cloudquery itself.
+2. Objects/artifacts generated by this repository have been spotted in S3. These repos have a `config.json`. The path field in that file, will correspond to prefixes in an S3 bucket. This indicates that an interactive has gone live.
+
+These checks can take a while to run, and may brush up against rate limits more quickly, so for now, they live in their own lambda. When we are able to use cloudquery to search for templates, and once the numbers have come down a bit, we can move this into repocop proper.
```
akash1810, line 10: Do we use dedicated GitHub credentials for this lambda? Or are we reusing the ones used by CloudQuery?

packages/interactive-monitor/src/index.ts
```diff
@@ -1,3 +1,32 @@
...
+	const repoData = await octokit.rest.repos.get({
+		owner: 'guardian',
+		repo,
+	});
+
+	return (
+		repoData.data.template_repository?.name.includes(
+			'interactive-atom-template',
+		) != null
+	);
```
akash1810, line 19: IIUC [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) returns a boolean. Is the `!= null` check doing what we intend to here?

packages/interactive-monitor/src/index.ts
```diff
@@ -1,3 +1,32 @@
...
 }
+
+export const handler: SNSHandler = async (event) => {
+	const message = event.Records[0]!.Sns.Message;
```
akash1810, line 23: Could we add a length check here?

```ts
const records = event.Records;

if(records.length === 0) {
  // exit early
} else {
  // the logic
}
```

packages/repocop/src/index.ts
```diff
@@ -24,6 +32,39 @@ async function writeEvaluationTable(
...
 }
 
+async function sendPotentialInteractives(
+	evaluatedRepos: repocop_github_repository_rules[],
+	config: Config,
+) {
+	const snsBatchMaximum = 10;
+	const potentialInteractives = shuffle(
+		findPotentialInteractives(evaluatedRepos),
+	).slice(0, snsBatchMaximum);
```
akash1810, line 42: I think passing this list into the function would make it easier to test, and make the function more "single responsibility".

packages/repocop/src/index.ts
```diff
@@ -24,6 +32,39 @@ async function writeEvaluationTable(
 	console.log('Finished writing to table');
 }
 
+async function sendPotentialInteractives(
+	evaluatedRepos: repocop_github_repository_rules[],
+	config: Config,
+) {
+	const snsBatchMaximum = 10;
+	const potentialInteractives = shuffle(
```
akash1810, line 40: Why `shuffle` here?

akash1810 (commented):
LGTM. Couple of small queries, nothing blocking.

akash1810 (approved):
LGTM. Couple of small queries, nothing blocking.

PR #466, opened 2023-11-13
Refactor: Move shareable functions into shared file

tjsilver:
## What does this change?

- Moves various functions that are currently used by `branch-protector` into a shared file.
- Adds some types and objects to allow the right app and messaging to be passed into the shared messaging functions.
- Adds some scaffolding code so that the new topic monitor for 'production' tags is set up to message correctly.

## Why?

- The shared utitilies can be used by the forthcoming `topic-monitoring-production` lambda that will automate the application of the `production` topic to repos, as it will follow a very similar messaging pattern to `branch-protector`.

## How has it been verified?
Tested by running repocop locally using CODE database. Repocop ran successfully and added branch protection repos to the queue as before.

### DIFF
--- a/packages/common/src/functions.ts
+++ b/packages/common/src/functions.ts
@@ -78,6 +78,27 @@ export function branchProtectionCtas(
 	];
 }
 
+export function topicProductionCtas(
+	fullRepoName: string,
+	teamSlug: string,
+): Action[] {
+	const githubUrl = `https://github.com/${fullRepoName}`;
+	const grafanaUrl = `https://metrics.gutools.co.uk/d/EOPnljWIz/repocop-compliance?var-team=${teamSlug}&var-rule=All&orgId=1`;
+	const topicsUrl = `https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/classifying-your-repository-with-topics#adding-topics-to-your-repository`;
+
+	return [
+		{ cta: 'Repository', url: githubUrl },
+		{
+			cta: 'Compliance information for repos',
+			url: grafanaUrl,
+		},
+		{
+			cta: 'How to add a topic',
+			url: topicsUrl,
+		},
+	];
+}
+
 export function anghammaradThreadKey(fullRepoName: string) {
 	return `service-catalogue-${fullRepoName.replaceAll('/', '-')}`;
 }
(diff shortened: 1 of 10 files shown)


packages/repocop/src/remediations/shared-utilities.ts
```diff
@@ -0,0 +1,160 @@
...
+import type { github_teams, view_repo_ownership } from '@prisma/client';
+import {
+	anghammaradThreadKey,
+	branchProtectionCtas,
+	topicProductionCtas,
+} from 'common/src/functions';
+import type { UpdateBranchProtectionEvent } from 'common/types';
+import type { Config } from '../config';
+
+function findTeamSlugFromId(
```
NovemberTang, line 15: while we're moving things around, this might be a good chance to test at least the non-async functions in this file

packages/repocop/src/remediations/shared-utilities.ts
```diff
@@ -0,0 +1,160 @@
...
+): string[] {
+	const owners = allRepoOwners.filter((owner) => owner.full_name === repo);
+	const teamSlugs = owners
+		.map((owner) => findTeamSlugFromId(owner.github_team_id, teams))
+		.filter((slug): slug is string => !!slug);
+
+	return teamSlugs;
+}
+
+export function createEntry(
```
NovemberTang, line 64: I can only tell what this method does in context. Could we rename it something like `createSqsEntry` or similar?

packages/common/src/functions.ts
```diff
@@ -21,6 +21,27 @@ export function branchProtectionCtas(
 	];
 }
 
+export function topicProductionCtas(
```
NovemberTang, line 24: as far as i can tell we are only using this method once in repocop. Does it need to live in `common`?

tjsilver: I haven't started using it yet! But I expect it to be used in the same way as the branch protector CTAs, once implemented.

packages/repocop/src/remediations/shared-utilities.ts
tjsilver: Good idea!

packages/common/src/functions.ts
tjsilver: I'll remove it from this PR though to avoid confusion.

packages/repocop/src/remediations/shared-utilities.ts
```diff
@@ -0,0 +1,161 @@
...
+	anghammaradThreadKey,
+	branchProtectionCtas,
+	topicProductionCtas,
+} from 'common/src/functions';
+import type { UpdateMessageEvent } from 'common/types';
+import type { Config } from '../config';
+
+export enum RemediationApp {
+	BranchProtector = 'branchProtector',
+	TopicProduction = 'topicProduction',
```
NovemberTang, line 17: this name is a bit unclear.

NovemberTang (approved):

PR #474, opened 2023-11-14
Include interactives

NovemberTang:
## What does this change?

Now that interactive repos are required to have a label, we can more easily filter them out, and don't need name-based matching. Not pre-filtering also helps us flag interactives that may need actioning, and will allow us to catch P+E repos that may have those prefixes

One of the rules has also been modified to make it clear what kinds of repos we do and don't care about

## Why?

Name based filtering allowed us to keep things manageable, but is a blunt tool. We are building more accurate ways of detecting interactives, and so need these repos to be part of evaluation again.


akash1810 (approved):
LGTM.

Worth double checking there is an Anghammarad mapping for the owning team of these repos?

scripts/ci.sh
```diff
@@ -34,5 +34,4 @@ npm ci
 npm run typecheck & npm run lint
 npm run test
 npm run synth & npm run build
-
-createRepocopZip & createZip "branch-protector" & createZip "interactive-monitor"
+createZip "branch-protector" & createZip "interactive-monitor"  & createRepocopZip
```
akash1810, line 37: Been struggling to understand this. Why does it fail, and why is this the fix?

NovemberTang: To be honest, I don't actually know the answer - I'm unfamiliar with the internal details of `&`. I moved these around on a hunch.

It looks like repocop.zip never gets created, so I suspect that this line is terminating early. The createZip calls both terminate after 10s of milliseconds, before createRepocopZip has had time to put its boots on. It's possible that ordering is important here, and there's a bug/feature where the last thing finishing causes all the other processes to halt. But this is just conjecture based on the fact that rearranging the commands fixes the problem.

PR #475, opened 2023-11-14
Allow interactive repos to not have owners

NovemberTang:
## What does this change?

Allow interactives to be unowned

## Why?

We can assume interactives are owned by the interactives team


tjsilver (approved):
Makes sense!

PR #476, opened 2023-11-14
Prevent lambda from timing out

NovemberTang:
## What does this change?

Allows more DB connections
Bumps the DB timeout
Provide additional memory to the repocop lambda

## Why?

The lambda has gotten busier, and more queries are being made. We should bump the number of connections to avoid timeouts

## How has it been verified?

Tested on PROD, repocop succeeds. Billed about the same number of GB seconds as before

### DIFF
--- a/packages/repocop/src/query.ts
+++ b/packages/repocop/src/query.ts
@@ -72,13 +72,6 @@ export async function getRepositoryTeams(
 		},
 	);
 
-	// `full_name` is typed as nullable, in reality it is not, so the fallback to `id` shouldn't happen
-	const repoIdentifier = repository.full_name ?? repository.id;
-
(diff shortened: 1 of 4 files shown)


packages/repocop/src/query.ts
```diff
@@ -72,13 +72,6 @@ export async function getRepositoryTeams(
 		},
 	);
 
-	// `full_name` is typed as nullable, in reality it is not, so the fallback to `id` shouldn't happen
-	const repoIdentifier = repository.full_name ?? repository.id;
```
NovemberTang, line 76: this was useful when we were testing out repo evaluation, but I don't think we need it any more

AshCorr (approved):

PR #477, opened 2023-11-14
create class for branch protector components

NovemberTang:
## What does this change?

Pulls out branch protector infrastructure into its own class

## Why?

Declutter the main stack file
More of these classes will be created over the coming days

## How has it been verified?

No difference in snapshots


tjsilver (approved):
Good idea!

PR #479, opened 2023-11-15
feat: Add docker labels to each task

akash1810:
## What does this change?
This should allow a task to identify itself using the [ECS task metadata endpoint](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v3.html#task-metadata-endpoint-v3-response).

## Why?
This will be useful in implementing a singleton pattern for our tasks.

## How has it been verified?
The updated snapshot should suffice.


tjsilver (approved):
Looks good!

PR #481, opened 2023-11-16
make repocop infrastructure its own class

NovemberTang:
## What does this change?

Similar to #477 , this PR puts all repocop infrastructure into its own class

## Why?

Better readability. This makes the separation between different parts of the service catalogue, and their interfaces (mostly SNS topics/SQS queues) clearer

## How has it been verified?

Snapshots are identical


tjsilver (approved):
This should make things much easier to work with!

PR #482, opened 2023-11-16
Move cloudquery infra into classes

NovemberTang:
## What does this change?

Abstracts cloudquery sources, and cloudquery infra such as the DB into their on classes, to keep `service-catalogue.ts` readable

## Why?

- Readability
- Boundaries between apps are expressed implicitly via code

## How has it been verified?

The snapshot has not meaningfully changed. This is quite a lot of code changes, but rather than trudging through every line, the easiest way to review this is probably just to make sure you're happy with the general shape of the classes, and make sure you're happy with the snapshot diff

## Out of scope

cloudquery-source class abstraction.
There's a lot of very similar looking classes in `cloudquery-sources/`. We could probably abstract this further, and create a cloudquery source class that holds a lot of their common properties. But this would further complicate an already complicated PR, and can probably wait until the next one.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -6,11 +6,11 @@ exports[`The ServiceCatalogue stack matches the snapshot 1`] = `
     "gu:cdk:constructs": [
       "GuSubnetListParameter",
       "GuVpcParameter",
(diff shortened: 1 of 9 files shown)


akash1810 (commented):
I'd be tempted to simplify this and not split the CloudQuery sources across multiple files. I think it'd also be easier to keep the database in `service-catalogue.ts` as it's a shared resource.

PR #483, opened 2023-11-16
make logs more useful, fix octokit request

NovemberTang:
## What does this change?

We are submitting the full repo name to github. We shouldn't be doing this
Also make repocop logging more useful

## Why?

Passing in the full name into the name field means we are searching for the wrong repo and return a 404. ie `guardian/guardian%2Fcapi` rather than `guardian/capi`

## How has it been verified?

Deployed to code, octokit is able to find the correct repo

### DIFF
--- a/packages/interactive-monitor/src/index.ts
+++ b/packages/interactive-monitor/src/index.ts
@@ -21,9 +21,12 @@ async function isFromInteractiveTemplate(
 
 export const handler: SNSHandler = async (event) => {
 	const config = getConfig();
-	if (event.Records.length !== 1) {
+	if (
+		event.Records.length !== 1 ||
+		!event.Records[0]!.Sns.Message.split('/')[1]
+	) {
 		throw new Error(
-			`Expected exactly one record, but got ${event.Records.length}`,
+			`Failed to process SNS message: ${JSON.stringify(event, null, 2)}`,
 		);
 	} else {
 		const repo = event.Records[0]!.Sns.Message.split('/')[1]!;
--- a/packages/interactive-monitor/src/index.ts
+++ b/packages/interactive-monitor/src/index.ts
@@ -26,7 +26,7 @@ export const handler: SNSHandler = async (event) => {
 			`Expected exactly one record, but got ${event.Records.length}`,
 		);
 	} else {
-		const repo = event.Records[0]!.Sns.Message;
+		const repo = event.Records[0]!.Sns.Message.split('/')[1]!;
 		const owner = 'guardian';
 		console.log('received repo', repo);
 		const githubAppConfig: GitHubAppConfig = await getGitHubAppConfig();
--- a/packages/interactive-monitor/src/index.ts
+++ b/packages/interactive-monitor/src/index.ts
@@ -2,15 +2,17 @@ import type { SNSHandler } from 'aws-lambda';
 import { getGitHubAppConfig, getGithubClient } from 'common/functions';
 import type { GitHubAppConfig } from 'common/types';
 import type { Octokit } from 'octokit';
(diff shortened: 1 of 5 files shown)


packages/interactive-monitor/src/index.ts
```diff
@@ -26,7 +26,7 @@ export const handler: SNSHandler = async (event) => {
 			`Expected exactly one record, but got ${event.Records.length}`,
 		);
 	} else {
-		const repo = event.Records[0]!.Sns.Message;
+		const repo = event.Records[0]!.Sns.Message.split('/')[1]!;
```
NovemberTang, line 29: the non null assertion is a bit controversial but i think we should trust that repocop has sent what it's promised to send. If not, we should throw an error anyway

packages/interactive-monitor/src/index.ts
```diff
@@ -21,12 +21,15 @@ async function isFromInteractiveTemplate(
 
 export const handler: SNSHandler = async (event) => {
 	const config = getConfig();
-	if (event.Records.length !== 1) {
+	if (
+		event.Records.length !== 1 ||
```
akash1810, line 25: Are we always guaranteed to have exactly one record? Might it be simpler to loop over them?

```ts
event.Records.forEach((record) => {
		const message = record.Sns.Message;
		const fullRepoName = message.split('/');
		if (fullRepoName.length === 2) {
			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we've checked the length above
			const repo = fullRepoName[1]!;
			
		}
	});
```

akash1810 (commented):
> Passing in the full name into the name field means we are searching for the wrong repo and return a 404. ie `guardian/guardian%2Fcapi` rather than `guardian/capi`

Should we not fix this at source instead, so that we always produce the right message?

packages/interactive-monitor/src/index.ts
NovemberTang: ok i think both of these are fixed now. It's unlikely we'll be handed multiple records but can't hurt to cover our bases. and it looks nicer.

packages/interactive-monitor/src/index.ts
```diff
@@ -19,30 +21,34 @@ async function isFromInteractiveTemplate(
...
-	} else {
-		const repo = event.Records[0]!.Sns.Message;
-		const owner = 'guardian';
-		console.log('received repo', repo);
-		const githubAppConfig: GitHubAppConfig = await getGitHubAppConfig();
-		const octokit: Octokit = await getGithubClient(githubAppConfig);
+async function applyTopics(repo: string, owner: string, octokit: Octokit) {
+	const topics = (await octokit.rest.repos.getAllTopics({ owner, repo })).data
+		.names;
+	console.log(`${repo}is from interactive template`);
```
akash1810, line 27: Missing a space?

```ts
	console.log(`${repo} is from interactive template`);
```

packages/repocop/src/remediations/repository-06-topic-monitor-interactive.ts
```diff
@@ -7,3 +15,51 @@ export function findPotentialInteractives(
 		.filter((repo) => !repo.topics)
 		.map((repo) => repo.full_name);
 }
+
+export function createBatchEntry(string: string): PublishBatchRequestEntry {
```
akash1810, line 19: minor: Could we give the argument a more intention revealing name? Maybe `fullRepoName`?

akash1810 (approved):

tjsilver (approved):
Looks good, and like we'll be able to move some of the functions into 'common' when the production-monitor comes online!

PR #484, opened 2023-11-17
feat(cloudquery): Make `OrgWideEc2` a singleton task

akash1810:
We have a few tasks that are starting to overlap on runs. This does not follow [CloudQuery’s advice](https://www.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel#no-overlapping-syncs).

It’s difficult to predict how long a task will take. That is stopping the overlap via clever scheduling is tricky. This change implements [singleton](https://en.wikipedia.org/wiki/Singleton_pattern) tasks with the following behaviour:
  1. Task starts
  2. Check if another task in the same ECS family is currently running
  3. If yes, stop the task
  4. If no, continue to the CloudQuery container

This manifests as a `runAsSingleton` boolean property, which is defaulted to `false`.

In this change, the `OrgWideEc2` task becomes a singleton.

## How has it been verified?
- Deploy to CODE.
- Run the `OrgWideEc2` task twice (or more):

   ```sh
   npm -w cli start run-task -- --stage CODE --name OrgWideEc2
   ```
- Observe all but one getting to the to CloudQuery container.

   This can be done in the AWS console: 

   image

   Or via [the logs](https://logs.gutools.co.uk/s/devx/goto/a8d0cb70-8539-11ee-b6f6-a11d207f34cc) (there should be a unique `ecs_task_arn` column).

---

Closes https://github.com/guardian/service-catalogue/pull/461.


packages/cdk/lib/ecs/policies.ts
```diff
@@ -34,3 +37,30 @@ export const readBucketPolicy = (...resources: string[]): PolicyStatement => {
...
 	});
 };
+
+export function singletonPolicies(cluster: Cluster) {
+	const { account, region } = GuStack.of(cluster);
+
+	return [
+		new PolicyStatement({
+			effect: Effect.ALLOW,
+			resources: ['*'],
```
akash1810, line 47: I couldn't work out how to scope this to the cluster. Attempts to do so resulted in the following error:

```log
An error occurred (AccessDeniedException) when calling the ListTasks operation: User: arn:aws:sts::REDACTED:assumed-role/deploy-CODE-service-catal-CloudquerySourceOrgWideEc-KzPt3L8oQO2Q/22c2eb98e5f746abae0932eca9ed0977 is not authorized to perform: ecs:ListTasks on resource: * because no identity-based policy allows the ecs:ListTasks action
```

Being able to list all tasks, isn't too bad from a security perspective.

packages/cdk/lib/ecs/policies.ts
```diff
@@ -34,3 +37,30 @@ export const readBucketPolicy = (...resources: string[]): PolicyStatement => {
...
+		new PolicyStatement({
+			effect: Effect.ALLOW,
+			resources: [
+				Arn.format({
+					partition: 'aws',
+					service: 'ecs',
+					region,
+					account,
+					resource: 'task',
+					resourceName: `${cluster.clusterName}/*`,
```
akash1810, line 59: `*` here as the ARN for the task can only be known at runtime

packages/cdk/lib/ecs/task.ts
```diff
@@ -191,6 +204,58 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
...
+		if (runAsSingleton) {
+			const singletonTask = task.addContainer(`${id}AwsCli`, {
+				image: awsImage,
+				entryPoint: [''],
+				command: [
+					'/bin/bash',
+					'-c',
+					[
+						// Install jq to handle JSON, and awscli to query ECS
+						'yum install -y -q jq awscli',
```
akash1810, line 216: We could use the [AWS CLI image](https://hub.docker.com/r/amazon/aws-cli) directly here, and just install `jq`. Not sure how that'll impact performance though, as the CLI image is significantly bigger than amazonlinux: ~420MB vs ~180MB. Suggest we experiment in a follow-up PR.

JuliaBrigitte (approved):
Akash walked me through the PR and I am now happy to approve

PR #489, opened 2023-11-20
feat(data-audit): Initialise data-audit lambda

akash1810:
## What does this change?
Add the infrastructure for the data-audit lambda. The lambda currently prints the time, and will be later updated to track the accuracy of our data.

In this initial step, we're testing:
- Can we build the lambda?
- Can we deploy the lambda?
- Can we see the lambda's logs in Central ELK?

## Why?
See https://github.com/guardian/service-catalogue/blob/main/ADR/03-data-accuracy.md.

## How has it been verified?
- The lambda built and [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/a3789524-fe0d-4256-9b85-d18f9073eb65) successfully
- The logs appear in [Central ELK](https://logs.gutools.co.uk/s/devx/goto/98a08e80-8795-11ee-a34c-11e7b53d0c58)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,5 +59,7 @@ jobs:
               - packages/branch-protector/dist/branch-protector.zip
             interactive-monitor:
               - packages/interactive-monitor/dist/interactive-monitor.zip
+            data-audit:
+              - packages/data-audit/dist/data-audit.zip
     env:
       NODE_OPTIONS: '--max_old_space_size=4096'
--- a/package-lock.json
+++ b/package-lock.json
@@ -7983,6 +7983,10 @@
       "integrity": "sha512-PjiQ659aQ+fUTQqSrd1XEDnOr52jh30RBurfzkscaE2tPaFsDH5wOAHJiw8XAHphRknCwMUE9KRayc4K/NbO8A==",
       "dev": true
     },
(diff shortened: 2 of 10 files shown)


NovemberTang (approved):

NovemberTang (commented):
should we refactor the other lambdas to use functions rather than classes? multiple approaches here might be confusing

akash1810:
> should we refactor the other lambdas to use functions rather than classes? multiple approaches here might be confusing

Already on my list. Did you have a preference? I find functions simpler, but conscious that https://github.com/guardian/service-catalogue/pull/482 uses classes.

akash1810:
> clear which resources are supposed to be interacting with other parts of the catalogue

An alternative to public members could be to defined shared resources in the `GuStack` and pass them through to the places they're needed. Defining shared resources at the top level would make it easier to understand what's shared, and also clarifies ownership; shared resources are owned by the entire stack, rather than an individual component.

akash1810:
> An alternative to public members could be to define shared resources in the `GuStack` and pass them through to the places they're needed.

https://github.com/guardian/service-catalogue/pull/491 does this, for example. WDYT?

akash1810:
Merging this PR. Let's continue the discussion on separate PR(s).

PR #490, opened 2023-11-20
refactor: Address linting error

akash1810:
## What does this change?
Restructure code to address a linting error. This change also removes a redundant `else` for brevity.


akash1810:
This is already addressed in https://github.com/guardian/service-catalogue/pull/483. Closing.

PR #491, opened 2023-11-20
refactor(cdk): Move CloudQuery infrastructure to separate file

akash1810:
## What does this change?
This change aims to make the top level `service-catalogue.ts` more readable by moving the CloudQuery infrastructure into a separate file.

This is an alternative take on https://github.com/guardian/service-catalogue/pull/482. https://github.com/guardian/service-catalogue/pull/482 splits things even further, separating the CloudQuery tasks. Propose we make those changes in a follow-up PR to keep a small diff.

## Why?
Our infrastructure is growing, and it's important to be able to reason about, and maintain it with ease.

## How has it been verified?
N/A. This is a no-op refactor, which can be demonstrated by the CDK snapshot being unaffected.


packages/cdk/lib/cloudquery/index.ts
```diff
@@ -0,0 +1,462 @@
...
+	dbAccess: GuSecurityGroup;
+	nonProdSchedule?: Schedule;
+}
+
+export function addCloudqueryEcsCluster(
+	scope: GuStack,
+	props: CloudqueryEcsClusterProps,
+) {
+	const { stage, stack, app = 'service-catalogue' } = scope;
+	const { vpc, db, dbAccess, nonProdSchedule } = props;
```
akash1810, line 44: The contents of this file is largely a cut and paste from `service-catalogue.ts`. The only changes are:
- Exposing a function
- A rename of `applicationToPostgresSecurityGroup` to `dbAccess`

packages/cdk/lib/cloudquery/index.ts
akash1810, line 1: The contents of this file is largely a cut and paste from `service-catalogue.ts`. The only changes are:
- Exposing a function
- A rename of `applicationToPostgresSecurityGroup` to `dbAccess`

tjsilver (approved):
👍

PR #493, opened 2023-11-20
Create local development story for Interactive Monitor

NovemberTang:
## What does this change?

Allows us to run interactive monitor locally, bypassing the need to submit events to SNS.

Adds a `stageAwareOctokit` function so that we fall back to an API token if we can't find the github app

## Why?

Shorter feedback loop for application code changes

## How has it been verified?

Running the lambda on a `DEV` environment now works.


packages/interactive-monitor/src/index.ts
```diff
@@ -26,23 +24,26 @@ async function applyTopics(repo: string, owner: string, octokit: Octokit) {
...
+	const isInteractive = await isFromInteractiveTemplate(repo, owner, octokit);
+	const onProd = stage === 'PROD';
+	if (isInteractive && onProd) {
 		await applyTopics(repo, owner, octokit);
 	} else {
-		console.log(`No action taken for ${repo}`);
+		const reason =
+			(!isInteractive ? ' Repo not from interactive template.' : '') +
+			(!onProd ? ' Not running on PROD.' : '');
+		console.log(`No action taken for ${repo}.` + reason);
```
akash1810, line 38: Similar [feedback](https://github.com/guardian/service-catalogue/pull/494#discussion_r1400106687) to #494 here.

akash1810 (approved):

PR #494, opened 2023-11-20
simplify logs

NovemberTang:
## What does this change?

Remove redundant log lines, simplify the language used, and combine some logs into one message

## Why?

Clarity & brevity

### DIFF
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -48,7 +48,7 @@ async function protectBranch(
 		console.log(`Notified teams ${event.teamNameSlugs.join(', ')}}`);
 	} else {
 		const reason =
-			(branchIsProtected ? ' Branch is already protected' : '') +
+			(branchIsProtected ? ' Branch is already protected.' : '') +
 			(!stageIsProd ? ' Not running on PROD.' : '');
 		console.log(`No action required. ${reason}`);
 	}
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -28,7 +28,7 @@ async function protectBranch(
 	try {
 		defaultBranchName = await getDefaultBranchName(owner, repo, octokit);
 	} catch (error) {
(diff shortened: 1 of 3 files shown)


packages/branch-protector/src/index.ts
```diff
@@ -37,23 +37,19 @@ async function protectBranch(
...
+			await notify(event.fullName, config, slug);
 		}
+		console.log(`Notified teams ${event.teamNameSlugs.join(', ')}}`);
 	} else {
-		// !stageIsProd
-		console.log(`Detected stage: ${config.stage}. No action taken`);
+		const reason =
+			(branchIsProtected ? ' Branch is already protected' : '') +
+			(!stageIsProd ? ' Not running on PROD.' : '');
+		console.log(`No action required. ${reason}`);
```
akash1810, line 52: Not sure this improves readability. What about early returns instead?

```ts
if (!stageIsProd) {
  console.log(`No action required. Not running in PROD. Detected stage as ${config.stage}.`);
  return;
}

if(branchIsProtected) {
  console.log('No action required. Branch is already protected.');
  return;
}

await updateBranchProtection(octokit, owner, repo, defaultBranchName);
for (const slug of event.teamNameSlugs) {
  await notify(event.fullName, config, slug);
}
console.log(`Notified teams ${event.teamNameSlugs.join(', ')}}`);
```

NovemberTang: The primary motivator behind doing it like this wasn't necessarily readability of the code, but clarity in the logs. I'm avoiding early returns here because I wanted to log **all** the reasons we've skipped branch protection, which makes testing/debugging easier and faster.

Keeping them on one line also makes rifling through logs a lot faster imo.

akash1810: I don't hold a strong enough opinion to block merging, but I wonder if the logs are easier to find like this as the `reason` variable is difficult to parse? I assume we'd find the logs by searching for "No action required."?

akash1810 (approved):

PR #496, opened 2023-11-20
Identify repos in PROD or INFRA that don't have a 'production' topic

tjsilver:
## What does this change?

This adds the functionality to ascertain if a repo that doesn't have a `production` topic has a cloudformation stack which has the tags of `PROD` or `INFRA`. This is the first part of the implementation to apply the topic to the repos automatically.  

Repos that have an `interactive` topic or that have the word 'interactive' in their name have also been excluded as we are not concerned with them at this time.  We are also excluding repos whose stacks sit in the Developer Playground account as they should not be in production (although there are some notable exceptions which are out of scope of this work).f

The next steps for this work are to message teams when a repo is found to be in production without a tag, then apply the topic automatically, and inform teams when it's done.

## Why?

This will help us to make sure we apply all the security measures to repos that are in production.

## How has it been verified?

Tested locally with the CODE databse.


packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -0,0 +1,75 @@
+import type { PrismaClient } from '@prisma/client';
+import type { JsonValue } from '@prisma/client/runtime/library';
+import type { Config } from '../config';
+import { getUnarchivedRepositories } from '../query';
+
+async function findProdStacks(client: PrismaClient): Promise<string[]> {
+	console.log('Discovering stacks with PROD or INFRA tags');
+	const stacks = await client.aws_cloudformation_stacks.findMany({
```
akash1810, line 8: Worth placing this query in [`query.ts`](https://github.com/guardian/service-catalogue/blob/main/packages/repocop/src/query.ts) for consistency?

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -0,0 +1,75 @@
...
+		const guRepoKey = 'gu:repo' as TagKey;
+		return Object.hasOwnProperty.call(tagValues, guRepoKey);
+	});
+	filtered.map((tags) => {
+		const tagValues = tags?.valueOf();
+		type TagKey = keyof typeof tagValues;
+		const guRepoKey = 'gu:repo' as TagKey;
+		tagValues &&
+			tagValues[guRepoKey] !== 'unknown' &&
+			repoNames.push(tagValues[guRepoKey]);
```
akash1810, line 42: This filter _could_ be done in SQL - worth adding an inline comment to explain why we're doing it in code?

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -0,0 +1,75 @@
+import type { PrismaClient } from '@prisma/client';
+import type { JsonValue } from '@prisma/client/runtime/library';
+import type { Config } from '../config';
+import { getUnarchivedRepositories } from '../query';
+
+async function findProdStacks(client: PrismaClient): Promise<string[]> {
```
akash1810, line 6: Is it worth making this a `Promise<Set<string>>` to remove any duplicates?

akash1810 (commented):
Could we add the word `cloudformation` (or `cfn`) to the function/variable names to help distinguish our multiple definitions of stack?

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
tjsilver: When you say 'in SQL', do you mean as a raw query or as part of the prisma query? @akash1810

akash1810: Ah, sorry, prisma.

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -0,0 +1,38 @@
...
+		`Found ${reposWithoutProductionTopic.length} repositories without a production topic`,
+	);
+
+	const repoTagsWithProdStacks = await findProdCfnStackTags(prisma);
+	console.log(
+		`Found ${repoTagsWithProdStacks.length} repos with a PROD or INFRA Cloudformation stack`,
+	);
+
+	const guReposWithProdCfnStacks: Array<string | undefined> =
+		repoTagsWithProdStacks.map((tag) => tag['gu:repo']);
```
akash1810, line 27: minor: Might be easier to filter out the `undefined` values, so this can be a `string[]` type:

```ts
const repos = repoTagsWithProdStacks.map((tag) => tag['gu:repo']).filter(r => !!r) as string[]
```

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -0,0 +1,38 @@
...
+
+	const reposInProdWithoutProductionTopic = reposWithoutProductionTopic.filter(
+		(repoName) => repoName && guReposWithProdCfnStacks.includes(repoName),
+	);
+
+	console.log(
+		`Found ${reposInProdWithoutProductionTopic.length} repos without a production tag that have a PROD or INFRA Cloudformation Stage tag`,
+	);
+	reposInProdWithoutProductionTopic.map((repoName) => console.log(repoName));
```
akash1810, line 36: This will produce a log line with a repo name, might be worth adding some context to it? Or add it to the previous log line?

```ts
console.log(
		`Found ${reposInProdWithoutProductionTopic.length} repos without a production tag that have a PROD or INFRA Cloudformation Stage tag: ${reposInProdWithoutProductionTopic.join(','}`,
	);
```

akash1810 (commented):
couple of minor, non-blocking comments

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
tjsilver: Oops I meant to take out the reponame console log! That was for my benefit!

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -0,0 +1,40 @@
...
+
+	const repoTagsWithProdStacks = await findProdCfnStackTags(prisma);
+	console.log(
+		`Found ${repoTagsWithProdStacks.length} repos with a PROD or INFRA Cloudformation stack`,
+	);
+
+	const guReposWithProdCfnStacks: Array<string | undefined> =
+		repoTagsWithProdStacks
+			.map((tag) => tag['gu:repo'])
+			.filter((r) => !!r) as string[];
```
akash1810, line 29: With the filter and casting, I think `guReposWithProdCfnStacks` can be typed as `string[]`? This type also allows a slight simplification of `reposInProdWithoutProductionTopic` below:

```ts
const reposInProdWithoutProductionTopic = reposWithoutProductionTopic.filter(
  (repoName) => guReposWithProdCfnStacks.includes(repoName),
);
```

JuliaBrigitte (approved):

PR #497, opened 2023-11-20
refactor: Initialise AWS clients in a consistent way

akash1810:
## What does this change?
Unifies how we initialise AWS clients for simplification, and to stay DRY. In this change we use a profile when running locally, and fallback to the default chain otherwise.

## Why?
Simpler code!

## How has it been verified?
- Ran each lambda locally
- [Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/e68731a9-5c44-4737-b4bc-00639afa4910) and executed each lambda successfully.

### DIFF
--- a/packages/branch-protector/src/index.ts
+++ b/packages/branch-protector/src/index.ts
@@ -1,5 +1,6 @@
 import type { Message } from '@aws-sdk/client-sqs';
 import { SQSClient } from '@aws-sdk/client-sqs';
+import { awsClientConfig } from 'common/aws';
 import { getGithubClient } from 'common/functions';
 import type { UpdateMessageEvent } from 'common/types';
 import type { Octokit } from 'octokit';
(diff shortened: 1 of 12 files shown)


packages/branch-protector/src/index.ts
```diff
@@ -78,7 +79,7 @@ async function handleMessage(
 export async function main() {
 	const config: Config = await getConfig();
 	const octokit: Octokit = await getGithubClient(config.githubAppConfig);
-	const sqsClient = new SQSClient({});
```
akash1810, line 81: This was previously working locally via the `AWS_PROFILE` and `AWS_REGION` environment variables set in `run-locally.ts` (below).

packages/common/src/aws.ts
```diff
@@ -0,0 +1,26 @@
+import { fromIni } from '@aws-sdk/credential-providers';
+import type { AwsCredentialIdentityProvider } from '@smithy/types';
+
+interface AwsClientConfig {
+	region: string;
+	credentials?: AwsCredentialIdentityProvider;
+}
+
+export function awsClientConfig(stage: string): AwsClientConfig {
```
NovemberTang, line 9: **for the future**: we should decide if we want to organise our common code by type (ie interface, function, class) or by concern (github, aws, utils), and be consistent

akash1810: 👍🏽  Grouping by concern is slightly more intention revealing IMO. Our current `function.ts` has a mixture of config, Anghammarad messaging, and JSON parsing.

PR #505, opened 2023-11-22
Add interactive monitor queue to setup script

NovemberTang:
## What does this change?

Allows us to fetch the interactive monitor topic ARN from SNS and set it as an env var

## Why?

Repcop will fail without this as it can't initialise the config object

## How has it been verified?

You should be able to run repocop locally now


tjsilver (approved):
Really useful! 👍

PR #504, opened 2023-11-22
feat(data-audit): Query the database

akash1810:
## What does this change?
Updates the data-audit lambda to perform a basic query against the database. There's a bit of wiring involved:
- Moving the database connection string logic into the `common` package
- Placing the lambda in the VPC, and adjust it's security groups to allow database connection
- Creation of a database user[^1]
- Include prisma in the lambda artifact

As such, it'll be easier to review this change [commit by commit](https://github.com/guardian/service-catalogue/pull/504/commits).

## Why?
The data-audit lambda will compare what is in the database to what AWS thinks, e.g. number of S3 buckets. This change begins that process.

## How has it been verified?
- Ran data-audit locally and observed the log line "There are X AWS accounts in the database"
- [Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/e4be404c-fa2d-4969-b102-040899c3cb2a) and manually invoked the lambda, and [observed the log line "There are X AWS accounts in the database"](https://logs.gutools.co.uk/s/devx/app/discover#/doc/626e7830-dd3a-11ea-88b6-37713dba5739/logstash-dev-tools-2023.11.22?id=LBWV94sBym0G9ucgjC-3)

[^1]: This user was created manually

PR #507, opened 2023-11-22
feat(data-audit): Check the number of accounts we have data for

akash1810:
Requires #504.

## What does this change?
Check that the number of accounts in the database is the same as the number of accounts in the AWS Organisation. At the moment we're just logging the result. We'll alarm/alert in the near future.

## Why?
We want Service Catalogue to collect information for all our AWS accounts.

## How has it been verified?
- Ran locally
- [Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/c6b3495a-2b40-45e8-8c16-413de5d13115), manually invoked the lambda, and observed the [logs](https://logs.gutools.co.uk/s/devx/goto/d1e852f0-8965-11ee-9308-f99e8aa2f574)

image

### DIFF
--- a/packages/data-audit/src/index.ts
+++ b/packages/data-audit/src/index.ts
@@ -1,10 +1,14 @@
+import {
+	OrganizationsClient,
+	paginateListAccounts,
+} from '@aws-sdk/client-organizations';
 import { PrismaClient } from '@prisma/client';
+import { awsClientConfig } from 'common/aws';
+import type { Config } from './config';
 import { getConfig } from './config';
 
-export async function main() {
-	const config = await getConfig();
-
-	const prisma = new PrismaClient({
+function numberOfAwsAccountsFromDatabase(config: Config): Promise<number> {
+	const client = new PrismaClient({
 		datasources: {
 			db: {
 				url: config.databaseConnectionString,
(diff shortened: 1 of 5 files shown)


packages/data-audit/src/index.ts
```diff
@@ -19,7 +23,34 @@ export async function main() {
...
+		{
+			client,
+			pageSize: 10,
+		},
+		{},
+	)) {
+		total += page.Accounts?.length ?? 0;
+	}
+	return total;
+}
```
akash1810, line 44: The list accounts endpoint has a maximum page size of 20, so we have to [paginate](https://aws.amazon.com/blogs/developer/pagination-using-async-iterators-in-modular-aws-sdk-for-javascript/) the response. The use of `let` is necessary here; it's scoped to the function, so it's mutability isn't too bad.

packages/data-audit/src/index.ts
```diff
@@ -19,7 +23,33 @@ export async function main() {
...
```
akash1810, line 43: The list accounts endpoint has a maximum page size of 20, so we have to [paginate](https://aws.amazon.com/blogs/developer/pagination-using-async-iterators-in-modular-aws-sdk-for-javascript/) the response. The use of let is necessary here; it's scoped to the function, so it's mutability isn't too bad.

packages/data-audit/src/index.ts
```diff
@@ -19,7 +23,33 @@ export async function main() {
...
+
+async function numberOfAwsAccountsFromAws(config: Config): Promise<number> {
+	const client = new OrganizationsClient(awsClientConfig(config.stage));
+
+	let total = 0;
+	for await (const page of paginateListAccounts(
+		{
+			client,
+			pageSize: 10,
```
NovemberTang, line 36:
```suggestion
			pageSize: 20,
```

NovemberTang (approved):
approved w a minor nitpick

PR #508, opened 2023-11-23
feat(data-audit): Audit the number of S3 buckets

akash1810:
Requires #507.

## What does this change?
Check that number of S3 buckets in the database matches the number of S3 buckets in AWS. At the moment we're just logging the result. We'll alarm/alert in the near future.

## Why?
Data accuracy is important!

## How has it been verified?
TBD.


AshCorr (approved):

akash1810:
> It highlights that we are missing 13 buckets.

This number is true of CODE. In PROD, the numbers are DB 1530, AWS 1531. Presumably a bucket has been created after the task completed.

PR #511, opened 2023-11-23
Fix migration script

NovemberTang:
## What does this change?

Prisma code has been moved to common, this PR updates the migration script to reflect this.

I haven't moved it to the common folder as repocop populates basically all the non cloudquery tables created by the migration, so I think it makes sense to keep it here for the time being

## Why?

The script was failing due to the prisma package having been moved

## How has it been verified?

DEV migrations succeed.
TODO: test CODE migration

### DIFF
--- a/packages/repocop/script/migrate.sh
+++ b/packages/repocop/script/migrate.sh
@@ -62,25 +62,30 @@ migrateDeploy() {
   continueMigrate "$stage"
 }
 
-# Read in input stage(s)
+stageAwareMigration() {
+  if [ "$stage" == "--dev" ]
+  then
+    devSetup
+    if [ -n "$migration_name" ]
+    then
+      migrateDEV "$migration_name"
+    else
+      :
+    fi
+  elif [ "$stage" == "--code" ]
+  then
+    migrateDeploy "CODE"
+  elif [ "$stage" == "--prod" ]
+  then
+    migrateDeploy "PROD"
+  else
+    echo -e "Stage '$stage' not recognised, please try again."
+  fi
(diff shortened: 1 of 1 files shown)


packages/repocop/script/migrate.sh
```diff
@@ -62,25 +62,30 @@ migrateDeploy() {
...
-  migrateDeploy "PROD"
-else
-  echo -e "Stage '$stage' not recognised, please try again."
-fi
\ No newline at end of file
+}
+
+(
+stage=$1
+migration_name=$2
```
akash1810, line 88: minor: should these two lines be indented too?

akash1810 (approved):
Thanks!

I think this is more evidence to exercise a migration as part of our CI checks.

akash1810:
Confirmed this to be working as expected.

```console
➜ npm -w repocop run migrate:code                                                     

> repocop@1.0.0 migrate:code
> script/migrate.sh --code

NOTE: You need to be on the VPN to migrate. 
Migration may fail if the table(s) already exist in the database.
See https://www.prisma.io/docs/guides/migrate/production-troubleshooting#failed-migration for how to fix.
WARNING: Changes made on your local machine will affect Service Catalogue CODE data as all pending migrations will be applied.
Do you want to continue (y/n)? (No is default)
y
CODE migration in progress...
Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "postgres", schema "public" at "REDACTED"

15 migrations found in prisma/migrations


No pending migrations to apply.

```

PR #513, opened 2023-11-23
Move branch protector to repocop

NovemberTang:
## What does this change?

Moves branch protector logic to repocop.

There's additional cleanup to be done as part of this work, like integrating branch protector into existing repocop abstractions. This PR is the simplest lift-and-shift (and delete) portion of the work.

## Why?

Following on from #509 , now that we are only messaging teams once about their repos, the logic of branch protector is simple enough to be a part of repocop proper. I thing this will also come with minor performance benefits outlined below

Repocop takes a 3-8 seconds to run. This includes coldestart times, reading and deleting messages from SQS, making database calls, on top of the bit we actually care about, which is protecting branches. Doing this as part of repocop means we only have to start up one lambda, and can completely eliminate SQS and DB calls, meaning less time waiting around for external resources to do their thing

## How has it been verified?

Deployed to CODE, and behaves as expected. Duration has increased by about 2s, and we've dropped a lambda that was taking about 5s to run


packages/cdk/lib/service-catalogue.ts
```diff
@@ -173,13 +173,12 @@ export class ServiceCatalogue extends GuStack {
...
 
-		const branchProtector = new BranchProtector(
+		const githubCredentials = new Secret(
 			this,
-			stageAwareMonitoringConfiguration,
-			nonProdSchedule ??
-				Schedule.cron({ minute: '0', hour: '9', weekDay: 'TUE-FRI' }),
-			vpc,
-			anghammaradTopic,
+			`branch-protector-github-app-auth`,
```
NovemberTang, line 178: The secret still refers to branch protector so it's not destroyed and recreated on deployment. It should be renamed at some point, but this is not a huge priority.

packages/repocop/src/config.ts
```diff
@@ -80,10 +75,7 @@ export async function getConfig(): Promise<Config> {
 		interactiveMonitorSnsTopic: getEnvOrThrow('INTERACTIVE_MONITOR_TOPIC_ARN'),
 		databaseConnectionString: getDatabaseConnectionString(databaseConfig),
 		withQueryLogging: queryLogging,
-		branchProtectorQueueUrl: getEnvOrThrow('BRANCH_PROTECTOR_QUEUE_URL'),
-		topicMonitoringProductionTagQueueUrl: getEnvOrThrow(
-			'BRANCH_PROTECTOR_QUEUE_URL',
-		), // TODO: remove this
+		topicMonitoringProductionTagQueueUrl: '???', // TODO: remove this
```
NovemberTang, line 78: this url isn't being used at the moment, so replacing it with a generic placeholder value

packages/repocop/src/remediations/branch-protector/repository-02-branch_protection.ts
```diff
@@ -0,0 +1,118 @@
+import type {
```
NovemberTang, line 1: this file looks like it's new, but it's actually just been moved from branch-protector and GitHub hasn't realised

packages/repocop/src/remediations/shared-utilities.ts
```diff
@@ -28,11 +26,6 @@ type AnghammaradMessages = {
 };
 
 const anghammaradMessages: AnghammaradMessages = {
-	branchProtector: {
```
NovemberTang, line 31: removing this for now as the abstraction is not quite flexible enough to be useful here

AshCorr (approved):
LGTM!

PR #515, opened 2023-11-24
Use a function instead of a view for collecting last sync times

AshCorr:
## What does this change?

Replace our cq_last_synced view with a cq_last_synced() function. This function does much the same thing but figures out which tables it needs to query at runtime, instead of it being hardcoded at creation.

## Why?

Because the old `cq_last_synced` view had hardcoded references to all Cloudquery tables in its DDL it caused Cloudquery to not be able to update table schemas due to Cascade issues. Since this new function doesn't directly reference any Cloudquery tables it hopefully shouldn't impact anything.

## How has it been verified?

Deployed this function and ran `SELECT * FROM cq_last_synced()`

image

Also verified that we're now able to drop Cloudquery tables again without having cascade issues
image


akash1810 (commented):
Ah, brilliant! Is it worth getting https://github.com/guardian/service-catalogue/pull/511 in before this, as it'll be a nice way to test it?

akash1810 (approved):

PR #519, opened 2023-11-24
Add tests for topic monitor (production tags)

tjsilver:
## What does this change?

Adds some tests for the topic monitor for production topics.

### DIFF
--- a/packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
+++ b/packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
@@ -61,15 +61,9 @@ describe('getGuRepoNames', () => {
 				'gu:repo': 'guardian/repo-1',
 				'gu:build-tool': 'guardian/some-build-tool',
 			},
-			{
-				App: 'app-2',
-				Stack: 'stack2',
-				Stage: 'INFRA',
-				'gu:repo': 'guardian/repo-2',
-			},
 		];
 
 		const result: string[] = getGuRepoNames(cfnTags);
-		expect(result).toEqual(['guardian/repo-1', 'guardian/repo-2']);
+		expect(result).toEqual(['guardian/repo-1']);
 	});
 });
--- a/packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
+++ b/packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
@@ -66,21 +66,10 @@ describe('getGuRepoNames', () => {
 				Stack: 'stack2',
 				Stage: 'INFRA',
 				'gu:repo': 'guardian/repo-2',
-				'gu:build-tool': 'guardian/tool',
-			},
-			{
-				App: 'app-3',
-				Stack: 'stack3',
-				Stage: 'PROD',
-				'gu:repo': 'guardian/repo-3',
 			},
 		];
 
 		const result: string[] = getGuRepoNames(cfnTags);
-		expect(result).toEqual([
-			'guardian/repo-1',
-			'guardian/repo-2',
-			'guardian/repo-3',
-		]);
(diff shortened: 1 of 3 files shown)


packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
```diff
@@ -0,0 +1,86 @@
...
+		const cfnTags: AWSCloudformationTag[] = [
+			{
+				App: 'app-1',
+				Stack: 'stack1',
+				Stage: 'PROD',
+				'gu:repo': 'guardian/repo-1',
+				'gu:build-tool': 'guardian/some-build-tool',
+			},
+			{
+				App: 'app-2',
```
NovemberTang, line 65: minor, non-blocking: as we're testing the same tag keys in every element of this array, it might quicker to understand with only one example in the list

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -1,36 +1,61 @@
...
+export async function findReposInProdWithoutProductionTopic(
+	prisma: PrismaClient,
+	unarchivedRepos: github_repositories[],
+) {
+	console.log('Discovering Cloudformation stacks with PROD or INFRA tags');
+	const reposWithoutProductionTopic: string[] =
+		getReposWithoutProductionTopic(unarchivedRepos);
 	console.log(
-		`Found ${repoTagsWithProdStacks.length} repos with a PROD or INFRA Cloudformation stack`,
+		`Found ${reposWithoutProductionTopic.length} repositories without a production topic`,
```
NovemberTang, line 40: do we need to be explicit about interactives also being excluded?

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -1,36 +1,61 @@
...
 			(repo) =>
 				!repo.topics.includes('production') &&
 				!repo.topics.includes('interactive'),
 		)
 		.map((repo) => repo.full_name)
-		.filter((name) => !name?.includes('interactive'));
-	console.log(
-		`Found ${reposWithoutProductionTopic.length} repositories without a production topic`,
+		.filter((name) => !name?.includes('interactive'))
+		.filter((name) => !!name) as string[];
```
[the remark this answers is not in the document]
tjsilver: Line 16?  IIUC `!!name` is ensuring that `name` is not undefined/null to make TS happy!

packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
tjsilver: Hmm yeah I wasn't sure about that - will clarify!

packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
tjsilver: They're not quite the same but I take your point - will vary them more!

packages/repocop/src/remediations/repository-06-topic-monitor-production.test.ts
```diff
@@ -0,0 +1,69 @@
...
+} from './repository-06-topic-monitor-production';
+
+describe('getReposWithoutProductionTopic', () => {
+	it('should return an empty array when unarchivedRepos array is empty', () => {
+		const unarchivedRepos: github_repositories[] = [];
+		const result: string[] = getReposWithoutProductionTopic(unarchivedRepos);
+		expect(result).toEqual([]);
+	});
+
+	it('should return only repositories without production or interactive tags and without "interactive" in the repo name', () => {
```
akash1810, line 16: minor: should this description be for repository _topics_ rather than tags?

NovemberTang (approved):

PR #521, opened 2023-11-25
chore(deps): Update to Node v20.10.0 (LTS)

akash1810:
## What does this change?
Moves to Node 20 (LTS) now that [AWS Lambda supports it](https://aws.amazon.com/blogs/compute/node-js-20-x-runtime-now-available-in-aws-lambda/).

This change also includes an update to Prisma's configuration, adopting OpenSSL v3. OpenSSL v1 reached end of life in September 2023. Node v20 has dropped it entirely now.

See:
  - https://nodejs.org/en/blog/announcements/v20-release-announce#call-to-action
  - https://nodejs.org/en/blog/announcements/nodejs16-eol/

## Why?
[This is the current active long term support version](https://nodejs.org/en/about/previous-releases).

## How has it been verified?
- Successfully ran lambdas locally
- [Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/d98c31d8-0914-4c74-85bf-a116c0973581) and successfully ran lambdas


tjsilver (approved):
LGTM!

PR #522, opened 2023-11-27
refactor: Use the type system to simplify the code

akash1810:
## What does this change?
Making `password` required in the `DatabaseConfig` interface allows us to remove an "if undefined" check.

## Why?
Simpler logic.

## How has it been verified?
N/A.

### DIFF
--- a/packages/common/src/database.ts
+++ b/packages/common/src/database.ts
@@ -15,11 +15,9 @@ export interface DatabaseConfig {
 	user: string;
 
 	/**
-	 * The database password.
-	 *
-	 * When not defined, a token (temporary password) will be generated for IAM authentication for RDS.
+	 * The database password for the provided user.
 	 */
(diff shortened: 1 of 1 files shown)


tjsilver (commented):
How will this affect the local running story - whereby we use the lack of password to trigger the creation of getting the RDS token if we want to connect to the CODE db?

akash1810:
> How will this affect the local running story - whereby we use the lack of password to trigger the creation of getting the RDS token if we want to connect to the CODE db?

This change does not change the existing behaviour. The behaviour was changed in #504 to [throw if the password env var is unset](https://github.com/guardian/service-catalogue/blob/d5e69c0ecf73f65fe0ffd549402ac7f2adc5e453/packages/common/src/database.ts#L60).

NovemberTang (approved):

PR #524, opened 2023-11-27
Expands S3 search to catch interactives.

NovemberTang:
## What does this change?

Interactives are usually be found in the `atoms` prefix in S3, but may live in the root. The prefix paths are usually defined in the `config.json`, but for really old ones, they might be defined in `cfg/s3.json`. This PR tries to handle all these scenarios.

We do end up making more requests to GitHub and S3, so the Lambda is less performant. However, I think evaluating all the booleans at the end makes for easier reading, and while interactive monitor is still new, I'd like to be able to see all the checks a given repo did/didn't pass in the logs, which would not always be possible if we terminated after the first successful check. Additionally, the lambda is making a small number of requests overall, so the risk of being rate limited is low.

## How has it been verified?

The code has been tested locally against a repo that has a `config.json`, another that has a `cfg/s3.json`, and one more that has neither, to verify the new logic works in all these cases.


package-lock.json
```diff
@@ -12273,6 +12280,18 @@
...
+      "extraneous": true,
+      "dependencies": {
+        "@aws-sdk/client-sqs": "^3.438.0",
+        "@aws-sdk/client-ssm": "^3.438.0",
+        "@aws-sdk/types": "^3.433.0",
+        "@guardian/anghammarad": "^1.8.1",
+        "@octokit/types": "^12.3.0",
+        "octokit": "^3.1.2"
+      }
+    },
```
akash1810, line 12294: I wasn't expecting the `package-lock.json` to be updated with this change. Suspect you've got some unwanted artifacts on your local machine as `branch-protector` has been removed now?

akash1810 (approved):

PR #525, opened 2023-11-27
send fewer messages on non-prod stages

NovemberTang:
## What does this change?

Only send one message if we are not running on prod

## Why?

We don't need to run the lambda 10 times to verify it works. Also, if we want to test the interactive monitor, the easiest thing to do would be to submit the name of the repo directly to SNS.

We're also getting rate limited by AWS on prod so let's lower the number of repos we send for now

### DIFF
--- a/packages/repocop/src/remediations/repository-06-topic-monitor-interactive.ts
+++ b/packages/repocop/src/remediations/repository-06-topic-monitor-interactive.ts
@@ -37,7 +37,10 @@ export async function sendPotentialInteractives(
 	const potentialInteractives = shuffle(
 		findPotentialInteractives(evaluatedRepos),
 	);
-	const snsBatchMaximum = Math.min(potentialInteractives.length, 10);
+	const snsBatchMaximum = Math.min(
+		potentialInteractives.length,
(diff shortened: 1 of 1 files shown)


packages/repocop/src/remediations/repository-06-topic-monitor-interactive.ts
```diff
@@ -37,7 +37,10 @@ export async function sendPotentialInteractives(
 	const potentialInteractives = shuffle(
 		findPotentialInteractives(evaluatedRepos),
 	);
-	const snsBatchMaximum = Math.min(potentialInteractives.length, 10);
+	const snsBatchMaximum = Math.min(
+		potentialInteractives.length,
+		config.stage === 'PROD' ? 5 : 1,
```
akash1810, line 42: I think it would be better to have this as part of the config, as it'll reduce the amount of branching needed. That is, if config had a `maxMessages` property (and other stage aware values), we'd need to perform the stage check once.

akash1810 (approved):
Added a comment. Not a blocker, but think we'd benefit from implementing the suggestion soon.

PR #526, opened 2023-11-27
Bunch potential interactives into one lambda invocation

NovemberTang:
## What does this change?

Previously, we were sending 10 individual messages to SNS at once. Now, we are creating a JSON list of x repositories, and submitting it in one invocation.

## Why?

Fewer lambda invocations means better performance overall. Theoretically we could send more in one go, but leaving it at 10 for now to keep us under AWS API's abuse limits, which seem to be triggered after about 10 repos have been checked. We can make the interactive monitor a bit smarter about which checks it runs in a later PR.

## How has it been verified?

Deployed to CODE and verified the messages are processed correctly

### DIFF
--- a/packages/interactive-monitor/package.json
+++ b/packages/interactive-monitor/package.json
@@ -6,7 +6,7 @@
 	"scripts": {
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node20 --outdir=dist --external:@aws-sdk",
 		"test": "echo \"No test specified\" && exit 0",
-		"start": "AWS_PROFILE=deployTools ts-node src/run-locally.ts"
+		"start": "ts-node src/run-locally.ts"
 	},
 	"author": "guardian",
 	"dependencies": {
--- a/packages/interactive-monitor/package.json
+++ b/packages/interactive-monitor/package.json
@@ -6,7 +6,7 @@
 	"scripts": {
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node20 --outdir=dist --external:@aws-sdk",
 		"test": "echo \"No test specified\" && exit 0",
-		"start": "APP=interactive-monitor ts-node src/run-locally.ts"
+		"start": "AWS_PROFILE=deployTools ts-node src/run-locally.ts"
 	},
 	"author": "guardian",
 	"dependencies": {
--- a/packages/interactive-monitor/src/index.ts
+++ b/packages/interactive-monitor/src/index.ts
@@ -136,7 +136,6 @@ export async function assessRepo(repo: string, owner: string, config: Config) {
 export const handler: SNSHandler = async (event) => {
 	const config = getConfig();
 	const owner = 'guardian';
-	console.log(JSON.stringify(event.Records));
 	const events = event.Records.map(async (record) =>
 		Promise.all(
 			(JSON.parse(record.Sns.Message) as string[]).map(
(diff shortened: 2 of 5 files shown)


packages/interactive-monitor/package.json
```diff
@@ -6,7 +6,7 @@
 	"scripts": {
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node20 --outdir=dist --external:@aws-sdk",
 		"test": "echo \"No test specified\" && exit 0",
-		"start": "APP=interactive-monitor ts-node src/run-locally.ts"
+		"start": "AWS_PROFILE=deployTools ts-node src/run-locally.ts"
```
akash1810, line 9: Is this change intentional?

NovemberTang: Yes. the `APP` variable isn't being used for anything, and setting the `AWS_PROFILE` means we can make S3 requests

akash1810: If we use https://github.com/guardian/service-catalogue/blob/f3dbbdebd52e3d2185f01bbe4bdec0630a75c90e/packages/common/src/aws.ts#L22-L30 we shouldn't need to set `AWS_PROFILE`. This would also make us consistent too.

packages/interactive-monitor/src/index.ts
```diff
@@ -100,7 +101,8 @@ async function applyTopics(repo: string, owner: string, octokit: Octokit) {
 
 export async function assessRepo(repo: string, owner: string, config: Config) {
 	const octokit = await stageAwareOctokit(config.stage);
-	const s3 = new S3Client({ region: 'us-east-1' });
+	const awsConfig = awsClientConfig(config.stage);
+	const s3 = new S3Client({ ...awsConfig, region: 'us-east-1' });
```
NovemberTang, line 105: overriding the region rather than making the region configurable, as there are unlikely to be many scenarios where eu-west-1 is not appropriate

packages/interactive-monitor/package.json
NovemberTang: we're using this now, see later commits

packages/interactive-monitor/src/index.ts
```diff
@@ -134,8 +136,12 @@ export async function assessRepo(repo: string, owner: string, config: Config) {
 export const handler: SNSHandler = async (event) => {
 	const config = getConfig();
 	const owner = 'guardian';
-	const events = event.Records.map(
-		async (record) => await assessRepo(record.Sns.Message, owner, config),
+	const events = event.Records.map(async (record) =>
```
akash1810, line 139: What type is `events` here? I think it might be an array of arrays? Not sure that the `Promise.all` on line 146 would work as we'd expect, and we might want to call [`.flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) events (or use [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)) to produce a one-dimensional array.

akash1810 (commented):
added a minor comment

packages/interactive-monitor/src/index.ts
NovemberTang: Having run this on CODE, all the logs look normal, what unexpected behaviour do you anticipate?

tjsilver (approved):
👍

PR #529, opened 2023-11-27
interactive-monitor: Make fewer requests to S3 per repo

NovemberTang:
## What does this change?

interactives with a `config.json` deploy into a different prefix than those with a `cfg/s3.json`. We know what these prefixes are so can search only in the correct parts of the bucket for artifacts.

## Why?

Halving the number of requests we are making to S3 means we can look at more of them at one time

## How has it been verified?

Local execution still works

### DIFF
--- a/packages/interactive-monitor/src/index.ts
+++ b/packages/interactive-monitor/src/index.ts
@@ -78,15 +78,13 @@ async function s3PathIsInConfig(
 	owner: string,
 	repo: string,
 	path: string,
+	prefix?: string,
 ): Promise<boolean> {
 	const configJson = await getConfigJsonFromGithub(octokit, repo, owner, path);
 	if (configJson === undefined) {
 		return false;
 	} else {
-		const s3Response1 = !!(await findInS3(s3, `atoms/${configJson.path}`));
-		const s3Response2 = !!(await findInS3(s3, configJson.path));
-
-		return s3Response1 || s3Response2;
+		return !!(await findInS3(s3, `${prefix ?? ''}/${configJson.path}`));
 	}
 }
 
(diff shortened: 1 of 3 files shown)


packages/interactive-monitor/src/index.ts
```diff
@@ -78,16 +78,49 @@ async function s3PathIsInConfig(
...
+	repo: string,
+	owner: string,
+): Promise<boolean> {
+	const foundInConfigJson = await s3PathIsInConfig(
+		octokit,
+		s3,
+		owner,
+		repo,
+		'config.json',
+		'atoms',
```
akash1810, line 104: Does this evaluate to `atoms/config.json`? Might it be simpler to pass that in directly, and not add the optional `prefix` to the `s3PathIsInConfig` function?

packages/interactive-monitor/src/index.ts
```diff
@@ -78,16 +78,49 @@ async function s3PathIsInConfig(
...
+	if (foundInConfigJson) {
+		return true;
+	}
+
+	const foundInS3Json = await s3PathIsInConfig(
+		octokit,
+		s3,
+		owner,
+		repo,
+		'cfg/s3.json',
```
akash1810, line 116: Why is this not `s3PathIsInConfig(..., 's3.json', 'cfg`) like `config.json` above?

packages/interactive-monitor/src/index.ts
NovemberTang: No, `config.json` is the name of the file we are looking for in github, `atoms` is a prefix we are looking under in S3. I can rename the function parameters to make that clearer

packages/interactive-monitor/src/index.ts
NovemberTang: similar reasoning. cfg/s3.json is the full path to the config from the root of the github repo, and there's nothing prepended to the s3 path so that parameter has been left empty.

packages/interactive-monitor/src/index.ts
NovemberTang: resolving this comment and the one below as the parent function has been deleted

tjsilver (approved):

PR #531, opened 2023-11-28
Add final interactive check.

NovemberTang:
## What does this change?

This adds a third way of checking if a repo is an interactive, by checking the contents of `project.config.js`. This is a little more complicated as it's a javascript file, not JSON.

The evaluation of these checks has also been delegated to a series of functions, rather than variables, so that they evaluate lazily, improving performance and reducing the number of GitHub/S3 calls (The S3 API has a particularly low abuse limit, so we want to avoid making calls unless absolutely neccessary)

## Why?

New interactives generally contain this `project.config.js`, and use it for deployment. We should check for them in case someone has created a repo in a way other than template generation.

## How has it been verified?

New unit tests pass, and expected behaviour is observed in repositories with and without the file.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -36,5 +36,12 @@ module.exports = {
 			],
 			testMatch: ['<rootDir>/packages/repocop/**/*.test.ts'],
 		},
+		{
+			displayName: 'interactive-monitor',
+			transform: {
+				'^.+\\.tsx?$': 'ts-jest',
+			},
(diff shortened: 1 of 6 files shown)


NovemberTang:
> Looks good - do you have a sense of roughly how many interactives are covered by this new rule?

I don't have a solid number. But I doubt it'll be many. Most of these repos will be caught by the template check first. I've moved it to the end of the list to reflect this.

PR #533, opened 2023-11-30
chore: Reduce build time noise from linter

akash1810:
Add some annotations to help the linter understand we know what we're doing, and therefore reduce noise in the build.

## How has it been verified?
Observe CI - the annotations appearing on `main` have been removed on this branch.

<details><summary>Annotations on main</summary>
<p>

image

</p>
</details>

<details><summary>Annotations on this branch</summary>
<p>

image

</p>
</details>

### DIFF
--- a/packages/interactive-monitor/src/js-parser.test.ts
+++ b/packages/interactive-monitor/src/js-parser.test.ts
@@ -5,8 +5,11 @@ describe('getPathFromConfigFile', () => {
 		const rawFile1 = String.raw`export default {
             title: "Iran protests",
             path: "2022/10/iran-protests"}`;
-		const file1 = parseFileToJS(rawFile1);
-		expect(getPathFromConfigFile(file1!)).toEqual('2022/10/iran-protests');
+
+		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- content provided above, so can be sure this is never `undefined`
+		const file1 = parseFileToJS(rawFile1)!;
+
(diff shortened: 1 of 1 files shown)


packages/interactive-monitor/src/js-parser.test.ts
```diff
@@ -18,19 +21,21 @@ describe('getPathFromConfigFile', () => {
...
 	it('should return undefined if the JS file is invalid', () => {
 		const rawFile = String.raw`export default {
 		    title: "Iran protests",
 		    asdf: "2022/10/iran-protests"}`;
-		const file = parseFileToJS(rawFile);
-		expect(getPathFromConfigFile(file!)).toBeUndefined();
-	});
-	it('should return undefined if the JS file is empty', () => {
-		const file = parseFileToJS('');
-		expect(getPathFromConfigFile(file!)).toBeUndefined();
```
akash1810, line 34: The `getPathFromConfigFile` function takes an argument of type `SourceFile`. This test is testing what happens when `undefined` is passed to `getPathFromConfigFile`,
which is confusing, and not needed.

NovemberTang: This looks to me like it's testing an empty file, not an undefined one, unless I've missed something?

akash1810: Ah! Hadn't realised that, thanks for the clarification. Updated.

NovemberTang (approved):

PR #536, opened 2023-11-30
New view view_ec2_instance_types

JuliaBrigitte:
## What does this change?
Added to the view view_running_instances.

## Why?
To help with reservations dashboard.

## How has it been verified?
Changed the view on code


packages/common/prisma/migrations/20231130122000/migration.sql
```diff
@@ -0,0 +1,18 @@
...
+ The view below is used to to indentify the instance types of all running EC2 instances in the environment.
+ Providing account name, instance type, app name, instance id, and launch time.
+ */
+create or replace view view_ec2_instance_types as
+SELECT *
+FROM (
+         SELECT DISTINCT ON (instances.instance_id)
+             accts.name               as account_name,
+             instances.instance_type as type,
+             instances.tags ->> 'App' as app,
```
akash1810, line 11: The `App` tag in isolation is not guaranteed to be unique. For example, it could have a value of `api`. Should we extract the `Stack`, `Stage` tags too? Should we also display the AWS region?

packages/common/prisma/migrations/20231130122000/migration.sql
```diff
@@ -0,0 +1,18 @@
+/*
+ The view below is used to to indentify the instance types of all running EC2 instances in the environment.
+ Providing account name, instance type, app name, instance id, and launch time.
+ */
+create or replace view view_ec2_instance_types as
```
akash1810, line 5: Would `view_running_ec2_instances` be a more intention revealing name for this view? The current name is quite simpler to [CloudQuery's `aws_ec2_instance_types` table](https://hub.cloudquery.io/plugins/source/cloudquery/aws/v22.19.2/tables/aws_ec2_instance_types) which holds very different data.

JuliaBrigitte: Actually we have a view running instances, I do not know where it is used so I did not change/extend it with what I needed. Do you think that would be better?

akash1810: Ah! I think the only usage of `view_running_instances` is the [Enterprise transformation programme data](https://metrics.gutools.co.uk/d/XgwFt8r4z/kpmg?orgId=1) dashboard.

I think adding to this view would be simpler to maintain, vs having two views that look very similar. WDYT?

packages/common/README.md
akash1810, line 1: Since https://github.com/guardian/service-catalogue/pull/539 the database migration instructions have moved to [docs/database-migrations.md](https://github.com/guardian/service-catalogue/blob/main/docs/database-migrations.md). The content of this file is also stale now. Would it be simpler to decouple these changes across multiple PRs?

packages/common/prisma/migrations/20231130122000/migration.sql
JuliaBrigitte: Great suggestion, I added Stack and Stage

packages/common/prisma/migrations/20231130122000/migration.sql
JuliaBrigitte: Test both dashboard and will now used add to view_running_instances

packages/common/README.md
JuliaBrigitte: removed from this PR

packages/common/prisma/migrations/20231220080000_view_running_instances/migration.sql
```diff
@@ -0,0 +1,38 @@
...
+                                           instances.tags ->> 'App' as app,
+                                           instances.image_id,
+                                           instances.instance_id,
+                                           CASE
+                                               WHEN images.built_by_amigo THEN true
+                                               ELSE false
+                                               END                  as built_by_amigo,
+                                           cast(instances.launch_time as date),
+                                           instances.instance_type as type,
+                                           instances.launch_time as launch_time_stamp,
```
akash1810, line 30: Why are we pulling out `instances.launch_time` twice? Would it be simpler to select it once, and have the consumer cast it if needed?

akash1810 (approved):
Can we update the PR description too?

PR #538, opened 2023-11-30
Check for markdown changes before passing CI

NovemberTang:
## What does this change?

Add a CI check to make sure

1. the generate script still works
2. we cannot merge if the markdown file does not match the contents of `definitions.ts`

## Why?

1. The best practices project had been silently broken all week as we did not have a CI check for it.
2. Keeping the markdown file up to date is important so teams can easily see what's expected of their repos

## How has it been verified?

Tested with matching and mismatching markdown and typescript files, saw the expected output in both cases.


akash1810 (approved):
YEAH! 🏆

PR #539, opened 2023-12-01
feat(database): Support total Prisma migration

akash1810:
Support the scenario where a database migration has gone wrong, wrong enough that we've deleted the `_prisma_migrations` table.

There are now two new commands:
  1. `npm -w repocop run migrate:code:from-start`
  2. `npm -w repocop run migrate:prod:from-start`

When used, these scripts will first perform a [baseline migration](https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate/add-prisma-migrate-to-a-project#baseline-your-production-environment) to avoid data loss. The remaining migrations will then be applied as normal.

To enable this, the existing migrations have been edited to allow them to be infinitely replayed. This just means `IF EXISTS` has been added.

Note, updating previous migration scripts changes the file's checksum. As the checksum is a column in the `_prisma_migrations` table, we'd need to perform `npm -w repocop run migrate:prod:from-start` on PROD upon merge.

## How has it been verified?
I've successfully run `npm -w repocop run migrate:code:from-start` on CODE multiple times.


packages/common/prisma/migrations/20230907105726_create_table_repocop_github_repository_rules/migration.sql
```diff
@@ -1,3 +1,5 @@
+DROP TABLE IF EXISTS repocop_github_repository_rules;
```
AshCorr, line 1: Why is this table in particular dropped?

AshCorr (approved):
I feel like "total prisma migration" sounds way too cool for what its actually doing

packages/common/prisma/migrations/20230928080813_guardian_non_p_and_e_github_teams/migration.sql
```diff
@@ -1,3 +1,5 @@
+DROP TABLE IF EXISTS guardian_non_p_and_e_github_teams;
```
AshCorr, line 1: and this table too? is there a chance someone might manually add a value to this table which will be lost if we do a total migration?

akash1810:
Confirming that I've applied this to PROD this morning using:

```sh
npm -w repocop run migrate:prod:from-start
```

PR #540, opened 2023-12-01
Highlight unmaintined repos

NovemberTang:
## What does this change?

This creates a **preliminary** definition of an unmaintained/unused repo, and applies it to repocop. I expect this definition to evolve with time, maybe to take into account things like CFN stacks or action run dates, but definition allows us to catch a reasonably sized group of repos that have been inactive for a while and are potential candidates for archiving.

There are three main timestamp fields associated with a repository

1. `created_at`. The date of the repo's creationg
2. `pushed_at`. The date a code change was last made
3. `updated_at`. The date of the most recent metadata change. This includes things like branch 

This PR proposes we mark a repo as unmaintained if none of these dates fall within the last two years.

## Why?

Two years is quite a long time for a repo to have gone untouched, especially if we include the updated_at field, as we gave nearly every unarchived repo a topic at some point between Apr 2022 and Nov 2023, which would be reflected in `updated_at`. This gives us a list of repositories that we missed or ignored during that time. It's possible that teams are unaware that they own these projects.

Using this definition of unmaintained, by mid 2024, more repos will start to appear in this group, including repos that have not been updated since the great en-Snyk-ening of Spring 2022.  We can provide these lists to teams so they can make better decisions about what to keep in their estate, how to prioritise things like dependency updates, and cost saving measures, by shutting down unused services.

## How has it been verified?

This has been tested on PROD


tjsilver (approved):
This looks good to me, and important to know 👍

PR #545, opened 2023-12-01
feat(repocop): Find live stacks associated with archived repos

NovemberTang:
## What does this change?

- Adds a new rule that expects archived repos to not have any corresponding stacks running on AWS.
- Creates a function evaluating said rule against a repo and known cloudformation stacks, using the tags, or the repository name as a backup

## Why?

Archived repos are exempt from all security expectations (and most other expectations). If there are archived repos running code on AWS, we need to know about it, as these systems are not receiving patches and may be vulnerable. N.B. This logic is not perfectly precise, and there are likely to be a few false negatives (if a stack has been arbitrarily named) or false positives (more likely with very short repo names). But the results in the logs are promising, and this should help us to catch low hanging fruit, saving money on infrastructure, and keeping our estate more secure and up-to-date.

## How has it been verified?

Deployed to code, checked logs, and verified the expected behaviour.

## Performance

This adds about 4 seconds to the lambda There are performance improvements we could make to reduce the number of DB calls, and potentially to the order of filters, but these may not make a huge difference, and to keep this PR more focused, they have not been included for now


packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -51,17 +47,13 @@ export function getRepoNamesWithoutProductionTopic(
...
-
 export function getReposInProdWithoutProductionTopic(
 	reposWithoutProductionTopic: string[],
 	guRepoStacks: GuRepoStack[],
 ): GuRepoStack[] {
 	return guRepoStacks.filter((stack) => {
-		const guRepoName: string = stack.guRepoName;
-		return reposWithoutProductionTopic.includes(guRepoName);
+		const guRepoName = stack.guRepoName;
+		return !!guRepoName && reposWithoutProductionTopic.includes(guRepoName);
```
NovemberTang, line 56: The changes to this file reflect that the type of guRepoName has changed from `string` to `string | undefined`

packages/common/src/types.ts
```diff
@@ -26,5 +26,5 @@ export interface AWSCloudformationStack {
 }
 
 export interface GuRepoStack extends AWSCloudformationStack {
-	guRepoName: string;
+	guRepoName: string | undefined;
```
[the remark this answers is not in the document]
NovemberTang: Ok these types have been combined now, and guRepoName is optional

packages/repocop/src/index.ts
```diff
@@ -27,6 +31,37 @@ async function writeEvaluationTable(
...
+		archived: repo.archived,
+	}));
+
+	const stacks: AWSCloudformationStack[] = (await getStacks(prisma))
+		.map((s) => parseTagsFromStack(s))
+		.filter((s) => !(s.tags['Stack'] === 'playground')); //ignore playground stacks for now.
+
+	//TODO exclude repos where we have found a match for unarchived repos.
+	const archivedRepos = allRepos.filter((repo) => repo.archived);
+	const unarchivedRepos = allRepos.filter((repo) => !repo.archived);
```
akash1810, line 52: IIUC we were previously [performing these filters in SQL](https://github.com/guardian/service-catalogue/blob/0d4bdd43cd87a0b16daee7752ec55d0d685bee2a/packages/repocop/src/query.ts#L92-L133). Does doing it in code impact performance?

packages/repocop/src/rules/repository.ts
```diff
@@ -108,6 +109,43 @@ function isMaintained(repo: github_repositories): boolean {
 	return isInteractive || recentlyUpdated;
 }
 
+//TODO find somehwere else to put this
+export interface RepoAndStatus {
+	full_name: string | null;
+	name: string | null;
+	archived: boolean | null;
+}
```
akash1810, line 117: Might it be easier for our interfaces to always be the happy path. That is:

```ts
interface RepositoryArchivedStatus {
  fullName: string;
  name: string;
  archived: boolean;
}
```

We'd need to convert from the Prisma(?) type to this (once), but from then on the logic becomes a lot simpler as we're not having to handle the `null` cases.

packages/repocop/src/types.ts
```diff
@@ -0,0 +1,4 @@
+interface RepoAndStack {
```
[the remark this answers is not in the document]
akash1810: I'd suggest the linter is correctly complaining here. Could we add once we need it?

packages/repocop/src/index.ts
NovemberTang: Reducing the number of DB calls wasn't really a change motivated by performance, it was more about lowering the complexity and number of raw SQL queries we're running, and keeping as much filtering/transformation as possible in the code itself, reducing the number of side effecting calls. I think whether to optimise for performance, cost, readability, testability, or FP-ness are becoming opposing concerns as repocop develops, and we should probably agree on a common approach at some point.

packages/repocop/src/types.ts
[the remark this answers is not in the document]
NovemberTang: this is being used now! it's just not explicitly exported anywhere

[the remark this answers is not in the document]
akash1810: I'm a little surprised TypeScript is happily annotating a variable as this type in another file w/out it being exported 🤔.

packages/repocop/src/index.ts
```diff
@@ -45,6 +92,11 @@ export async function main() {
 		`Found ${unmaintinedReposCount} unmaintained repositories of ${unarchivedRepositories.length}.`,
 	);
 
+	console.log(
+		'Archived repos with potentially active stacks:',
+		(await findArchivedReposWithStacks(prisma, config)).slice(0, 10),
```
akash1810, line 97: Why `slice` here?

packages/repocop/src/rules/repository.ts
```diff
@@ -108,6 +109,38 @@ function isMaintained(repo: github_repositories): boolean {
...
+ */
+export function findStacks(
+	repo: RepoAndArchiveStatus,
+	stacks: AWSCloudformationStack[],
+): RepoAndStack {
+	const stackMatches = stacks.filter((stack) => {
+		return (
+			!!stack.stackName &&
+			(stack.guRepoName === repo.full_name ||
+				stack.stackName.includes(repo.name))
```
akash1810, line 131: Interested to know how many CloudFormation stacks this `stackName.includes` matches 🤔.

packages/repocop/src/rules/repository.ts
```diff
@@ -108,6 +109,38 @@ function isMaintained(repo: github_repositories): boolean {
 	return isInteractive || recentlyUpdated;
 }
 
+//TODO find somehwere else to put this
+export interface RepoAndArchiveStatus {
+	full_name: string;
```
akash1810, line 114: minor: Could we use camelCase for our interfaces/types for consistency?

akash1810 (commented):
Couple of minor comments, nothing blocking.

packages/repocop/src/index.ts
akash1810: Worth creating a contributing guide once we agree on the guidelines(?) for how to use Prisma to query the data?

packages/repocop/src/index.ts
NovemberTang: We're not actually doing anything with this information, so there's no need to log more than a subset of the data for sense-checking.

packages/repocop/src/rules/repository.ts
NovemberTang: Depends on the repo. Usually it's one or two. If it's caught false positives, it could be up to 10. The [kee](https://github.com/guardian/kee) repo is one example of a repo that catches a lot of stacks it probably shouldn't, as it's a substring in words like `keeper`

akash1810: Discussed offline, and the logic makes sense with real examples.

akash1810 (approved):

PR #553, opened 2023-12-05
refactor: Move database migrations from `repocop` to `cli`

akash1810:
> **Note**
> Easiest to review [commit by commit](https://github.com/guardian/service-catalogue/pull/553/commits).

Enforces a stronger separation of concerns by removing the database migration process out of RepoCop and into the CLI.

This change also allows us to replace (difficult) bash with (simpler, and type-safe) TypeScript, paving the way to automate the migrations.

To migrate in DEV we now run `npm -w cli start migrate -- --stage DEV`.

Migration to CODE or PROD has an additional step for safety/friction. `npm -w cli start migrate -- --stage CODE` will perform a [status](https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-status) check, allowing one to check how many migrations are about to be applied. Once happy, the `--confirm` flag can be added.

## How has it been verified?
I've successfully used the CLI to perform a migration locally, and on CODE.


packages/cli/src/aws.ts
```diff
@@ -19,6 +26,65 @@ export const getSsmClient = () => {
...
+	const listRequest = new ListSecretsCommand({
+		Filters: [
+			{
+				Key: 'description',
+
+				// hopefully AWS CDK doesn't change the format of this string!
+				Values: [
+					`Generated by the CDK for stack: deploy-${stage}-service-catalogue`,
+				],
+			},
```
akash1810, line 46: This is a sneaky alternative to the more typical tag based look up. A Stack, Stage, App tag filter doesn't capture a single value, so we'd then have to filter by secret name, which gets complicated. See the existing [bash script](https://github.com/guardian/service-catalogue/blob/0d4bdd43cd87a0b16daee7752ec55d0d685bee2a/packages/repocop/script/migrate.sh#L27).

packages/cli/src/database.ts
```diff
@@ -0,0 +1,83 @@
...
+	const connectionString = getDatabaseConnectionString(dbConfig);
+
+	console.log('Setting DATABASE_URL');
+	process.env.DATABASE_URL = connectionString;
+
+	console.log(`Running prisma migrate reset --force`);
+	const { stdout } = await $`npx -w common prisma migrate reset --force`;
+	console.log(stdout);
+
+	return Promise.resolve(0);
```
akash1810, line 27: `0` here represents the exit code, i.e. success.

packages/cli/src/database.ts
```diff
@@ -0,0 +1,83 @@
...
+	);
+	config({ path: `../../.env` });
+	const dbConfig = await getDevDatabaseConfig();
+	const connectionString = getDatabaseConnectionString(dbConfig);
+
+	console.log('Setting DATABASE_URL');
+	process.env.DATABASE_URL = connectionString;
+
+	console.log(`Running prisma migrate reset --force`);
+	const { stdout } = await $`npx -w common prisma migrate reset --force`;
```
akash1810, line 24: Unfortunately there isn't a programatic way to perform a migration - https://github.com/prisma/prisma/issues/4703. We're using [`execa`](https://github.com/sindresorhus/execa) here, as it offers a slightly easier API than [Node's `child_process`](https://nodejs.org/api/child_process.html).

packages/cli/src/database.ts
```diff
@@ -0,0 +1,83 @@
...
+	return Promise.resolve(0);
+}
+
+async function isConnectedToVpn(): Promise<boolean> {
+	console.log('Checking if connected to VPN');
+
+	const { stdout } = await $`ifconfig`;
+	const connected = stdout.includes('10.249.');
+	return Promise.resolve(connected);
+}
```
akash1810, line 83: Inspiration taken from [teamcity-scripts](https://github.com/guardian/teamcity-scripts/blob/8cbab93a369d6565c1c818e648d9a999f2e4d920/scripts/setup.sh#L7-L15).

NovemberTang: Is it safe to make this range publicly available?

akash1810: Yes. The `10.0.0.0` range represents the private network - https://en.wikipedia.org/wiki/Reserved_IP_addresses. Whilst `10.249.X.X` might resolve for others, it'll be resolve on their network, not ours.

We're using it to determine VPN connectivity as our AWS account has [direct connect](https://aws.amazon.com/directconnect/) setup, essentially making the office and AWS appear as one network.

NovemberTang (approved):
fancy!

PR #556, opened 2023-12-05
ci: Check Prisma schema file matches migrations

akash1810:
## What does this change?
Check, in CI, that the `schema.prisma` file lines up with the database schema as defined by the migrations.

## Why?
The `schema.prisma` file is used to generate the prisma client, which is then used in our TypeScript applications to interact with the database. If the `schema.prisma` file does not line up with the database schema, then the queries might could be invalid, and will only be known at runtime.

## How has it been verified?
TBD.

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -41,13 +41,19 @@ verifyMarkdown() {
     echo "Best practices markdown file is out of date. Please regenerate the project and commit the changes."
     exit 1
   fi
+}
 
+checkPrismaSchema() {
+  npm -w cli start migrate -- --stage DEV
+  git diff --exit-code packages/common/prisma/schema.prisma
 }
 
 npm ci
 npm run typecheck & npm run lint
 npm run test
 npm run synth & npm run build
+
+checkPrismaSchema
 
 verifyMarkdown & \
   createZip "interactive-monitor" & \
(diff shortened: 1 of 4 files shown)


scripts/ci.sh
```diff
@@ -41,14 +41,20 @@ verifyMarkdown() {
...
+  npm -w cli start migrate -- --stage DEV
+  git diff --exit-code packages/common/prisma/schema.prisma
 }
 
 npm ci
 npm run typecheck & npm run lint
 npm run test
 npm run synth & npm run build
 
+checkPrismaSchema
```
AshCorr, line 56: I'm wondering how much time this is going to add to your CI... Might be worth creating a seperate workflow or job to test the migration just so it can run in parallel with the rest of our CI!

AshCorr (approved):
Nice one! I was expecting this to be quite tricky!

Just a comment about CI times, might be worth splitting this off into a separate job.

scripts/ci.sh
akash1810: Yes, I think this is worth being it's own job. I'm not sure if we're able to (or even want to!) make it so we only upload to Riff-Raff once everything is happy, or if we only require the integration test to pass for merging?

akash1810: I've split this out now. Looks like this change totals 33 seconds: 22 seconds to start the Postgres container, 11 seconds to run the tests.

image

PR #557, opened 2023-12-05
chore(repocop): less logging

NovemberTang:
## What does this change?

Reduce the logs produced by the production monitor.

## Why?

We are no longer testing this feature, so logs don't need to be as verbose

### DIFF
--- a/packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
+++ b/packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
@@ -70,9 +70,6 @@ async function findReposInProdWithoutProductionTopic(
 
 	const repoNamesWithoutProductionTopic: string[] =
 		getRepoNamesWithoutProductionTopic(unarchivedRepos);
-	console.log(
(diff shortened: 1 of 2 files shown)


packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -163,15 +145,7 @@ export async function applyProductionTopicAndMessageTeams(
...
 		.filter((contactableRepo) => contactableRepo.teamNameSlugs.length > 0);
 
-	console.log(
-		`Found ${reposWithContactableOwners.length} repos with contactable owners.`,
-	);
-	reposWithContactableOwners.map((repo) => console.log(repo));
-
 	if (config.stage === 'PROD') {
-		console.log(
-			`Applying production topic to ${reposWithContactableOwners.length} repos and messaging their teams.`,
```
NovemberTang, line 173: We already log each individual repo name when we apply the topics, and it's **overwhelmingly** unlikely that we'll see more than one or two of these pop up on any given day

akash1810 (approved):

PR #558, opened 2023-12-06
allow a grace period for new production stacks

NovemberTang:
## What does this change?

Don't reapply to stacks until they are three months old

## Why?

Since [actions-riff-raff/86](https://github.com/guardian/actions-riff-raff/pull/86), it has been acceptable to deploy a project if the user has applied a topic of `production`, `hackday`, `prototype` or `learning`. DevX coming in the next day and relabelling the project as `production` is a confusing user experience. We should allow a grace period for projects to mature before relabelling them.


packages/repocop/src/remediations/repository-06-topic-monitor-production.ts
```diff
@@ -24,7 +24,12 @@ async function notifyOneTeam(
 	const client = new Anghammarad();
 	await client.notify({
 		subject: `Production topic monitoring (for GitHub team ${teamSlug})`,
-		message: `The 'production' topic has applied to ${fullRepoName} as it appears to have a PROD or INFRA stack in AWS. Repositories should have one of the following topics, to help understand what is in production: 'production', 'testing', 'documentation', 'hackday', 'prototype', 'learning', 'interactive'. Visit the links below to learn more about topics and how to add/remove them if you need to.`,
+		message:
+			`The 'production' topic has applied to ${fullRepoName}.` +
+			' This is because it has been associated with a live PROD or INFRA stack that is over three months old.' +
```
NovemberTang, line 29: This is the line that has actually changed, everything else is just formatting.

akash1810: Are we moving to concatinated strings because the line is wide? Getting the spacing correct is [tricky](https://github.com/guardian/service-catalogue/pull/542).

Could we enable soft wrapping in the IDE to avoid changing swapping between the two forms?

Alternatively, we could use [tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) to mimic Scala's `stripMargin`[^1]:

```ts
  /**
   * A helper function to make multi-line strings easier to write.
   *
   * Example usage:
   * ```typescript
   * const message = stripMargin`
   *   |mkdir /etc/gu
   *   |cat > /etc/gu/my-application.conf <<-'EOF'
   *   |  include "application"
   *   |EOF
   *   |`;
   * ```
   *
   * @param template a template string
   * @param args extra arguments
   */
  export const stripMargin = (template: TemplateStringsArray, ...args: unknown[]): string => {
    const result = template.reduce((acc, part, i) => [acc, args[i - 1], part].join(""));
    return result.replace(/\r?(\n)\s*\|/g, "$1");
  };
```

[TS Playground demo here](https://www.typescriptlang.org/play?#code/MYewdgzgLgBNBOBLADgWQIbwOaLDAvDABRQCmAtsgDbpkBcMAKhdbaQMpRJhYQCC8eOgCeAGhgA6KZl4MArmADWYEAHcwAbQC6ASgYJcWAgD4YAbwBQMGKEix4pCHKqxCZSjTISHAEznBSIiJ0YGBxZEwocUQdExgNELCYGQgNRBgAWhgARi1wyK0JACsQXCIAInKdHQBuKxgHKDl4PAcnF29SVgCiAHoAHXgAfiJ+sB1+iAAqfoAfXqxxcoASbKq6gF86i1toGHJHCHQsUgI4LhQMbFwAA3rZgAlSKioQGB9STAbPj-hRe-uAElYIgIDZmg4wFAqMIYMszGBSKoYAARNhEHQbCQA6yzPhgHw5ADU2TO8OyJI29zuO3AEBAVFIEleWCIBwgRxOtSAA).

[^1] I've been trying to find a place for this for a while https://github.com/guardian/cdk/pull/298 😅.

akash1810 (approved):

PR #559, opened 2023-12-06
docs: How to create a Prisma migration

akash1810:
## What does this change?
Document how to create a Prisma migration. This was lost in #553.

## How has it been verified?
Ran `npx -w common prisma migrate dev --create-only --name test` and observed a new `migration.sql` file being created.


packages/common/prisma/.env
```diff
@@ -0,0 +1 @@
+../../../.env
```
akash1810, line 1: The `schema.prisma` says the datasource is stored in the environment variable `DATABASE_URL`.

Prisma supports `.env` files, and will look in the following places:
- `./.env`
- `./prisma`

Therefore, symlink the root `.env` file to this location too.

See https://www.prisma.io/docs/guides/development-environment/environment-variables/env-files.

AshCorr (approved):
I've been creating migration files manually... oops!

PR #560, opened 2023-12-06
Add a new schema which holds tables and functions dedicated to monitoring cloudquery status

AshCorr:
## What does this change?

Drops the old (well, older) cq_sync_time() function and replace it with a new function which is aware of exactly which tables to check and what their frequencies should be.

## Why?

The list of tables and their frequencies is current spread out across a few different dashboards and alerts in Grafana and its a bit of a pain in the arse to make sure they're all using the same query.

Part of this PR creates a new `monitoring.cloudquery_table_frequency` table, it has hardcoded values at the moment, but I'm hoping that at some point we'll be able to populate this table at runtime when a task starts. Maybe by running a little script at the start of a task which takes the `tables` config and appends it to this table.

## How has it been verified?

Ran on CODE

image

image

### DIFF
--- a/packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
+++ b/packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
@@ -83,7 +83,7 @@ BEGIN
         -- Unfortunately we rely on rows existing in a table in order to know when it was last synced
         -- in a future iteration we may want to consider adding a rule to require that some tables have data
         -- to be considered "in sync"
-        cloudquery_table_status.in_sync := coalesce(last_sync, NOW()) >=
+        cloudquery_table_status.in_sync := COALESCE(last_sync, NOW()) >=
             CASE
                 WHEN table_to_check.frequency = 'DAILY'  THEN NOW() - INTERVAL '36' HOUR
                 WHEN table_to_check.frequency = 'WEEKLY' THEN NOW() - INTERVAL '8'  DAY
--- a/packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
+++ b/packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
@@ -1,16 +1,14 @@
-CREATE SCHEMA IF NOT EXISTS monitoring;
-
 -- This migration creates a bit more of a robust solution of this function.
-DROP FUNCTION public.cq_last_synced();
+DROP FUNCTION cq_last_synced();
 
 -- List of all tables to monitor for out of sync issues
 -- Ideally we could generate this list automatically, but for now its hardcoded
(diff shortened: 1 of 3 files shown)


packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
```diff
@@ -0,0 +1,94 @@
+CREATE SCHEMA IF NOT EXISTS monitoring;
```
akash1810, line 1: Does Grafana have enough permission to read from this schema?

packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
```diff
@@ -0,0 +1,94 @@
...
+DROP FUNCTION public.cq_last_synced();
+
+-- List of all tables to monitor for out of sync issues
+-- Ideally we could generate this list automatically, but for now its hardcoded
+DROP TABLE IF EXISTS  monitoring.cloudquery_table_frequency;
+CREATE TABLE monitoring.cloudquery_table_frequency (
+     table_name TEXT,
+     frequency TEXT
+);
+INSERT INTO monitoring.cloudquery_table_frequency(table_name, frequency) VALUES
```
akash1810, line 13: Would we need to keep this in-sync with the schedules defined in CDK? If so, worth adding a comment somewhere in the CDK part of the repo?

akash1810 (approved):
Couple of minor comments added.

packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
AshCorr: Already got a follow up PR to keep it in sync 😉 but I'll add a comment for now!

packages/common/prisma/migrations/20231205105200_monitoring/migration.sql
AshCorr: good question 🤔 

How do you feel about having separate schemas? I'm wondering if perhaps repocop should have a different schema too so that its tables aren't mixed in with the cloudquery ones, the main benefit is mostly just organization and separation of concerns, but perhaps its not worth the extra complexity...

akash1810: Sounds sensible. I think the deciding factor might be whether Prisma supports multiple schemas or not.

AshCorr: prisma only supports multiple schemas in preview 😭

AshCorr: also doesn't appear to support functions, if we ever want to use this table in repocop we'll need to create a view that calls the function.

PR #563, opened 2023-12-06
Make DB calls up front

NovemberTang:
## What does this change?

Make as many DB calls as possible up front

## Why?

See the README change in this PR for a full explanation

## How has it been verified?

TODO: test on CODE environment


packages/repocop/src/index.ts
```diff
@@ -34,33 +39,28 @@ async function writeEvaluationTable(
 function toRepoAndArchiveStatus(
 	repo: github_repositories,
 ): RepoAndArchiveStatus | undefined {
-	if (!repo.archived || !repo.name || !repo.full_name) {
+	if ([repo.archived, repo.name, repo.full_name].includes(null)) {
```
NovemberTang, line 42: non-null assertions are mitigated here

packages/repocop/src/index.ts
```diff
@@ -34,33 +39,28 @@ async function writeEvaluationTable(
...
 		return undefined;
 	} else {
 		return {
-			archived: repo.archived,
-			name: repo.name,
-			fullName: repo.full_name,
+			archived: repo.archived!,
```
tjsilver, line 46: Will the next PR deal with these forbidden non-null assertions?

NovemberTang: yes, promise!

packages/repocop/src/index.ts
```diff
@@ -34,33 +39,28 @@ async function writeEvaluationTable(
...
-
-	const stacks = (await getStacks(prisma))
-		.map((s) => parseTagsFromStack(s))
-		.filter((s) => !(s.tags['Stack'] !== 'playground')); //ignore playground stacks for now.
-
-	const archivedRepos = allRepos.filter((repo) => repo.archived);
-	const unarchivedRepos = allRepos.filter((repo) => !repo.archived);
+	const archivedRepos: RepoAndArchiveStatus[] = archivedRepositories
+		.map((repo) => toRepoAndArchiveStatus(repo))
+		.filter((r) => r === undefined) as RepoAndArchiveStatus[];
```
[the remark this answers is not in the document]
NovemberTang: oo, thanks, I'll take a look at this!

packages/repocop/src/index.ts
AshCorr: Not quite sure I understand, why use `.includes` instead of `||`? Aren't you sacrificing Typescripts ability to recognize that these fields aren't null this way?

PR #564, opened 2023-12-06
feat(data-audit): Save results to the database

akash1810:
## What does this change, and why?
Records the results of a data audit to the new database table `audit_results`. This'll allow us to add the results on a "Service Catalogue health" dashboard.

## How has it been verified?
Ran data-audit locally, and witnessed the table being written to.

## Rollout
This change requires a migration to be applied via:

```bash
npm -w cli start migrate -- --stage [CODE|PROD]
```

PR #566, opened 2023-12-06
shrink the repo type, and make always-there fields required

NovemberTang:
## What does this change?

Creates a new type, `Repository`, which is a subset of `github_repositories`, and stops fields being nullable if they are never actually null in reality

## Why?

- Doing the conversion from `T | undefined => T` at the query stage means much less null checking later on
- Using a type that's smaller than the original makes it clear which fields we actually care about, and reduces memory usage in the lambda.

## How has it been verified?

Deployed to CODE, running the lambda produces the expected results, consistent with previous outputs

### DIFF
--- a/packages/repocop/src/types.ts
+++ b/packages/repocop/src/types.ts
@@ -1,18 +1,16 @@
-import type { github_repositories } from '@prisma/client';
-
 export interface RepoAndStack {
 	fullName: string;
 	stacks: string[];
 }
 
-type RequiredRepositoryFields = Pick<
-	NonNullable<github_repositories>,
-	'archived' | 'created_at' | 'full_name' | 'id' | 'name' | 'topics'
->;
-
-type OptionalRepositoryFields = Pick<
-	github_repositories,
-	'updated_at' | 'pushed_at' | 'default_branch'
->;
-
-export type Repository = RequiredRepositoryFields & OptionalRepositoryFields;
+export interface Repository {
+	archived: boolean;
+	name: string;
+	full_name: string;
+	topics: string[];
+	updated_at: Date | null;
+	pushed_at: Date | null;
+	created_at: Date;
+	id: bigint;
+	default_branch: string | null;
+}
--- a/packages/repocop/src/types.ts
+++ b/packages/repocop/src/types.ts
@@ -1,16 +1,18 @@
+import type { github_repositories } from '@prisma/client';
+
 export interface RepoAndStack {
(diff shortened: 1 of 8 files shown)


packages/repocop/src/types.ts
```diff
@@ -3,8 +3,14 @@ export interface RepoAndStack {
 	stacks: string[];
 }
 
-export interface RepoAndArchiveStatus {
-	fullName: string;
-	name: string;
+export interface Repository {
```
akash1810, line 6: Did we consider extending the Prisma type, [`Pick`ing](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys), or [`Omit`ing](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys), or making fields [`Required`](https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype)?

e.g. https://github.com/guardian/service-catalogue/pull/564/files#diff-c75078d4fa0a95f4b071f267146f8ce65aa0cf6f28c455ea0204198e59d68018R3.

NovemberTang: TIL!

NovemberTang: We (me and @akash1810 ) could not get this to work, so reverting, and will try again later)

akash1810 (approved):

packages/repocop/src/types.ts
akash1810: I think I've got it - https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgILIN4ChnIPYgQBcyAzmFKAOYDcOyYA7nieZSFcgD7IgCuAGwFYAvliyhIsRCgBCyCAA9IIACak0mekxbIAcgT2CBcAEYCIAHlQBtAEQ67AXQB8o8QgLlkVPHlUk8gC8WrgExMgA5HCkqpEANNrMJNGxMXHuWJ4g3qZwAcjBofiEKekJSbr8QqJAA.

Something like this should do the trick:

```ts
interface Repository extends github_repositories {
  archived: NonNullable<github_repositories["archived"]>,
  created_at: NonNullable<github_repositories["created_at"]>,
  full_name: NonNullable<github_repositories["full_name"]>,
  id: NonNullable<github_repositories["id"]>,
  name: NonNullable<github_repositories["name"]>,
  topics: NonNullable<github_repositories["topics"]>,
}
```

PR #567, opened 2023-12-07
feat(data-audit): Audit AWS Lambda functions

akash1810:
## What does this change?
Add an audit for AWS Lambda functions. This is the first multi-region audit 🎉.

## Why?
[See the original ADR](https://github.com/guardian/service-catalogue/blob/main/ADR/03-data-accuracy.md).

## How has it been verified?
Ran locally, and on CODE.

### DIFF
--- a/packages/data-audit/src/audit/aws-lambda.ts
+++ b/packages/data-audit/src/audit/aws-lambda.ts
@@ -1,0 +1,92 @@
+import type { FunctionConfiguration } from '@aws-sdk/client-lambda';
+import { LambdaClient, paginateListFunctions } from '@aws-sdk/client-lambda';
+import type { Account } from '@aws-sdk/client-organizations';
+import type { PrismaClient } from '@prisma/client';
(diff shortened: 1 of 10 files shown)


packages/data-audit/src/audit/aws-lambda.ts
```diff
@@ -0,0 +1,92 @@
...
+	const functions: FunctionConfiguration[] = [];
+
+	for await (const page of paginateListFunctions(
+		{ client, pageSize: 50 },
+		{},
+	)) {
+		functions.push(...(page.Functions ?? []));
+
+		// (attempt to) avoid rate limiting by sleeping for 3 seconds between pages
+		await sleep(3000);
```
akash1810, line 39: The list lambda function endpoint is rate limited. This is a slightly hacky way to avoid the rate limit. We might want to monitor how this holds up over time!

tjsilver (approved):
👍

PR #569, opened 2023-12-07
Only retrieve stacks once, don't prefilter

NovemberTang:
## What does this change?

Instead of querying for non-prod stacks directly from the Cloudquery databse, we perform that filtering as part of the lambda, in the `isProdStack()` function.

At some point we could simplify this further by deriving the `AWSCloudformationStack` type from `aws_cloudformation_stacks`, but that's beyond the scope of this PR.

## Why?

- We can test the filtering logic
- `findReposInProdWithoutProductionTopic` is now completely unit testable, as we pass the list of stacks in, rather than making an inline DB call
- subfunctions used by `findReposInProdWithoutProductionTopic` no longer need to be exported, as we can test their behaviour indirectly as part of the parent function's tests.
- The number of custom queries we need to maintain is reduced
- The number of calls out to the DB is reduced

Writing these tests exposed a bug in our production monitoring logic, where guRepoName was always resolving to `undefined`. This has now been fixed.

## How has it been verified?

All old and new tests pass, behaves as expected on CODE


packages/repocop/src/remediations/shared-utilities.ts
```diff
@@ -95,21 +87,6 @@ export async function addMessagesToQueue(
 	}
 }
 
-export function getGuRepoName(tag: AWSCloudformationTag): string | undefined {
```
NovemberTang, line 98: moved to `query`

packages/repocop/src/remediations/topics/topic-monitor-production.ts
```diff
@@ -66,43 +62,48 @@ export function getReposInProdWithoutProductionTopic(
 	});
 }
 
-async function findReposInProdWithoutProductionTopic(
-	prisma: PrismaClient,
+function isProdStack(stack: AWSCloudformationStack) {
+	return (
+		!!stack.tags.Stage &&
+		(stack.tags.Stage === 'PROD' || stack.tags.Stage === 'INFRA') &&
+		stack.tags.Stack !== 'playground' && // Ignore playground stacks
```
NovemberTang, line 69: we do filter out playground stacks at an earlier stage, but I think it's best to be explicit here. Also, you never know when that earlier filter will be removed so we can use all the stacks somewhere else.

tjsilver (approved):
Looks good!

PR #570, opened 2023-12-07
feat: log out Anghammarad  message if not in PROD

tjsilver:
## What does this change?

- Pulls the creation of the Anghammarad message for `topic-monitor'-production` into a separate function.
- Console logs out the contents of the message.
- Moves the logic to perform the topic addition and messaging according to the stage closer to their calls, to allow for easier logging.

## Why?

This will allow us to preview the messaging on CODE so we can see if it is formatted correctly. 

## How has it been verified?

Tested on CODE.

### DIFF
--- a/packages/repocop/src/remediations/topics/topic-monitor-production.ts
+++ b/packages/repocop/src/remediations/topics/topic-monitor-production.ts
@@ -12,7 +12,8 @@ import { getRepoOwnership, getTeams } from '../../query';
 import type { Repository } from '../../types';
 import { findContactableOwners, removeRepoOwner } from '../shared-utilities';
 
-interface AnghammaradMessageText {
+const MONTHS = 3;
+interface AnghammaradTextFields {
 	subject: string;
 	message: string;
 }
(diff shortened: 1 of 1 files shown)


packages/repocop/src/remediations/topics/topic-monitor-production.ts
```diff
@@ -16,21 +16,37 @@ import {
 	removeRepoOwner,
 } from '../shared-utilities';
 
-async function notifyOneTeam(
+interface AnghammaradMessageText {
```
NovemberTang, line 15: could we rename this `AnghammaradMessage`? the `Text` implies it just contains the body

packages/repocop/src/remediations/topics/topic-monitor-production.ts
```diff
@@ -123,9 +139,21 @@ async function applyProductionTopicToOneRepoAndMessageTeams(
 	const owner = 'guardian';
 	const topic = 'production';
 	const shortRepoName = removeRepoOwner(fullRepoName);
-	await applyTopics(shortRepoName, owner, octokit, topic);
+	const { stage } = config;
+	if (stage === 'PROD') {
+		await applyTopics(shortRepoName, owner, octokit, topic);
+	} else {
+		console.log(
+			`Would have applied the ${topic} topic to ${shortRepoName} with stack ${stackName} owned by ${owner}`,
```
NovemberTang, line 147: The naming of `owner` means its ambiguous if we are talking about the owner of the repo or the owner of the stack

packages/repocop/src/remediations/topics/topic-monitor-production.ts
tjsilver: It is the text part of the message - the subject and message text, but there are other bits to Anghammarad messages so I thought this would be clearer. But can rename!

NovemberTang (approved):

PR #571, opened 2023-12-07
Derive AwsCloudFormationStack from aws_clouformation_stacks

NovemberTang:
## What does this change?

Simplify the custom cfn stack type so it can be inferred directly from the prisma type, 

## Why?

- Fewer transformations/transformation tests (aka less code)
- The new type is closer in shape to the original

## How has it been verified?

Deployed to CODE, verified output is the same as it was before the deploy.


tjsilver (approved):
👍

PR #572, opened 2023-12-08
Demonstrate we are able to detect if a repo has sufficient vulnerability tracking.

NovemberTang:
## What does this change?

Creates a method that

1. Checks if vulnerability tracking is required for a repo
2. Retrieves the languages used by that repo
3. Looks for the repo on Snyk
3a. Checks that all languages used by the repo are supported by Snyk
4. If not on Snyk, checks that all languages used by the repo are supported by dependabot.

## Why?

We expect all production repositories to have dependency tracking enabled, either via Snyk or Dependabot. This feature allows us to track if teams are complying with this requirement. Teams that are not using more niche languages such as Scala, do not 

## How has it been verified?

All new and old tests pass. Deployed to CODE, logs look happy.

### DIFF
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -10,12 +10,17 @@ import {
 	getRepositories,
 	getRepositoryBranches,
 	getRepositoryTeams,
+	getSnykProjects,
 	getStacks,
 } from './query';
 import { protectBranches } from './remediations/branch-protector/branch-protection';
 import { sendPotentialInteractives } from './remediations/topics/topic-monitor-interactive';
 import { applyProductionTopicAndMessageTeams } from './remediations/topics/topic-monitor-production';
-import { evaluateRepositories, findStacks } from './rules/repository';
+import {
+	evaluateRepositories,
+	findStacks,
+	isTracked,
+} from './rules/repository';
 import type { AwsCloudFormationStack, RepoAndStack, Repository } from './types';
 
 async function writeEvaluationTable(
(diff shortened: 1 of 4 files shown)


packages/repocop/src/index.ts
```diff
@@ -93,10 +98,16 @@ export async function main() {
 		archivedWithStacks.slice(0, 10),
 	);
 
+	const octokit = await stageAwareOctokit(config.stage);
+	unarchivedRepos
+		.filter((r) => r.topics.includes('production'))
+		.slice(0, 10)
+		.map((r) => isTracked(octokit, r, snykProjects));
```
NovemberTang, line 105: this is the line that does the thing.

packages/repocop/src/rules/repository.ts
```diff
@@ -104,6 +106,129 @@ function isMaintained(repo: Repository): boolean {
...
+			!!repo.full_name &&
+			tags.repo == repo.full_name &&
+			tags.branch === repo.default_branch,
+	);
+
+	const repoIsOnSnyk = !!matchingSnykProject;
+
+	if (repoIsOnSnyk) {
+		const containsOnlySnykSupportedLanguages = languages.every((language) =>
+			supportedSnykLanguages.includes(language),
```
tjsilver, line 199: How does this work? It looks as though `supportedSnykLanguages` contains the dependabot and untested languages too?

NovemberTang: Yes, that's right. Snyk supports a superset of Dependabot supported languages.

tjsilver (approved):

PR #573, opened 2023-12-08
refactor: remove unused shared functions

tjsilver:
## What does this change?

Removes unused functions from the shared utilities file in repocop.

## Why?

They are now redundant.

## How has it been verified?
Ran tests locally, will test on CODE.


JuliaBrigitte (approved):

NovemberTang (approved):
😍

PR #574, opened 2023-12-08
Move functions out of index. Create a parent function for experimental features

NovemberTang:
## What does this change?

There's a lot going on in the main file. This PR moves a lot of it out to the `repository` file, and creates a function for evaluating experimental repocop features

## Why?

Readability. The main method is less cluttered, and its easy to tell which features are and aren't experimental


tjsilver (approved):
Good idea to keep the index file as clear as possible.

PR #576, opened 2023-12-11
Only call getTeams() once

NovemberTang:
## What does this change?

Simplify the `github_teams` type to:

- Only keep the fields we need (reduce memory usage)
- Make those fields required (less null checking, more accurate reflection of the DB)

Only call the getTeams function once

## Why?

Making the DB call up front and reusing the object means fewer side-effecting calls, better testability, and a lower chance of failure overall as we are making fewer DB calls.

## How has it been verified?

- [x] Deploy to CODE
- [x] Verify execution succeeds
- [x] Verify no teams are being lost in the conversion to `Team`


tjsilver (approved):
Looks good!

PR #584, opened 2023-12-14
Ditch branch protector temporarily

NovemberTang:
## What does this change?

Temporarily stop branch protector from running until we have a solution that allows us to publish libraries

## Why?

It's interfering with library releases.

### DIFF
--- a/packages/repocop/src/config.ts
+++ b/packages/repocop/src/config.ts
@@ -72,6 +72,6 @@ export async function getConfig(): Promise<Config> {
 			'guardian/pluto-', // Multimedia team
 		],
 		interactivesCount: stage === 'PROD' ? 15 : 1,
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
:(

packages/repocop/src/config.ts
```diff
@@ -67,5 +72,6 @@ export async function getConfig(): Promise<Config> {
 			'guardian/pluto-', // Multimedia team
 		],
 		interactivesCount: stage === 'PROD' ? 15 : 1,
+		branchProtectionEnabled: process.env.BRANCH_PROTECTION_ENABLED === 'true',
```
akash1810, line 75: minor: absence of this env var will disable branch protection. Worth setting this env var in CDK for explicitness?

akash1810 (approved):

PR #587, opened 2023-12-18
Add decidated CloudQuery sources for RDS and Backup

jacobwinch:
## What does this change?

This allows us to collect data about the RDS and Backup services once per day, instead of once per week.

## Why?

Similar to https://github.com/guardian/service-catalogue/pull/428.

Next quarter we have an OKR related to RDS backups and it would be nice to see daily updates for this data.

I'm also hoping that the Backup tables will be useful for ensuring that https://github.com/guardian/aws-backup [continues working as expected](https://trello.com/c/Z3Ds2ENF/2337-spike-how-can-we-ensure-that-devx-backup-continues-working). 

## How has it been verified?

As we are copying an existing pattern, I think the snapshot testing is sufficient here, but please let me know if you'd like me to check anything else.

### DIFF
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -183,6 +183,26 @@ export function addCloudqueryEcsCluster(
 			policies: [listOrgsPolicy, cloudqueryAccess('*')],
 		},
 		{
+			name: 'OrgWideRDS',
+			description:
+				'Collecting RDS data across the organisation. Uses include auditing backup configuration.',
+			schedule: nonProdSchedule ?? Schedule.cron({ minute: '0', hour: '6' }),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_rds*'],
+			}),
+			policies: [listOrgsPolicy, cloudqueryAccess('*')],
+		},
+		{
+			name: 'OrgWideBackup',
+			description:
+				'Collecting Backup data across the organisation. Uses include auditing backup configuration.',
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/cloudquery/index.ts
```diff
@@ -182,6 +182,26 @@ export function addCloudqueryEcsCluster(
 			}),
 			policies: [listOrgsPolicy, cloudqueryAccess('*')],
 		},
+		{
+			name: 'OrgWideRDS',
+			description:
+				'Collecting RDS data across the organisation. Uses include auditing backup configuration.',
+			schedule: nonProdSchedule ?? Schedule.cron({ minute: '0', hour: '6' }),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_rds*'],
```
akash1810, line 191: This wildcard will collect data for RDS events, which:

> Returns events related to DB instances, DB clusters, DB parameter groups, DB security groups, DB snapshots, DB cluster snapshots, and RDS Proxies for the past 14 days. Events specific to a particular DB instance, DB cluster, DB parameter group, DB security group, DB snapshot, DB cluster snapshot group, or RDS Proxy can be obtained by providing the name as a parameter.
> – https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeEvents.html

From https://hub.cloudquery.io/plugins/source/cloudquery/aws/v22.18.0/tables:

image

It might result in the task running for longer than we'd expect. Is this desired?

akash1810 (approved):

PR #594, opened 2023-12-19
Bump memory usage of various failing tasks to 1024MB

AshCorr:
## What does this change?

Bump `GitHubIssuesTask`, `OrgWideCloudFormation`, and `OrgWideInspectorTask`

## Why?

It looks like these 3 tasks are running out of memory

image

(Ignore the EC2, I'm not sure why thats failing just yet)


akash1810 (approved):

NovemberTang (approved):

AshCorr:
Errors from EC2 task are due to the Cloudquery task being killed seconds after its launched because of how Singleton tasks work. Fix being worked on in https://github.com/guardian/service-catalogue/pull/595

PR #595, opened 2023-12-19
Exit 1 in singleston startup script if its not the only task running

AshCorr:
## What does this change?

Instead of always exiting on a 0 code (successfull), exit on 1 if theres other running tasks.

## Why?

The singleton task sends a request to AWS to stop itself, but this happens asynchronously and the singleton task still gives the go ahead for Cloudquery to start running as it exits with a 0 exit code.

Instead exit with a 1, this way there is no race condition between ECS stopping the task in time and Cloudquery starting to run.

We now have alerting to let us know when Cloudquery exits with an error, because the Cloudquery task is allowed to start for a couple of seconds before getting killed by AWS it throws off our alerting.


packages/cdk/lib/cloudquery/ecs/task.ts
```diff
@@ -226,11 +226,8 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 						// How many more of me are there?
 						`RUNNING=$(aws ecs list-tasks --cluster $ECS_CLUSTER --family $ECS_FAMILY | jq '.taskArns | length')`,
 
-						// If there's more than one, then I don't need to be here.
-						'[[ ${RUNNING} -gt 1 ]] && aws ecs stop-task --cluster $ECS_CLUSTER --task $ECS_TASK_ARN --no-cli-pager 1> /dev/null',
-
-						// I'm the only one, let's go on.
-						'exit 0',
+						// Exit zero (successfull) if I'm the only one running
+						'exit [[ ${RUNNING} -gt 1 ]] && 1 || 0',
```
akash1810, line 230: Let's drop the [`ecs:StopTask` permission](https://github.com/guardian/service-catalogue/blob/fce5741cad7641736d790dcaffd4b94fcb4940c2/packages/cdk/lib/cloudquery/ecs/policies.ts#L63) too, to be least privilege?

akash1810: A 1 exit code implies error. Would 114 be more accurate here?

> 114	Operation already in progress
> – https://www.cyberciti.biz/faq/linux-bash-exit-status-set-exit-statusin-bash/

packages/cdk/lib/cloudquery/ecs/policies.ts
```diff
@@ -38,29 +36,15 @@ export const readBucketPolicy = (...resources: string[]): PolicyStatement => {
...
-		}),
-	];
+export function singletonPolicy(cluster: Cluster) {
+	return new PolicyStatement({
+		effect: Effect.ALLOW,
+		resources: ['*'],
+		conditions: {
+			StringEquals: {
+				'ecs:cluster': cluster.clusterArn,
+			},
```
akash1810, line 46: 😍 Do we need to test if this works? I can't remember the policy tried initially, but [previous attempts to scope this failed](https://github.com/guardian/service-catalogue/pull/484#discussion_r1397110473).

packages/cdk/lib/cloudquery/ecs/task.ts
```diff
@@ -226,11 +226,8 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 						// How many more of me are there?
 						`RUNNING=$(aws ecs list-tasks --cluster $ECS_CLUSTER --family $ECS_FAMILY | jq '.taskArns | length')`,
 
-						// If there's more than one, then I don't need to be here.
-						'[[ ${RUNNING} -gt 1 ]] && aws ecs stop-task --cluster $ECS_CLUSTER --task $ECS_TASK_ARN --no-cli-pager 1> /dev/null',
-
-						// I'm the only one, let's go on.
-						'exit 0',
+						// Exit zero (successfull) if I'm the only one running
+						'[ ${RUNNING} -gt 1 ] && exit 114 || exit 0',
```
akash1810, line 230: https://www.baeldung.com/linux/bash-single-vs-double-brackets suggests use of single brackets here won't work as expected:

image

(bash is fun...)

akash1810 (commented):
Couple of minor comments - think we'd want to test these changes on CODE?

packages/cdk/lib/cloudquery/ecs/policies.ts
AshCorr: Seems to work now! Tested in CODE!

packages/cdk/lib/cloudquery/ecs/task.ts
AshCorr: Ah right, it sounds like single brackets are available in all shell types, but in Bash you can optionally use double brackets for more features

image

Specifically in this case instead of using `-gt` we could use `>` if we used double brackets. I think that makes sense, I find `>` to be a bit more clearer than `-gt`

AshCorr: updated now!

akash1810 (approved):

PR #597, opened 2023-12-20
feat(cloudquery): Collect AWS CloudTrail data

akash1810:
We're seeing some deployments failing in Riff-Raff with the ASG reporting:

> Instance stayed in Pending state too long.

Contact AWS Support has led us to believe this is an issue in AWS's data centres.

It would be helpful to understand how many applications are experiencing this symptom. If it's the majority, maybe we need to escalate.

The ASG activity has this information. Sadly, CloudQuery doesn't have this table. I've started to build https://github.com/guardian/aws-asg-monitor to collect the information outside of Service Catalogue, however [CloudTrail](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-user-guide.html) should also have it, which CloudQuery can collect!

This change collects all CloudQuery tables. There is going to be a lot of data here, so run the task once a day, as a singleton.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -8480,6 +8480,698 @@ spec:
       },
       "Type": "AWS::Events::Rule",
     },
+    "CloudquerySourceOrgWideCloudTrailScheduledEventRule1DBF6483": {
+      "Properties": {
+        "ScheduleExpression": "rate(1 day)",
+        "State": "ENABLED",
+        "Targets": [
+          {
+            "Arn": {
+              "Fn::GetAtt": [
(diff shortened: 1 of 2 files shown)


AshCorr (approved):
Looks good!

I seem to remember this particular API is quite heavily rate limited, 1-3 TPS according to https://docs.aws.amazon.com/awscloudtrail/latest/userguide/WhatIsCloudTrail-Limits.html

The Cloudtrail tables support incremental syncs, maybe it might be time to try using them if this job takes too long?

akash1810:
> The Cloudtrail tables support incremental syncs, maybe it might be time to try using them if this job takes too long?

Yes! Will look at this in the New Year.

akash1810:
I can't find a CloudTrail event that corresponds to the below ASG activity 😢.

image

PR #602, opened 2023-12-21
Double interactive monitor capacity

NovemberTang:
## What does this change?

Increases the number of repos we check for interactives from 20 to 40

## Why?

Testing a larger subsection of repos means that we find new interactives faster.

## How has it been verified?

Manually increased the interactive count to 40 on PROD, and ran the lambda to verify that the memory/timeouts/API could handle it.

n.b. the lambda now takes 8 seconds to run, rather than 5.

### DIFF
--- a/packages/cdk/lib/repocop.ts
+++ b/packages/cdk/lib/repocop.ts
@@ -40,7 +40,7 @@ export class Repocop {
 				QUERY_LOGGING: 'false', // Set this to 'true' to enable SQL query logging
 				INTERACTIVE_MONITOR_TOPIC_ARN: interactiveMonitorTopic.topicArn,
 				GITHUB_APP_SECRET: githubAppSecret.secretArn,
-				INTERACTIVES_COUNT: guStack.stage === 'PROD' ? '20' : '3',
+				INTERACTIVES_COUNT: guStack.stage === 'PROD' ? '40' : '3',
 			},
 			vpc,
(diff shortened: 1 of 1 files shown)


akash1810:
> Testing a larger subsection of repos means that we find new interactives faster.

QQ - would it be worth adding date filters too? That is, if we get 40 repositories created since the last run, might we catch new interactives faster still?

akash1810 (approved):

NovemberTang:
> > Testing a larger subsection of repos means that we find new interactives faster.
> 
> QQ - would it be worth adding date filters too? That is, if we get 40 repositories created since the last run, might we catch new interactives faster still?

Unfortunately I think that might end up sacrificing eventual accuracy for speed.

- We haven't done a full run of this against all the unlabelled repos we have currently, so don't know how many still exist in the backlog (admittedly, I suspect there's not loads). Grabbing the 40 most recent may mean there are some older ones we may never catch.
- Most of the checks compare github repos to an S3 bucket for **deployed** interactives. If there's a considerable amount of time between repo creation and repo deployment (an example of this might be an election, where we know it's going to happen months beforehand), it may fall out of the 40 most recently created repos by the time it's deployed and we might never catch it again.

But with the additional measures the Security team plan to bring in, around labelling and archiving, I'm hoping that the list of unlabelled repos will shrink down enough that we'll start catching them faster anyway

PR #604, opened 2023-12-22
Add the ability to override a tasks plugin with a binary downloaded from a specified URL

AshCorr:
## What does this change?

Allows developers to specify a `pluginBinaryUrl` property on a CloudQuery source which specifies the URL to a plugin binary. This will prompt our ECS task to download the specified binary and use it instead of fetching the plugin from CloudQuery Hub or Github.

## Why?

When building cloudquery plugins from source, for example when making fixes upstream, it can be quite annoying to test, especially if you don't happen to have Janus permissions to the particular account that you want to run CloudQuery on.

This allows us to build a fix on our local machine, and then check it on CODE using Service Catalogues access instead of our individual permissions.

## How has it been verified?

Tested in CODE
image

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -9917,10 +9917,10 @@ spec:
             "Command": [
               "/bin/sh",
               "-c",
-              "wget -O /usr/local/share/ca-certificates/global-bundle.crt -q https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem && update-ca-certificates;printf 'kind: source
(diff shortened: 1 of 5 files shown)


adamnfish (commented):
Could we enable this only in CODE, or put it behind some kind of switch to temporarily enable it when needed? I get strong "why do we even have that lever" vibes from this feature, but understand why it will be useful!

AshCorr:
> Could we enable this only in CODE, or put it behind some kind of switch to temporarily enable it when needed? I get strong "why do we even have that lever" vibes from this feature, but understand why it will be useful!

Its not exactl a switch you can just pull, as this is all handled by CDK/Cloudformation you would still need to raise a PR to update a task definition to use the new `pluginBinaryUrl` prop and then do a deployment. Since you still have to make a branch, and do a CODE deployment, any controls I add could be commented out.

It shouldn't leave CODE unless:
  a) Someone intentionally deploys their branch to PROD which we shouldn't be doing regardless
  b) Their PR is approved and merged, in which case theres probably a valid reason why they're doing it, for example running our own fixed version of the Snyk plugin

AshCorr:
After discussion with @adamnfish we've decided to close this for now as it could encourage some bad behaviours should people start to use this flag for non debug purposes. We may take another look at this again in future should we need to debug more cloudquery plugins.

PR #610, opened 2023-12-27
chore(deps): Configure Dependabot groups

akash1810:
## What does this change?
Use [grouping](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) to bump AWS SDK libraries together, and Prisma libraries together. This should result in fewer PRs.

For example, these five open PRs would become two:
image


NovemberTang (approved):

akash1810:
https://github.com/guardian/service-catalogue/pull/615 suggests this is working 🎉.

PR #622, opened 2023-12-27
feat(cloudquery): Authenticate with CloudQuery

akash1810:
## What does this change?
Sets the `CLOUDQUERY_API_KEY` environment variable to [authenticate with CloudQuery](https://docs.cloudquery.io/docs/deployment/generate-api-key). By authenticating, we provide CloudQuery with more telemetry data, which they can use to understand our usage rates.

Being authenticated also enables us to use premium plugins, such as the latest Snyk version.

## How has it been verified?
By itself, this should be a no-op change. Paired with #623, we're able to update the Snyk plugin.

### DIFF
--- a/.env
+++ b/.env
@@ -3,7 +3,7 @@
 # Environment variables shared between ci and DEV.
 
 # See https://github.com/cloudquery/cloudquery/releases?q=cli
-CQ_CLI=3.28.0
+CQ_CLI=4.3.5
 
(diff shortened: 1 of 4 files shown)


NovemberTang (approved):

akash1810:
> By authenticating, we provide CloudQuery with more telemetry data, which they can use to understand our usage rates.

The CloudQuery Hub is reporting our Snyk plugin has ~85k rows being synced. That is, this usage reporting appears to be working.

PR #625, opened 2023-12-29
refactor(cdk): Move image references to own file

akash1810:
This is an attempt to make `task.ts` a bit smaller/easier to reason about.

## How has it been verified?
N/A. This is a no-op, as demonstrated by the CDK snapshots not changing.

### DIFF
--- a/packages/cdk/lib/cloudquery/task.ts
+++ b/packages/cdk/lib/cloudquery/task.ts
@@ -5,7 +5,6 @@ import type { ISecurityGroup } from 'aws-cdk-lib/aws-ec2';
 import type { Cluster } from 'aws-cdk-lib/aws-ecs';
 import {
 	ContainerDependencyCondition,
-	ContainerImage,
 	FargateTaskDefinition,
 	FireLensLogDriver,
 	FirelensLogRouterType,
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/cloudquery/task.ts
```diff
@@ -256,7 +243,7 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 		}
 
 		task.addFirelensLogRouter(`${id}Firelens`, {
-			image: firelensImage,
```
akash1810, line 259: Renamed `firelensImage` to `devxLogs` as it's more representative of the image location.

NovemberTang (approved):

packages/cdk/lib/cloudquery/task.ts
```diff
@@ -210,7 +197,7 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 
 		if (runAsSingleton) {
 			const singletonTask = task.addContainer(`${id}AwsCli`, {
-				image: awsImage,
```
akash1810, line 213: Renamed `awsImage` to `amazonLinux` as it's more representative of the image location.

PR #626, opened 2023-12-29
GitHub languages

NovemberTang:
## What does this change?

Adds GitHub languages as a table for local execution

## Why?

Enables security work around dependabot

## How has it been verified?

Able to collect information on repo languages locally


packages/cloudquery/dev-config/cloudquery.yaml
```diff
@@ -1,5 +1,14 @@
 ---
 kind: source
+spec:
+  name: 'github-languages'
+  path: 'guardian/github-languages'
+  registry: 'github'
+  version: v0.0.4
```
NovemberTang, line 7: Hardcoding this version here as we are only using this plugin locally, not on AWS. This will become an env var when doing the setup for AWS

akash1810 (approved):

PR #337, opened 2023-09-27
feat(cloudquery): Collect NS1 data

akash1810:
## What does this change?
Adds the (WIP) NS1 source plugin from https://github.com/guardian/cq-source-ns1 to collect NS1 data.

Currently, only zone information is returned, see https://github.com/guardian/cq-source-ns1/pull/8.

## How has it been verified?
TBD.

PR #629, opened 2023-12-29
feat: Collect DNS records from NS1

akash1810:
The NS1 source plugin (https://github.com/guardian/cq-source-ns1) is written in TypeScript, and [distributed as a Docker image](https://docs.cloudquery.io/docs/developers/creating-new-plugin/javascript-source#releasing-and-deploying-your-plugin).

Previous attempts (see https://github.com/guardian/service-catalogue/pull/337) to use this plugin failed, because ECS containers cannot interact with Docker (see https://github.com/aws/containers-roadmap/issues/1356).

In this change, we run the NS1 source plugin directly, in its own container, resulting in a GRPC server on `localhost:7777`. We then configure the CloudQuery sync container to connect to this endpoint.

Collecting this data provides more information about our running services. For example, we can identify which services use Fastly, CloudFront, or no CDN.

We can also use this data to identify "dangling DNS records". That is, records whose answer no longer exist, and thus should be deleted, saving on the limited number of records available to us.

#### Standard task
```mermaid
flowchart TD
    A[Fire Lens] --> B(CloudQuery Sync)
```

#### NS1 task
```mermaid
flowchart TD
    A[Fire Lens] --> C(CloudQuery Sync to localhost:7777)
    B[CloudQuery NS1] --> C(CloudQuery Sync to localhost:7777)
```

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/9b03a705-f1e8-4dbf-90eb-b2390dff1098), ran the task, and seen [NS1 data in Grafana CODE](https://metrics.code.dev-gutools.co.uk/goto/4I8I8qOSk?orgId=1) 🎉.

> [!NOTE]
> The NS1 API key used in CODE has been copied from https://github.com/guardian/dns-validation-lambda in CODE. This key has been limited to particular zones. We'd need to generate a new API key, with read-only access to all zones.

### DIFF
--- a/packages/cdk/lib/cloudquery/task.ts
+++ b/packages/cdk/lib/cloudquery/task.ts
@@ -2,7 +2,6 @@ import type { AppIdentity, GuStack } from '@guardian/cdk/lib/constructs/core';
 import type { GuSecurityGroup } from '@guardian/cdk/lib/constructs/ec2';
 import { Tags } from 'aws-cdk-lib';
 import type { ISecurityGroup } from 'aws-cdk-lib/aws-ec2';
-import type { Cluster } from 'aws-cdk-lib/aws-ecs';
 import {
 	ContainerDependencyCondition,
 	FargateTaskDefinition,
@@ -11,6 +10,7 @@ import {
 	LogDrivers,
 	Secret,
 } from 'aws-cdk-lib/aws-ecs';
+import type { Cluster, RepositoryImage } from 'aws-cdk-lib/aws-ecs';
 import type { ScheduledFargateTaskProps } from 'aws-cdk-lib/aws-ecs-patterns';
 import { ScheduledFargateTask } from 'aws-cdk-lib/aws-ecs-patterns';
 import type { IManagedPolicy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
@@ -101,6 +101,17 @@ export interface ScheduledCloudqueryTaskProps
 	 * @see https://cloud.cloudquery.io/teams/the-guardian/api-keys
 	 */
 	cloudQueryApiKey: Secret;
+
+	/**
+	 * The image of a CloudQuery plugin that is distributed via Docker,
+	 * i.e. plugins not written in Go.
+	 *
+	 * This image will be run on its own, exposing the GRPC server on localhost:7777.
+	 * The CloudQuery source config should be configured with a registry of grpc, and path of localhost:7777.
+	 *
+	 * @see https://docs.cloudquery.io/docs/reference/source-spec
+	 */
+	dockerDistributedPluginImage?: RepositoryImage;
 }
 
 export class ScheduledCloudqueryTask extends ScheduledFargateTask {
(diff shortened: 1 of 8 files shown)


packages/cdk/lib/cloudquery/task.ts
```diff
@@ -101,6 +101,17 @@ export interface ScheduledCloudqueryTaskProps
...
+	/**
+	 * The image of a CloudQuery plugin that is distributed via Docker,
+	 * i.e. plugins not written in Go.
+	 *
+	 * This image will be run on its own, exposing the GRPC server on localhost:7777.
+	 * The CloudQuery source config should be configured with a registry of grpc, and path of localhost:7777.
+	 *
+	 * @see https://docs.cloudquery.io/docs/reference/source-spec
+	 */
+	dockerDistributedPluginImage?: RepositoryImage;
```
AshCorr, line 114:
```suggestion
	additionalPluginImage?: RepositoryImage;
```
or
```suggestion
	pluginContainer?: RepositoryImage;
```

Think this would work with any kind of Container image right? Not just Docker images?

packages/cdk/lib/cloudquery/task.ts
```diff
@@ -201,6 +213,22 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 			logging: fireLensLogDriver,
 		});
 
+		if (dockerDistributedPluginImage) {
+			const additionalCloudQueryContainer = task.addContainer(
+				`${id}PreContainer`,
+				{
+					image: dockerDistributedPluginImage,
+					logging: fireLensLogDriver,
+					essential: false,
```
AshCorr, line 222: I think this container would be essential no? I guess the main CloudQuery container will crash either way so probably not that important.

packages/cdk/lib/cloudquery/task.ts
```diff
@@ -201,6 +213,22 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 			logging: fireLensLogDriver,
 		});
 
+		if (dockerDistributedPluginImage) {
+			const additionalCloudQueryContainer = task.addContainer(
+				`${id}PreContainer`,
```
AshCorr, line 218:
```suggestion
				`${id}PluginContainer`,
```

AshCorr (approved):
Nice 👌

packages/cdk/lib/cloudquery/task.ts
akash1810: I used `docker` here as a way to distinguish CloudQuery plugins written in Go, vs not-Go. The former is distributed as a binary, and the latter as a Docker image, see https://docs.cloudquery.io/docs/developers/creating-new-plugin/javascript-source#releasing-and-deploying-your-plugin.

packages/cdk/lib/cloudquery/task.ts
akash1810: I got an [error from this previously](https://github.com/guardian/service-catalogue/blob/f632b7db1e3c4c7fc4fa9303107cea37abaf0f19/packages/cdk/lib/cloudquery/task.ts#L232-L236)... I can't remember the exact error, I've a vague memory that CloudFormation failed to deploy the change.

mchv:
A new read-only API key has now been created and is ready to be used.

akash1810:
> In this change, we run the NS1 source plugin directly, in its own container, resulting in a GRPC server on `localhost:7777`. We then configure the CloudQuery sync container to connect to this endpoint.

Confirming this is how CloudQuery suggest doing this too! https://docs.cloudquery.io/docs/advanced-topics/using-cloud-query-docker-registry-plugins-inside-a-containerized-environment

PR #642, opened 2024-01-03
Authenticate CQ github plugin with GitHub app locally

NovemberTang:
## What does this change?

CloudQuery's github plugin now uses the CODE auth

## Why?

This is part of the work to simplify project setup. Eventually, all GitHub plugins on CODE will use GitHub app auth so the user does not have to mess around setting up PATs with the right permisisons. Next steps are:

- Move github-languages plugin to GitHub app only auth
- Use snyk equivalent instead of requiring users to get their own tokens

The snyk work will significantly reduce the chance of accidents, as snyk API tokens are always scoped to the permissions of the user, which in the case of DevX engineers, is quite broad.

## How has it been verified?

Tested locally and verified that we are able to collect data

## Additional notes/concerns

**NB:** users will need to delete their `.env.local` and rerun the setup script in order to run CloudQuery locally. Make sure to retain a copy your GitHub and Snyk tokens first.

With several people using the GitHub token, there's a potential concern of us being rate limited. I don't think this is a significant issue as

- The CODE infrastructure only runs once a month
- Cloudquery DEV has been set up locally in such a way that it's only gathering information for four repositories, so we are making about 1/1000 of the number of calls we would be on PROD or CODE

### DIFF
--- a/packages/cloudquery/docker-compose.yaml
+++ b/packages/cloudquery/docker-compose.yaml
@@ -25,6 +25,7 @@ services:
       CQ_AWS: ${CQ_AWS}
       CQ_GITHUB: ${CQ_GITHUB}
       CQ_SNYK: ${CQ_SNYK}
+      GITHUB_ACCESS_TOKEN: ${GITHUB_ACCESS_TOKEN}
       SNYK_TOKEN: ${SNYK_TOKEN}
       GALAXIES_BUCKET: ${GALAXIES_BUCKET}
     command: sync /dev-config.yaml --log-console --log-level info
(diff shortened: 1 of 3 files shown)


AshCorr (approved):
Nice 👌 this'll make onboarding so much easier!

packages/cloudquery/docker-compose.yaml
```diff
@@ -16,6 +16,9 @@ services:
     volumes:
       - ~/.aws/credentials:/.aws/credentials
       - ./dev-config/cloudquery.yaml:/dev-config.yaml
+      - ~/.gu/service_catalogue/app-id:/app-id
```
AshCorr, line 19: Might be worth making it clear that these are the github app ID and private key?

PR #644, opened 2024-01-03
Retrieve CloudQuery API key for local development

chrislomaxjones:
## What does this change?

As part of the setup script, retrieve a `CLOUDQUERY_API_KEY` from AWS Secrets Manager, and append it to the `.env.local` file.`docker-compose.yml` is changed to make it available within the CloudQuery container itself.

As a small drive-by change, I've also added `registry: github`  to plugins that produced the following errors:
```
ERR exiting with error error="failed to download plugin source guardian/galaxies@v1.1.0: plugin version not found. If you're trying to use a private plugin you'll need to run `cloudquery login` first. Hint: make sure to use the latest plugin version from hub.cloudquery.io or to keep using an outdated version add `registry: github` to your configuration" module=cli
```

## Why?

Some of the CloudQuery plugins used for local development require a `CLOUDQUERY_API_KEY` to run. This PR retrieves a key from Secrets Manager, for use locally.

## How has it been verified?

I've deleted my `.env.local` file and regenerated it using `scripts/setup.sh`. This successfully populates the `CLOUDQUERY_API_KEY` environment variable, and CloudQuery succeeds in populating the tables locally.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -83,6 +83,13 @@ setup_environment() {
   ANGHAMMARAD_SNS_ARN=$(aws ssm get-parameter --name /account/services/anghammarad.topic.arn --profile deployTools --region eu-west-1 | jq '.Parameter.Value' | tr -d '"')
 
   INTERACTIVE_MONITOR_TOPIC_ARN=$(aws sns list-topics --profile deployTools --region eu-west-1 --output text --query 'Topics[*]' | grep interactive-monitor-CODE)
+
+  CLOUDQUERY_API_KEY=$(
+    aws secretsmanager get-secret-value \
+    --secret-id /CODE/deploy/service-catalogue/cloudquery-api-key \
+    --profile deployTools --region eu-west-1 | jq '.SecretString | fromjson["api-key"]'
+  )
+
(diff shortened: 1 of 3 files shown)


scripts/setup.sh
```diff
@@ -83,6 +83,13 @@ setup_environment() {
   ANGHAMMARAD_SNS_ARN=$(aws ssm get-parameter --name /account/services/anghammarad.topic.arn --profile deployTools --region eu-west-1 | jq '.Parameter.Value' | tr -d '"')
 
   INTERACTIVE_MONITOR_TOPIC_ARN=$(aws sns list-topics --profile deployTools --region eu-west-1 --output text --query 'Topics[*]' | grep interactive-monitor-CODE)
+
+  CLOUDQUERY_API_KEY=$(
+    aws secretsmanager get-secret-value \
+    --secret-id /CODE/deploy/service-catalogue/cloudquery-api-key \
```
chrislomaxjones, line 89: Happy to take any suggestions on what stage to use here instead of `CODE`.

NovemberTang: I using the CODE secret makes sense, we're using it everywhere else

NovemberTang (approved):

PR #645, opened 2024-01-04
Do not require snyk token for setup

NovemberTang:
## What does this change?

This is a follow up from #642

Pulls the CODE snyk credentials from secrets manager during repo setup

## Why?

The user no longer needs to set up their own snyk token.
The token provided has the correctly scoped permissions. Snyk user tokens are given the same level of permissions as the respective users, and this cannot be changed. Using the service account token closes this security loophole

## How has it been verified?

Verified that local setup succeeds

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -91,7 +91,7 @@ setup_environment() {
   CLOUDQUERY_API_KEY=$(
     aws secretsmanager get-secret-value \
     --secret-id /CODE/deploy/service-catalogue/cloudquery-api-key \
-    --profile deployTools --region eu-west-1 | jq '.SecretString | fromjson["api-key"]'
+    --profile "$PROFILE" --region "$REGION" | jq '.SecretString | fromjson["api-key"]'
   )
 
   github_info_url="https://github.com/settings/tokens?type=beta"
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -8,6 +8,8 @@ clear='\033[0m'
 yellow='\033[1;33m'
 cyan='\033[0;36m'
 
+PROFILE=deployTools
+REGION=eu-west-1
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR=$(realpath "${DIR}/..")
 STEP_COUNT=0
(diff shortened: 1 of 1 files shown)


scripts/setup.sh
```diff
@@ -78,11 +80,13 @@ setup_environment() {
 
   size_threshold=1
 
-  GALAXIES_BUCKET=$(aws ssm get-parameter --name /INFRA/deploy/services-api/galaxies-bucket-name --profile deployTools --region eu-west-1 | jq '.Parameter.Value' | tr -d '"')
+  SNYK_TOKEN=$(aws secretsmanager get-secret-value --secret-id /CODE/deploy/service-catalogue/snyk-credentials  --profile "$PROFILE" --region "$REGION" | jq -r '.SecretString' | jq -r '."api-key"' | tr -d "'")
```
NovemberTang, line 83: this line is the important bit

scripts/setup.sh
```diff
@@ -70,6 +70,8 @@ check_credentials() {
 }
 
 setup_environment() {
+  PROFILE=$1
```
AshCorr, line 73: Curious when would we ever not be using deployTools and eu-west-1?

scripts/setup.sh
```diff
@@ -91,22 +95,20 @@ setup_environment() {
   )
```
chrislomaxjones, line 95: Now https://github.com/guardian/service-catalogue/pull/644 is in, it might be worth using the region/profile env vars here as well

```bash
  CLOUDQUERY_API_KEY=$(
    aws secretsmanager get-secret-value \
    --secret-id /CODE/deploy/service-catalogue/cloudquery-api-key \
    --profile "$PROFILE" --region "$REGION" | jq '.SecretString | fromjson["api-key"]'
  )
```

scripts/setup.sh
NovemberTang: probably never. I set this function up this way to mimic `check_credentials` (see line 168 for usage)

scripts/setup.sh
NovemberTang: good shout! i didn't catch that in the merge conflict!

AshCorr (approved):

PR #650, opened 2024-01-05
Scan CDK folder

NovemberTang:
## What does this change?

Removes CDK exclusion from snyk

## Why?

It would be useful to know if CDK is vulnerable.
This wasn't working anyway, as our infrastructure now lives in `packages/cdk`.


adamnfish (approved):
Sounds sensible on the face of it, although I'm not familiar with our snyk workflow.

👍 if it works, nice one!

PR #652, opened 2024-01-05
Generate snyk.yml file based on repository languages

NovemberTang:
## What does this change?

Created a function that generates a snyk.yaml, so a repo uses the [DevX snyk action](https://github.com/guardian/.github/blob/main/.github/workflows/sbt-node-snyk.yml) based on the languages present in that repo

This is phase two of larger piece of work that will autogenerate PRs for production repos that contain languages not supported by dependabot. Phase one was retrieving **all** languages from production repos on GitHub, and making that information available to repocop (see #626, #639, #648 and the new [github-languages](https://github.com/guardian/cq-source-github-languages) plugin for some examples of this work).

## Why?

We aim to have all production repos integrated with a dependency monitoring system that covers all its languages. Often, dependabot will be sufficient, but if a repo contains languages not supported by dependabot (such as Scala, or Kotlin), we should integrate it with Snyk. Autogenerating the snyk.yml minimises the amount of work teams have to do towards this aim.

## How has it been verified?

- Unit tests have been written as part of this PR, and run as part of CI.
- The code is also running as part of a set of experimental features, and produces the desired output. One example of this output is:

```
on:
  push:
    branches:
      - main
  workflow_dispatch: 
jobs:
  security:
    uses: guardian/.github/.github/workflows/sbt-node-snyk.yml@main
    with:
      ORG: <SNYK_ORG_ID>
      SKIP_SBT: true
      SKIP_GO: false
    secrets:
      SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
```

## Further work

This work represents an MVP for yaml autogeneration. Nice-to-haves (but nonessentials) include:

- figuring out which snyk orgs belong to which teams
- autodetecting the python version a project will use (even less essential as Python is not a core language).

The next part of phase two is to write the code that actually raises the Snyk integration PR. Some prior art exists in [configure-dependency-management](https://github.com/guardian/configure-dependency-management/), but this example relies heavily on thinly wrapped bash commands to interact with git, as it is designed to run on developer laptops.


packages/repocop/src/remediations/snyk-integrator/snyk-integrator.ts
```diff
@@ -0,0 +1,36 @@
...
+export function createYaml(languages: string[]): string {
+	const inputs = {
+		ORG: '<REPLACE ME>',
+		SKIP_SBT: languages.includes('Scala') ? undefined : true,
+		SKIP_NODE:
+			languages.includes('TypeScript') || languages.includes('JavaScript')
+				? undefined
+				: true,
+		SKIP_PYTHON: languages.includes('Python') ? false : undefined,
+		PYTHON_VERSION: languages.includes('Python') ? '<REPLACE ME>' : undefined,
```
tjsilver, line 12: Maybe this could be more explicit, i.e. `<REPLACE WITH VERSION>`?

tjsilver, line 12: Is `PYTHON_VERSION` ignored if python is skipped?

NovemberTang: yup, anything listed as `undefined` doesn't show up in the final yaml (you can see an example of this in the tests)

chrislomaxjones (approved):
Nice!

PR #653, opened 2024-01-08
Automatically record table frequency

AshCorr:
## What does this change?

Adds a sidecar container to our CloudQuery tasks which handles running an INSERT query on our DB to record the frequency of the CloudQuery task being executed. This frequency is derived from our Schedule specified in CDK.

## Why?

We currently have to manually keep the `cloudquery_table_frequency` up to date which can lead to some tables being missed. By automating this process we can make this table more reliable.

## How has it been verified?

Truncated `cloudquery_table_frequency` in CODE and ran the Galaxies task, observed the creation of new records in `cloudquery_table_frequency`

### DIFF
--- a/packages/cdk/lib/cloudquery/images.ts
+++ b/packages/cdk/lib/cloudquery/images.ts
@@ -14,4 +14,7 @@ export const Images = {
 	ns1Source: ContainerImage.fromRegistry(
 		`ghcr.io/guardian/cq-source-ns1:${Versions.CloudqueryNs1}`,
 	),
+	postgres: ContainerImage.fromRegistry(
+		'public.ecr.aws/docker/library/postgres:16-alpine',
+	),
 };
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -408,7 +408,7 @@ export function addCloudqueryEcsCluster(
 		{
 			name: 'Galaxies',
 			description: 'Galaxies data',
-			schedule: Schedule.rate(Duration.days(1)),
+			schedule: nonProdSchedule ?? Schedule.rate(Duration.days(1)),
 			policies: [
 				readBucketPolicy(
 					`${actionsStaticSiteBucket.bucketArn}/galaxies.gutools.co.uk/data/*`,
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -408,7 +408,7 @@ export function addCloudqueryEcsCluster(
 		{
 			name: 'Galaxies',
 			description: 'Galaxies data',
-			schedule: nonProdSchedule ?? Schedule.rate(Duration.days(1)),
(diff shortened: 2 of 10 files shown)


packages/cdk/lib/cloudquery/images.ts
```diff
@@ -9,4 +9,7 @@ export const Images = {
 	amazonLinux: ContainerImage.fromRegistry(
 		'public.ecr.aws/amazonlinux/amazonlinux:latest',
 	),
+	postgres: ContainerImage.fromRegistry(
+		'public.ecr.aws/docker/library/postgres:16-alpine',
```
AshCorr, line 13: I wasn't able to find an official postgres container which only included PSQL and not the whole Postgress DB binary. In future we might want to publish our own Postgres container which only includes the PSQL binary.

packages/cdk/lib/cloudquery/index.ts
```diff
@@ -406,7 +406,7 @@ export function addCloudqueryEcsCluster(
 		{
 			name: 'Galaxies',
 			description: 'Galaxies data',
-			schedule: nonProdSchedule ?? Schedule.rate(Duration.days(1)),
+			schedule: Schedule.rate(Duration.days(1)),
```
AshCorr, line 409:
```suggestion
			schedule: nonProdSchedule ?? Schedule.rate(Duration.days(1)),
```

packages/cdk/lib/cloudquery/task.ts
```diff
@@ -103,6 +105,79 @@ export interface ScheduledCloudqueryTaskProps
 	cloudQueryApiKey: Secret;
 }
 
+/**
+ * Takes a schedule and figures out its frequency. Supports both CRON and RATE schedules.
+ *
+ * @param schedule - An AWS EventBridge Schedule
+ * @returns an enum representing the rate at which a schedule runs
+ */
+const scheduleFrequency = (
```
akash1810, line 114: This function is a little hard to read, I thin due to the number of unhappy paths. Can't yet think how we could refactor, so happy to leave this to a future PR, however, should we add some unit tests to this PR?

AshCorr: Added some tests!

packages/cdk/lib/cloudquery/schedule.test.ts
```diff
@@ -0,0 +1,52 @@
...
+					weekDay: '1',
+				}),
+			),
+		).toBe('WEEKLY');
+		expect(
+			scheduleFrequency(
+				Schedule.cron({
+					minute: '1',
+					hour: '1',
+					day: '1',
```
akash1810, line 28: TO check my understanding, this'll be monthly?

akash1810 (approved):

packages/cdk/lib/cloudquery/schedule.test.ts
AshCorr: Yep, which I'm just classifying as 'OTHER' for now

PR #655, opened 2024-01-08
Create test PRs for Snyk remediation

chrislomaxjones:
## What does this change?

Add the ability to create pull requests against repositories within the Guardian organisation:

- This change is currently introduced as an experimental feature, and is hardcoded to create a pull request against [test-repocop-prs](https://github.com/guardian/test-repocop-prs/pulls) with a hardcoded set of languages.
- It uses logic introduced in #656 to generate the PR title/body.
- It uses the [`octokit-create-pull-request`](https://www.npmjs.com/package/octokit-plugin-create-pull-request) package (thanks for pointer @NovemberTang) to handle the creation of the pull request itself. I've wrapped this in a small wrapper, so we could drop this dependency in the future with minimal refactoring of the caller. This package can be used as an Octokit plugin, although this seemed more fiddly than just passing an instance of Octokit in directly, hence the usage of `composeCreatePullRequest`.

## Why?

This will allow us to begin automatically raising pull requests against repositories that require Snyk integration.

## How has it been verified?

I've tested this using my own Github PAT locally and am able to create PRs against [test-repocop-prs](https://github.com/guardian/test-repocop-prs/pulls).

See https://github.com/guardian/test-repocop-prs/pull/8 for an example of a successful invocation. For now, we're hiding the logic behind a feature flag.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9974,6 +9974,27 @@
         "node": ">= 18"
       }
     },
+    "node_modules/octokit-plugin-create-pull-request": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/octokit-plugin-create-pull-request/-/octokit-plugin-create-pull-request-5.1.1.tgz",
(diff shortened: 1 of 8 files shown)


tjsilver (approved):
👍

PR #661, opened 2024-01-09
Fix Snyk Python options

chrislomaxjones:
## What does this change?

Add options `PIP_REQUIREMENTS_FILES` and `PIPFILES` to the Snyk inputs. 

We also modify the workflow so that it is run when commits are pushed to the generated branch containing the snyk.yml file.

## Why?

These options are required to properly run the Snyk workflow when a repository contains Python dependencies.

## How has it been verified?

See the updated tests.


tjsilver (approved):
👍

PR #665, opened 2024-01-09
Create a separate lambda for raising snyk PRs

NovemberTang:
## What does this change?

**n.b.** The actual diff for this change is only about 130 lines long, the rest is snapshot changes

We have moved the code that raises snyk PRs into its own lambda.

Eventually, repocop will send a list of repository names to snyk-integrator via SNS, and it will generate a PR for each of them. This work, however, is scoped to setting up the initial lambda and making sure that separating the snyk code from the rest of the repocop code works.

New credentials will not be required for local setup, as readonly access is sufficient for non-prod environments.

## Why?

Having a lambda that is able to write arbitrarily to CloudQuery tables, as well as having write access to repos provides a lot of potential to cause chaos if something goes wrong, so let's split them up.

## How has it been verified?

CI passes, deployed to CODE and both functions still work.


tjsilver (approved):
Looks good!

PR #666, opened 2024-01-09
Run unit tests in parallel

NovemberTang:
## What does this change?

When using NPM workspaces, there is not a built in way of running the same command across multiple workspaces in parallel, so I've created a way to do this in bash.

It does now mean that every package now requires an explicit test command, which is a little bit of additional friction for new projects. However, forcing the user to declare that their project has no unit tests is not necessarily a bad thing, and may encourage people to create tests in the future.

## Why?

Make CI faster

## How has it been verified?

Check [this](https://github.com/guardian/service-catalogue/actions/runs/7465620961/job/20315198646) action run. CI time has been cut 20% (from 2m30s to 2m), and the benefit is more pronounced on local machines with more cores.


scripts/ci.sh
```diff
@@ -45,7 +45,7 @@ verifyMarkdown() {
 
 npm ci
 npm run typecheck & npm run lint
-npm run test
+npm query .workspace | jq -r '.[].location' | sed -e 's,.*/,,' | xargs -P 0 -L1 npm run test -w
```
chrislomaxjones, line 48: We might be able to avoid having to define the empty test scripts by using the `--if-present` flag.

```suggestion
npm query .workspace | jq -r '.[].location' | sed -e 's,.*/,,' | xargs -P 0 -L1 npm run test --if-present -w
```

PR #667, opened 2024-01-10
Use esbuild to transpile Jest tests

chrislomaxjones:
## What does this change?

Use `esbuild` (via [`esbuild-jest`](https://github.com/aelbore/esbuild-jest)) to transform our Jest tests.

## Why?

Should speed up our testing (especially when combined with #666).

## How has it been verified?

On my laptop, here's the output of running `time npm run test` on `main` at the time of this PR:

```
npm run test  37.39s user 4.24s system 161% cpu 25.810 total
```

And here it is on this branch

```
npm run test  4.69s user 0.89s system 101% cpu 5.481 total
```

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,15 +4,17 @@ const dotenv = require('dotenv');
 // Load environment variables from .env file at the root of the repository
 dotenv.config({ path: `${__dirname}/.env` });
 
+const transform = {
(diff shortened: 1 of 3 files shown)


AshCorr (approved):
🤯 what an improvement!

chrislomaxjones:
> Can we get rid of ts-jest from our dependencies now or is it still needed?

Yes, great spot! Have added that to this PR.

PR #670, opened 2024-01-10
Run tests in parallel

NovemberTang:
## What does this change?

When using NPM workspaces, there is not a built in way of running the same command across multiple workspaces in parallel, so we've created a way to do this in bash.

## Why?

Make CI faster

## How has it been verified?

Cuts CI time by about 15 seconds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5851,10 +5851,6 @@
         "node": ">=0.8"
       }
     },
-    "node_modules/cloudquery": {
-      "resolved": "packages/dev-environment",
-      "link": true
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
Nice!

akash1810:
I think this is masking some errors. https://github.com/guardian/service-catalogue/pull/683 takes a different approach.

Will need to perform some testing, but I think the lint step is the most expensive in our build. https://github.com/guardian/csnx/pull/1083 aims to improve that.

PR #673, opened 2024-01-11
Fail repocop if a required table is empty

NovemberTang:
## What does this change?

Checks that tables are populated before executing repocop

## Why?

Having an empty table means that our data is innacurate, which means potentially unwanted actions could occur. For example, if it looks like no snyk projects exist, we could start raising PRs against lots of repos that are already correctly integrated. The lambda should halt and fail if data is missing.

## How has it been verified?

Tests have been added to verify the expected behaviour.

Deployed and tested, logs have enough information to deduce which table failed.
```
{"errorType":"Error","errorMessage":"Expected a non-empty array. Source table may be empty.",
"stack":["Error: Expected a non-empty array for object","    at toNonEmptyArray
(/var/task/index.js:844551:11)","    at getSnykProjects (/var/task/index.js:844617:10)",
"    at async Runtime.main [as handler] (/var/task/index.js:845154:24)"]}
```

### DIFF
--- a/packages/repocop/src/query.test.ts
+++ b/packages/repocop/src/query.test.ts
@@ -2,7 +2,7 @@ import { toNonEmptyArray } from './query';
 import type { NonEmptyArray } from './query';
 
 describe('Failure on empty arrays', () => {
-	test('test me', () => {
+	test('throw an error if input is an empty array', () => {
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
👍

PR #676, opened 2024-01-15
Check for existence of snyk workflow file

NovemberTang:
## What does this change?

While we can't check snyk project tags directly, we can use the presence of a snyk workflow file as an indicator that a repo is integrated with snyk.

Adds a helper function to the tests. Rather than having to declare all empty arrays up front, we use a function that defaults them to empty unless otherwise specified. This means if we add or remove a field from `evaluateOneRepo`, there's far fewer changes we need to make to keep the tests happy.

Removes the non-empty check from snyk_projects while we have the snyk workflow check in place, as the latter is a workaround for the former being empty.

## Why?

To allow us to continue to measure and make progress on our OKR while work to bring back the table continues, albeit with slightly lowered precision

## How has it been verified?

Added new unit test to verify new behaviour.
Deployed to code and verified number of unintegrated repos dropped to about expected level. (Expected ~60 based on last week's numbers. Actual = 56)

### DIFF
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -201,12 +201,16 @@ export function hasDependencyTracking(
 
 	//This is a temporary workaround until we get the snyk_projects table back.
 	function findSnykYaml(repo: Repository, workflowFiles: github_workflows[]) {
-		return workflowFiles.find(
+		const result = workflowFiles.find(
 			(file) =>
 				file.repository_id === repo.id &&
 				!!file.path &&
 				file.path.includes('snyk'),
 		);
+		if (result !== undefined) {
+			console.log(`${repo.name} has a snyk workflow file.`);
+		}
+		return result;
 	}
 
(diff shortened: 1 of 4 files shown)


packages/repocop/src/rules/repository.ts
```diff
@@ -197,16 +199,41 @@ export function hasDependencyTracking(
...
-			tags.repo == repo.full_name &&
-			tags.branch === repo.default_branch,
-	);
+	//This is a temporary workaround until we get the snyk_projects table back.
+	function snykYamlExists(repo: Repository, workflowFiles: github_workflows[]) {
+		const result = workflowFiles.find(
+			(file) =>
+				file.repository_id === repo.id &&
+				!!file.path &&
+				file.path.includes('snyk'),
```
NovemberTang, line 208: Leaving ambiguous catch all potential name variations. yaml/yml etc.

tjsilver (approved):
👍

PR #677, opened 2024-01-15
update PR body

NovemberTang:
## What does this change?

Adds more detailed instructions to the PR body. Removes gh cli step.

## Why?

The yaml generated includes a field that will run on every push to the feature branch. This is because a user cannot run an action manually if it has not been kicked off by github at least once. Now we are running on every push, we don't need to instruct users to run the action via the CLI.

Python is not used in every repo, so make clear that this part is optional. In future, maybe we could only display that checklist item if we detect python in the repo?

## How has it been verified?

Unit tests verify expected behaviour


tjsilver (approved):

chrislomaxjones (approved):
👍

PR #678, opened 2024-01-15
Allow snyk integrator lambda to take input

NovemberTang:
## What does this change?

`snyk-integrator` is triggered in response SNS events. It tries to unwrap those events and generate PRs based on their contents.

## What does this NOT change?

`repocop` does not currently send notifications to snyk integrator via this queue. This will be set up via a separate PR to keep this one manageable.

## Why?

So repocop can (eventually) tell snyk integrator which repos require PRs to be raised.

## How has it been verified?

Tested on CODE

### DIFF
--- a/packages/cdk/lib/repocop.ts
+++ b/packages/cdk/lib/repocop.ts
@@ -11,7 +11,9 @@ import type { Schedule } from 'aws-cdk-lib/aws-events';
 import { Runtime } from 'aws-cdk-lib/aws-lambda';
 import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
 import { Secret } from 'aws-cdk-lib/aws-secretsmanager';
-import type { ITopic, Topic } from 'aws-cdk-lib/aws-sns';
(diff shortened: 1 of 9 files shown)


packages/cdk/lib/repocop.ts
```diff
@@ -27,6 +29,14 @@ export class Repocop {
 		dbSecurityGroup: SecurityGroup,
 		repocopGithubSecret: Secret,
 	) {
+		const snykIntegratorInputTopic = new Topic(
+			guStack,
+			`snyk-integrator-input-topic-${guStack.stage}`,
```
NovemberTang, line 34: Two topics cannot have the same name, so make them stage aware

packages/common/src/functions.ts
```diff
@@ -67,6 +68,10 @@ export async function stageAwareOctokit(stage: string) {
 	}
 }
 
+//TODO test me
```
chrislomaxjones, line 71:
```suggestion
```

chrislomaxjones (approved):
Looks good!

PR #683, opened 2024-01-15
ci: Orchestrate using `npm-run-all`

akash1810:
## What does this change?
We're currently not exiting the build on the first error (see https://github.com/guardian/service-catalogue/pull/680).

This change adopts https://github.com/mysticatea/npm-run-all to run npm scripts in parallel, on the assumption that observes exit codes more reliably.

## Why?
CI needs to be reliable.

## How has it been verified?
TBD.


scripts/ci.sh
```diff
@@ -44,12 +44,15 @@ verifyMarkdown() {
...
-  createZip "interactive-monitor" & \
-  createZip "snyk-integrator" & \
-  createLambdaWithPrisma "repocop" & \
-  createLambdaWithPrisma "data-audit"
+npm test
+
+# Run the following in parallel.
+# Logs will be interleaved, but the `--print-label` flag will prefix each line with the name of the script being run.
+# See https://github.com/mysticatea/npm-run-all.
+npx npm-run-all --print-label --parallel typecheck lint synth build
```
NovemberTang, line 52: can we use this as the test script so developers can get the speed benefit of parallelisation too?

akash1810: Possibly. I'm uncertain of the use-case for performing all these at once locally though?

AshCorr (approved):
Looks good. Might want to split these into separate CI jobs at one point?

PR #684, opened 2024-01-16
Minor fixes/improvements

NovemberTang:
## What does this change?

~~1. Use the parallel test command in the npm script and reference it in CI script~~
2. Fix the dev environment README commands
3. Fix the pre commit hook so it searches for an exact file match (previously it was warning users about all staged files in `dev-environment` as it matches the regex `.env`)

## Why?

~~1. Faster test execution~~
2. Better developer experience
3. Prevent the pre-commit hook from flagging files we don't care about

## How has it been verified?

Test, hooks and local execution of the dev environment have all been verified on developer machine

## Additional notes

In order to use the pre-commit hook it will need to be reinstalled. The fastest way to do this is by rerunning the setup script.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -9,7 +9,7 @@ if [[ $(git diff --cached --name-only | grep -w .env) ]]; then
   exec < /dev/tty
   read REPLY
   if [[ ! $REPLY =~ ^[Yy]$ ]]; then
-    echo -e "${red}Aborting commit.${clear} Please discard unintended changes with $(git restore .env) before committing again."
+    echo -e "${red}Aborting commit.${clear} Please discard unintended changes with `git restore .env` before committing again."
     exit 1
   fi
 fi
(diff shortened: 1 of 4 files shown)


akash1810:
This way of running the tests is masking some errors. https://github.com/guardian/service-catalogue/pull/683 proposes an alternative solution.

.hooks/pre-commit
```diff
@@ -3,7 +3,7 @@ red="\033[0;31m"
 clear="\033[0m"
 yellow="\033[1;33m"
 #check for staged changes in .env and ask user to confirm
-if [[ $(git diff --cached --name-only | grep .env) ]]; then
+if [[ $(git diff --cached --name-only | grep -w .env) ]]; then
```
adamnfish, line 6: Rather than (or perhaps as well as) trying to avoid secrets being *committed*, can we keep them out the repository completely? We often use ~/.gu as a place to put things that shouldn't accidentally be added to version control.

NovemberTang: we already do this, see the `setup_environment()` function in the setup script. The .env file holds non-private, frequently changing information such as plugin versions. However, people were committing secret information to this file often enough that an extra layer of protection was necessary.

AshCorr (approved):

PR #686, opened 2024-01-16
Send repocop metrics to cloudwatch

NovemberTang:
## What does this change?

Publishes a selection of repocop compliance metrics as cloudwatch datapoints. This proof of concept uses three, the rest will be added in a subsequent PR to keep things.

I've given the metrics different names to the columns in the schema to make the meaning of the numbers clearer, rather than prioritising consistency with the DB schema

n.b the metrics were relabelled _after_ the screenshot below was taken. However, the lambda has been run using the latest commit on this branch, and still works.

## Why?

So we can track our departmental compliance over time.

## How has it been verified?

- Created new unit tests to verify arithmetic, rounding, null handling.
- Deployed to CODE to test that we have sufficient permissions to publish the datapoints

## Cost considerations

Cloudwatch custom metrics cost $0.30 per metric, per month. We have ~10 repocop metrics, which currently represents an additional cost of about $3 a month. This is manageable. However, if we start tracking hundreds of these numbers, the cost increases considerably.

## Future considerations

It would be nice to be able to generate these metrics automatically from the table schema. This would mean that new datapoints could be added automatically, rather than the currently proposed system, which is to create the datapoints by hand. While more convenient from a maintenance point of view, the fully automatic approach would represent a significant amount of additional work, and potentially research into [metaprogramming](https://en.wikipedia.org/wiki/Metaprogramming), which is beyond the scope of this PR, and may not be worth it until we have a large enough number of datapoints that the code becomes unmanageable.

## Images

Screenshot 2024-01-16 at 11 42 44


packages/cdk/lib/repocop.ts
```diff
@@ -64,11 +65,23 @@ export class Repocop {
 			repocopLampdaProps,
 		);
 
+		//allow repocop to create and put metrics
+		const policyStatement = new PolicyStatement({
+			actions: ['cloudwatch:PutMetricData'],
```
NovemberTang, line 70: PutMetricData is sufficient to create metrics if they don't exist

packages/repocop/src/metrics.ts
```diff
@@ -0,0 +1,86 @@
...
+export async function sendToCloudwatch(
+	evaluatedRepos: repocop_github_repository_rules[],
+	cloudwatch: CloudWatchClient,
+) {
+	console.log('Sending metrics to Cloudwatch');
+	await cloudwatch.send(
+		new PutMetricDataCommand({
+			Namespace: 'repocop',
+			MetricData: [
+				createMetric(
```
NovemberTang, line 59: I've included only metrics that are currently implemented. The others always return `true` or `null`

packages/repocop/src/metrics.ts
```diff
@@ -0,0 +1,86 @@
+import {
+	type CloudWatchClient,
+	type MetricDatum,
+	PutMetricDataCommand,
+} from '@aws-sdk/client-cloudwatch';
+import type { repocop_github_repository_rules } from '@prisma/client';
+import { getEnvOrThrow } from 'common/src/functions';
+
+export function getPercentageTrue(booleanishArray: Array<boolean | null>) {
+	const array = booleanishArray.filter((x) => x !== null) as boolean[];
```
chrislomaxjones, line 10: It might be a slightly more idiomatic to use a type guard ([some docs here](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types)) here than an assertion, but up to you!

```suggestion
	const array = booleanishArray.filter((x): x is boolean => x !== null);
```

packages/repocop/src/metrics.ts
```diff
@@ -0,0 +1,86 @@
...
+	cloudwatch: CloudWatchClient,
+) {
+	console.log('Sending metrics to Cloudwatch');
+	await cloudwatch.send(
+		new PutMetricDataCommand({
+			Namespace: 'repocop',
+			MetricData: [
+				createMetric(
+					'DefaultBranchIsMain',
+					evaluatedRepos.map((x) => x.default_branch_name),
```
chrislomaxjones, line 61: Again, up to you, but you could destructure these to avoid defining a parameter `x`.

```suggestion
					evaluatedRepos.map(({ default_branch_name }) => default_branch_name),
```

NovemberTang: I tried this suggestion, but got a linting error so will leave it how it is for now

Screenshot 2024-01-16 at 13 55 46

NovemberTang: I tried this and got a bunch of linting errors, so will leave it how it is for now

tjsilver (approved):
Great!

PR #690, opened 2024-01-16
Get repocop to send messages to the snyk integrator

NovemberTang:
## What does this change?

Added an additional feature to repocop. It grabs DevX repos that are not correctly integrated with Snyk, and sends one of them to the Snyk Integrator lambda for processing. I am limiting it to DevX repos for now so that we can verify the first set of PRs before rolling out to the rest of the department

It will only send a repo on if **ALL** of its languages are supported by our action. Currently, this includes Scala, Typescript, JavaScript, Rust and Python. We hope to add a few more in the coming weeks.

## Why?

Automatically raising PRs against repos that we know can be successfully integrated with snyk makes it much easier for teams to track their vulnerablities accurately.

## How has it been verified?

- Unit tests have been added.
- This change has been deployed to CODE and tested end to end. Snyk integrator is able to pick up the message and create the correct looking snyk.yml

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -16,3 +16,8 @@ export type GitHubAppConfig = {
 	strategyOptions: StrategyOptions;
 	installationId: string;
 };
+
+export interface SnykIntegratorEvent {
+	name: string;
+	languages: string[];
+}
--- a/packages/repocop/src/languages.ts
+++ b/packages/repocop/src/languages.ts
@@ -1,0 +1,46 @@
+export const ignoredLanguages = [
+	'HTML',
+	'CSS',
+	'Shell',
+	'Jupyter Notebook',
+	'Makefile',
+	'Dockerfile',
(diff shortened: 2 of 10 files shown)


packages/repocop/src/languages.ts
```diff
@@ -0,0 +1,54 @@
+export const ignoredLanguages = [
```
NovemberTang, line 1: this has moved from `packages/repocop/src/rules/repository.ts`

packages/common/src/types.ts
```diff
@@ -16,3 +16,8 @@ export type GitHubAppConfig = {
 	strategyOptions: StrategyOptions;
 	installationId: string;
 };
+
+export interface SnykIntegratorEvent {
```
NovemberTang, line 20: this has moved from `packages/snyk-integrator/src/index.ts`

tjsilver (approved):
Excited to see PRs appear!

PR #691, opened 2024-01-16
Collect fastly statistics

akash1810:
## What does this change?
@davidfurey is interested in surfacing Fastly statistic data in Grafana to make it easier to anticipate expenditure. This change starts to [collect this information](https://hub.cloudquery.io/plugins/source/cloudquery/fastly/v3.0.5/tables/fastly_stats_services). It'll run hourly. Depending how long the task takes, we can collect the data more/less frequently - being real-time would be the most valuable.

CloudQuery provide an [example query](https://hub.cloudquery.io/plugins/source/cloudquery/fastly/v3.0.5/docs?table=fastly_stats_services#overview-retrieve-stats-for-a-service-for-a-specific-time-period) that this data can answer.

This change also bumps the version of the Fastly plugin. Whilst it is a major version number change, it looks like [v3.0.0](https://hub.cloudquery.io/plugins/source/cloudquery/fastly/v3.0.0/versions) was created to move it to the premium tier, rather than for any breaking changes. Arguably this should be it's own PR - happy to split it if necessary.

### DIFF
--- a/.env
+++ b/.env
@@ -18,7 +18,7 @@ CQ_AWS=23.3.1
 CQ_GITHUB=7.4.2
 
 # See https://github.com/cloudquery/cloudquery/releases?q=plugins-source-fastly
-CQ_FASTLY=2.1.12
+CQ_FASTLY=3.0.6
 
 # See https://github.com/guardian/cq-source-galaxies
 CQ_GUARDIAN_GALAXIES=1.1.0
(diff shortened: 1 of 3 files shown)


akash1810:
Running this on CODE has yielded this error:

> 400 - Bad Request:
>
>    Title:  You cannot retrieve hourly historical statistics data older than 375 days from the current date. Learn more about our availability of data at: https://docs.fastly.com/api/stats#Availability

https://developer.fastly.com/reference/api/metrics-stats/historical-stats/ suggests we can contact Fastly about our data needs. [CloudQuery](https://hub.cloudquery.io/plugins/source/cloudquery/fastly/v3.0.5/docs#overview-configuration) doesn't appear to support setting of start dates, so we'll be collecting all the data all the time, which feels quite expensive.

It might be easier to use https://github.com/fastly/fastly-exporter, and https://grafana.com/grafana/dashboards/8951-fastly-dashboard/ to see this data in Grafana.

PR #693, opened 2024-01-17
Add tags to CloudQuery task instances

AshCorr:
## What does this change?

Adds tags to running instances of Tasks.

## Why?

They previously did not have any, this allows us to track ECS costs.


AshCorr:
Closed in favour of #777

PR #702, opened 2024-01-18
Add PRs to project board

NovemberTang:
## What does this change?

In PROD, once a PR has been raised, we should add it to the [project board](https://github.com/orgs/guardian/projects/110/views/1)

## Why?

Allows us to keep track of the PRs we've raised, so we know who we might need to follow up with

## How has it been verified?

Deployed to PROD and verified that the PRs are raised and added to the project board

## Images

Screenshot 2024-01-18 at 12 09 40

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3448,7 +3448,8 @@
     },
     "node_modules/@octokit/graphql": {
       "version": "7.0.2",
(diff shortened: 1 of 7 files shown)


tjsilver (approved):
This is great!

PR #705, opened 2024-01-18
Only trigger snyk integrator if a repo exists

NovemberTang:
## What does this change?

Previously, repocop would try to pass an undefined value to SNS, which would cause a failure. Now, we are checking to make sure the value exists first.

## Why?

Prevent repocop failures


tjsilver (approved):
👍

PR #706, opened 2024-01-18
Tweak logging

NovemberTang:
## What does this change?

1. Remove log lines in `packages/repocop/src/rules/repository.ts` that show up hundreds of times
2. Logs in `send-to-sns.ts` returns a string rather than [Object object]
3. Only log repos with contactable owners if we find some

## Why?

Make the logs less noisy
Number 3 in particular was taking up a lot of space in the test output


chrislomaxjones (approved):
Nice 🧹

PR #707, opened 2024-01-19
Fix local execution of repocop/cloudquery. Add yaml linting

NovemberTang:
## What does this change?

1. Skips some GitHub tables
2. Collect github-languages last
3. Lint YAML in dev-environment as part of testing
4. Set required environment variables for local repocop execution

## Why?

1. For release and dependabot tables, we are skipping them in PROD. For the workflow tables, we are not using them for anything. Skipping them means less time waiting around for rate limits to reset, and faster data collection overall
2. This job is quite slow.
3. More confidence in changes we make to YAML files
4. repocop fails immediately if it cannot find these env vars.

## How has it been verified?

- Collecting CQ data via dev-environment still works
- Tests pass
- Repocop errors relating to missing environment variables go away.


packages/repocop/package.json
```diff
@@ -6,7 +6,7 @@
 	},
 	"scripts": {
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node20 --outdir=dist --external:@aws-sdk --external:@prisma/client --external:prisma",
-		"start": "APP=repocop tsx src/run-locally.ts",
+		"start": "APP=repocop STACK=deploy tsx src/run-locally.ts",
```
akash1810, line 9: `STACK` is going to be the same for every app in this repository. Should we set it in the [.env file, similar to `STAGE`](https://github.com/guardian/service-catalogue/blob/0e28c4e78480b62d8248ffaea5a61c2cbfbfae28/.env#L40)?

akash1810:
> 3\. Lint YAML in dev-environment as part of testing

Do we need to do this? Will we ever run the added tests locally? Or would we use a combination of IDE hints, and using the YAML file as a way to validate the file?

packages/repocop/package.json
NovemberTang: I stuck it in here as it's the only place `STACK` is required, I'm agnostic about which approach we use.

NovemberTang:
> > 3. Lint YAML in dev-environment as part of testing
> 
> Do we need to do this? Will we ever run the added tests locally? Or would we use a combination of IDE hints, and using the YAML file as a way to validate the file?

This depends on how the file has been edited. If it was a quick nano/vim thing, or someone edits the file directly on GitHub, it's possible to commit invalid yaml, and I think it's worth catching that at CI time. FWIW, it's a very quick process, running in under a second on my machine. `0.23s user 0.07s system 0.684 total`

akash1810:
> This depends on how the file has been edited. If it was a quick nano/vim thing, or someone edits the file directly on GitHub, it's possible to commit invalid yaml, and I think it's worth catching that at CI time.
> 
> Personally, I'm not super confident with yaml and having something else check my mistakes gives me peace of mind.

Wonder if a [VIM plugin](https://vimawesome.com/?q=yaml), installed on individual developer's machines might be an alternative. Not blocking though.

akash1810 (approved):

PR #708, opened 2024-01-19
Create snyk-integrator README

NovemberTang:
## What does this change?

Documents what the purpose of snyk-integrator is, and how to use it.

## Why?

Documentation is good!

## How has it been verified?

Markdown is correctly formatted, all the links work.
You can see a nicely rendered version [here](https://github.com/guardian/service-catalogue/blob/eb523a5d3ad2049a96cf3a475fec430ba532642d/packages/snyk-integrator/README.md)

### DIFF
--- a/packages/snyk-integrator/README.md
+++ b/packages/snyk-integrator/README.md
@@ -2,7 +2,7 @@
 
 ### What does it do?
 
-The snyk integrator takes repo name and it's languages as JSON, like so:
+The snyk integrator takes a repo name and it's languages as JSON, like so:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


packages/snyk-integrator/README.md
```diff
@@ -0,0 +1,40 @@
...
+The snyk integrator takes a repo name and it's languages as JSON, like so:
+
+```json
+{
+	"name": "service-catalogue",
+	"languages": ["TypeScript", "Shell", "Dockerfile"]
+}
+```
+
+It uses this input to generates a YAML file, matching the specification of our [GitHub action](https://github.com/guardian/.github/blob/main/.github/workflows/sbt-node-snyk.yml). This action runs `snyk monitor` on pushes to the main branch, so that we can keep track of vulnerabilities introduced by third party dependencies.
```
akash1810, line 14: minor: It's a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows), rather than an action.

packages/snyk-integrator/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Running on non-production environments
+
+In non-production environments, such as CODE, or when running locally, the snyk integrator will not create a PR. Instead, it will print the contents of the YAML file to the console. This allows developers to test the the core logic (creating the yaml file), without unintended side effects.
+
+#### CODE
+
+The lambda can safely be invoked in the CODE environment, as not only will it not attempt to create a PR, but there are no GitHub credentials available to it, attempting to instantiate a GitHub client will fail.
+
+The format of input to the lambda on the CODE environment is that of an SNS message. It's not trivial to construct, so some sample events have been provided in the Test tab of the lambda in the AWS console.
```
akash1810, line 36: minor: If we [script](https://docs.aws.amazon.com/cli/latest/reference/lambda/invoke.html) this, we could keep the shape of the payload in sync with the expected input as tracked in VCS.

akash1810 (approved):
Couple of minor comments, nothing blocking.

packages/snyk-integrator/README.md
NovemberTang: Yeah definitely a good shout! This is something I'd like to do for all the security jobs at some point.

PR #709, opened 2024-01-19
fix(cloudquery): Skip the `aws_emr_supported_instance_types` table

akash1810:
This table has ~5 million rows, and isn't especially useful.

It lists the instance types we could use, and their specification. It is fairly repetitive across all regions and AWS accounts.

As we don't use AWS EMR, skip the table.

See https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-supported-instance-types.html.

Note: We'll need to [apply the migration](https://github.com/guardian/service-catalogue/blob/main/docs/database-migrations.md) manually.

<details><summary>Exactly how many rows?</summary>
<p>

```sql
SELECT relname, reltuples AS estimate 
FROM pg_class 
WHERE relname IN (
  SELECT TABLE_NAME
  FROM information_schema.tables
  WHERE TABLE_NAME not like 'pg_%' AND table_schema in ('public')
) 
ORDER BY estimate DESC
limit 5;
```

| relname | estimate |
|--------|--------|
| aws_cloudtrail_events | 5870568 |
| aws_emr_supported_instance_types | 4968076 | 
| aws_securityhub_findings | 1076861 | 
| riffraff_deploys | 711383 | 
| aws_iam_policy_last_accessed_details | 229564 | 

</p>
</details>

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -15307,6 +15307,7 @@ spec:
     - aws_elasticache_parameter_groups
     - aws_elasticache_reserved_cache_nodes_offerings
     - aws_elasticache_service_updates
+    - aws_emr_supported_instance_types
     - aws_neptune_cluster_parameter_groups
     - aws_neptune_db_parameter_groups
     - aws_rds_cluster_parameter_groups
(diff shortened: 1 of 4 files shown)


adamnfish (approved):
Great idea

akash1810:
> Note: We'll need to [apply the migration](https://github.com/guardian/service-catalogue/blob/main/docs/database-migrations.md) manually.

Confirming that I've now done this on CODE and PROD.

PR #711, opened 2024-01-19
Improve PR instructions

NovemberTang:
Thanks to @michaelwmcnamara for his invaluable feedback on the PR body

## What does this change?

While testing this feature on some repos belonging to other DevX streams, we received feedback that the instructions were not as clear as they could be. This PR aims to rectify that, by using more specific language, and clarifying terms where required.

We now also log the PR title and body on CODE and DEV

## Why?

To make it easier for teams to follow the instructions

## How has it been verified?

- PR body still formats as expected
- Verified that the Python checklist item shows up when Python is listed as a language, and vice versa.

### DIFF
--- a/packages/snyk-integrator/src/snyk-integrator/snyk-integrator.ts
+++ b/packages/snyk-integrator/src/snyk-integrator/snyk-integrator.ts
@@ -86,14 +86,21 @@ function checklist(items: string[]): string {
 
 function createPRChecklist(languages: string[], branchName: string): string[] {
 	const pythonText = ['Replace the relevant python fields'];
-	const defaultText = [
-		'Replace the SNYK_ORG variable with one that your team already uses (you should have other repos integrated with Snyk. ' +
-			'If you can’t find any, reach out to DevX)',
-		'Replace the relevant python fields, if they exist in the file. If not, skip this step',
+	const step1 =
+		'Replace the SNYK_ORG variable with the org name that your team ' +
+		'already uses (you should have other repos integrated with Snyk. ' +
+		'If you can’t find any, reach out to DevX). Examples are ' +
+		'`guardian-devtools` and `guardian-dotcom-n2y`';
+	const step2 =
 		'The job should run automatically on every commit to this branch. ' +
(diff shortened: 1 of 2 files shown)


packages/snyk-integrator/src/snyk-integrator/snyk-integrator.ts
```diff
@@ -84,6 +84,31 @@ function checklist(items: string[]): string {
...
+	const step2 =
+		'The Snyk job should run automatically on every commit to this branch. ' +
+		'Click through on the Snyk status check see the logs of the latest run on ' +
+		'this PR, and verify it has generated one project per dependency manifest ' +
+		'(except pnpm and deno). ' +
+		'Examples of dependency manifests are a build.sbt, or a package-lock.json, ' +
+		'essentially, any file that lists the dependencies of your project.';
+	const step3 =
+		`When you are happy the action works, remove the branch name \`${branchName}\`` +
+		'trigger from the snyk.yml (aka delete line 6), approve, and merge.';
```
akash1810, line 104: Interested to hear opinions on https://github.com/guardian/service-catalogue/pull/712, which tries to create a simpler approach to creating long, multi-line strings.

NovemberTang: Looks useful, though would need modifying to suit this use case. These are transformed into checklist items, so each step is a single line.

tjsilver (approved):
👍

packages/snyk-integrator/src/snyk-integrator/snyk-integrator.ts
akash1810: Using the `stripMargin` function means string concatenation goes away:

```ts
const step1 = stripMargin`
  |Replace the SNYK_ORG variable with the org name that your team already uses (you should have other repos integrated with Snyk.
  |If you can’t find any, reach out to DevX).
  |Examples are 'guardian-devtools' and 'guardian-dotcom-n2y'`;
```

PR #713, opened 2024-01-22
Get critical dependabot vulnerabilities for some production repos

NovemberTang:
## What does this change?

Creates a function that, given a repo, will return the first page of dependabot alerts for that repo.

## Why?

So we know if a repository has dependabot vulnerabilities that need addressing.

## How has it been verified?

Run locally using known vulnerable repos, returns expected output. Run on CODE, verified output, and that error handling works as expected

## Additional notes

Sometimes the request fails because dependabot scanning is not enabled for the repo. We should look into whether or not this is opt-in or opt-out, and whether it's appropriate/possible to switch it on at the org level for production repos, similar to branch protection rulesets.

### DIFF
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -12,6 +12,7 @@ import {
 	supportedSnykLanguages,
 } from '../languages';
 import type {
+	Alert,
 	AwsCloudFormationStack,
 	RepoAndStack,
 	Repository,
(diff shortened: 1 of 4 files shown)


packages/repocop/src/rules/repository.ts
```diff
@@ -269,12 +272,52 @@ function findArchivedReposWithStacks(
 	return archivedReposWithPotentialStacks;
 }
 
-export function testExperimentalRepocopFeatures(
+export async function getAlertsForRepo(
+	octokit: Octokit,
+	name: string,
+): Promise<Alert[] | undefined> {
```
NovemberTang, line 278: We return undefined on error rather than an empty array to differentiate between a failed call and one that correctly returns no vulnerabilities.

tjsilver (approved):
Looks good!

PR #714, opened 2024-01-22
refactor: Remove IIFEs from `run-locally.ts` files

chrislomaxjones:
## What does this change?

Remove the [immediately invoked function expressions](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) from the `run-locally.ts` files.

## Why?

They aren't strictly necessary, as we don't need to do a top-level await when we wrap all of our logic in an async `main` function.


tjsilver (approved):
👍

PR #715, opened 2024-01-22
Prevent Snyk integrator raising duplicate pull requests

chrislomaxjones:
## What does this change?

This PR contains two main changes:

1. Check if a PR already exists for Snyk integration on a given repo, before creating the integration PR. We check for PR existence by searching the PRs for a given branch name.
2. The branch name is now derived from the set of languages passed to the snyk-integrator lambda. We take the languages into account here so that if we raise a Snyk integration PR and at a later date a new language is added to the repo (that requires integration), we'll go ahead and raise another PR.

## Why?

Prevent the same Snyk integration PR being raised multiple times between the `snyk.yml` file actually being merged. For example: https://github.com/guardian/devx-backup-monitoring/pull/14 https://github.com/guardian/devx-backup-monitoring/pull/16.

## How has it been verified?

I tested this locally against [test-repocop-prs](https://github.com/guardian/test-repocop-prs/pulls), removing the PROD check in the snyk integrator temporarily. The test plan was as follows:

1. Run the Snyk integrator with input `{ name: "test-repocop-prs", languages: ["Scala", "TypeScript"] }`. This created [this PR](https://github.com/guardian/test-repocop-prs/pull/15) in the repository.
2. Running the Snyk integrator again with the same input `{ name: "test-repocop-prs", languages: ["Scala", "TypeScript"] }` **does not** cause another PR to be created.
3. Running the Snyk integrator with a new languages `{ name: "test-repocop-prs", languages: ["Scala", "TypeScript", "Go"] }` causes [another PR ](https://github.com/guardian/test-repocop-prs/pull/16)to be created, to integrate all of the languages 

## Potential future work

1) When we detect that an open PR already exists, we could consider adding a follow-up nudge or comment on the PR if some amount of time has passed since it opened?
2) If a new language is detected and a PR is open, we could consider closing the existing PR?

### DIFF
--- a/packages/snyk-integrator/src/index.ts
+++ b/packages/snyk-integrator/src/index.ts
@@ -1,13 +1,14 @@
-import { randomBytes } from 'crypto';
 import type { SNSHandler } from 'aws-lambda';
 import { parseEvent, stageAwareOctokit } from 'common/functions';
 import type { SnykIntegratorEvent } from 'common/src/types';
 import type { Config } from './config';
 import { getConfig } from './config';
 import { addPrToProject } from './projects-graphql';
+import { getPullRequest } from './snyk-integrator/pull-requests';
 import {
 	createSnykPullRequest,
 	createYaml,
+	generateBranchName,
 	generatePr,
 } from './snyk-integrator/snyk-integrator';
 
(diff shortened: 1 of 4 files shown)


akash1810:
> 1. We check for PR existence by searching the PRs for a given branch name.

IIUC the PR is raised by a specific user too? Should we also filter against the PR author? Though unlikely, a human user could create a branch with this app's naming scheme.

chrislomaxjones:
> IIUC the PR is raised by a specific user too? Should we also filter against the PR author? Though unlikely, a human user could create a branch with this app's naming scheme. Filtering against the author also allows us to change the naming scheme in future, if necessary.

I'd thought about this, and was banking on the situation being highly unlikely, but there's no reason not to make this more robust.

I'm little more unsure about the branch naming, as this proposed scheme allows us to de-dupe PRs by languages, so I may keep this scheme, at least for the time being!

NovemberTang (approved):
I think the PR author comment makes sense, but as it's an additional feature rather than a modification, I'll approve this one for the sake of keeping changes small

PR #721, opened 2024-01-23
Increase test coverage in common package

NovemberTang:
## What does this change?

- Removes the custom graphQL auth function, as it's just duplicating regular octokit logic (less code means less untested code!)
- Add unit tests to untested parts of the common package

## Why?
This package in particular is used by several parts of the service catalogue, and it's important that its behaviour is well-understood, -documented and -tested

## How has it been verified?

`npm run test -w common -- --coverage`

nb: I think that using jest rather than ts-jest has has confused the coverage tool, as some of the details about which lines are covered/uncovered don't add up (any advice on how to make this more accurate is welcome) - but the general idea is that coverage (especially of non-async code) has increased significantly.

### Before

File          | % Stmts | % Branch | % Funcs | % Lines | 
--------------|---------|----------|---------|---------|
All files     |      68 |    53.57 |      64 |   67.14 |
 functions.ts |   52.52 |     37.5 |   51.42 |   52.12 |
 logs.ts      |     100 |    76.92 |     100 |     100 | 
 string.ts    |   95.65 |    72.72 |   85.71 |   95.23 | 

### After

File          | % Stmts | % Branch | % Funcs | % Lines | 
--------------|---------|----------|---------|---------|
All files     |    78.1 |    66.66 |      75 |   77.34 |
 functions.ts |   66.27 |    57.14 |   66.66 |   65.85 | 
 logs.ts      |     100 |    76.92 |     100 |     100 |
 string.ts    |   95.65 |    72.72 |   85.71 |   95.23 | 

## Additional notes

There's a lot of async functions in the service catalogue, so 100% test coverage is not achievable, but should we have a coverage target for each module? It might remind developers to write tests, and encourage writing code in a more testable, functional style.

### DIFF
--- a/packages/common/src/functions.test.ts
+++ b/packages/common/src/functions.test.ts
@@ -3,9 +3,11 @@ import {
 	anghammaradThreadKey,
 	branchProtectionCtas,
 	getEnvOrThrow,
+	getGithubAppSecret,
 	parseEvent,
 	parseSecretJson,
 	partition,
+	topicMonitoringProductionTagCtas,
 } from './functions';
 
 function isValidUrl(str: string) {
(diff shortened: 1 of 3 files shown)


packages/common/src/functions.test.ts
```diff
@@ -70,7 +94,42 @@ describe('getEnvOrThrow', () => {
...
+		process.env.GITHUB_APP_SECRET = secret;
+	});
+});
+
+describe('parseSecretJson', () => {
+	it('put the right values in the right fields, if the JSON string has the correct fields', () => {
+		const actual = parseSecretJson(
+			'{"appId": "myAppId", "base64PrivateKey": "aGVsbG8=", "clientId": "myClientId", "clientSecret": "myClientSecret", "installationId": "myInstallationId"}',
+		);
+		console.log(actual.strategyOptions);
```
chrislomaxjones, line 117: nit: is this logging still necessary?

```suggestion
```

packages/common/src/functions.test.ts
```diff
@@ -70,7 +94,42 @@ describe('getEnvOrThrow', () => {
...
 
 	it('should throw an error for a non-existing environment variable', () => {
-		expect(() => getEnvOrThrow('NON_EXISTING_VAR')).toThrow();
+		const err = 'Environment variable NON_EXISTING_VAR is not set.';
+		expect(() => getEnvOrThrow('NON_EXISTING_VAR')).toThrow(err);
+	});
+});
+
+describe('getGitHubAppSecret', () => {
+	it('should throw iff the GITHUB_APP_SECRET environment variable is not set', async () => {
```
AshCorr, line 103:
```suggestion
	it('should throw if the GITHUB_APP_SECRET environment variable is not set', async () => {
```

AshCorr (approved):

chrislomaxjones (approved):
👍

PR #723, opened 2024-01-23
ci(prisma): Ensure views directory is added

akash1810:
## What does this change?
Ensure the [directory created by Prisma for views](https://www.prisma.io/docs/orm/prisma-schema/data-model/views#the-views-directory) is tracked in VCS.

This directory gets re-created whenever we perform a [migration](https://github.com/guardian/service-catalogue/blob/main/docs/database-migrations.md), so add it for ease[^1], and check in CI that it's up-to-date.

The content has cross-over with the [`migrations` directory](https://github.com/search?q=repo%3Aguardian%2Fservice-catalogue+path%3Apackages%2Fcommon%2Fprisma%2Fmigrations%2F*view*%2Fmigration.sql&type=code), so it might turn out that we can `.gitignore` it...

## How has it been verified?
With untracked files, CI fails with this output:

image

[^1]: Ease, as in I can `git add .` on a feature branch again without fear of adding unrelated files 😄

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,3 +126,12 @@ jobs:
 
       - name: check schema.prisma file hasn't changed
         run: git diff --exit-code packages/common/prisma/schema.prisma
+
+      - name: check for untracked files
+        run: |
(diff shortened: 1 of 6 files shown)


chrislomaxjones (approved):
Think it makes sense to track these files in version control 👍

PR #724, opened 2024-01-23
Log if a repo contains a vulnerability older than it should

NovemberTang:
## What does this change?

- Pull all the first page of critical alerts from dependabot (oldest first)
- If a repo has old, unresolved critical or high vulnerabilities, log it.

## Why?

95% of dependabot alerts are not relevant as they pertain to old, decomissioned, or otherwise non-production repos. This feature will allow us to highlight to developers which repositories with vulnerabilities need addressing, focusing the noisy output from GitHub

## How has it been verified?

Deployed to CODE and saw expected behaviour in the logs

## Additional notes

Currently we do not collect dependabot data via CloudQuery as the job takes over 48 hours to complete (potentially much longer). We could probably collect it in a more timely way if we were able to provide the job with a list of repos we cared about. I've spoken to @guardian/devx-operations about this, and we have added this work to their backlog, but for now we are proceeding this way as it allows us to continue with our work this quarter without [yak-shaving](https://en.wiktionary.org/wiki/yak_shaving).

### DIFF
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -13,7 +13,9 @@ import {
 } from '../languages';
 import type {
 	AwsCloudFormationStack,
+	DependabotVulnResponse,
 	PartialAlert,
+	RepoAndAlerts,
 	RepoAndStack,
 	Repository,
 	TeamRepository,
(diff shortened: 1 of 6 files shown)


packages/repocop/src/rules/repository.ts
```diff
@@ -269,49 +270,78 @@ function findArchivedReposWithStacks(
...
+				owner: 'guardian',
+				repo: name,
+				per_page: 100,
+				severity: 'critical,high',
+				state: 'open',
+				sort: 'created',
+				direction: 'asc', //retrieve oldest vulnerabilities first
+			});
+
+		return alert.data.map((a) => a as PartialAlert);
```
chrislomaxjones, line 290: Probably don't need the `map` here?

```suggestion
		return alert.data as PartialAlert[];
```

Or potentially, given that the full and partial alert types overlap sufficiently, no assertion at all?

```suggestion
		return alert.data;
```

packages/repocop/src/types.ts
```diff
@@ -63,7 +63,20 @@ export interface TeamRepository extends TeamRepositoryFields {
 	role_name: NonNullable<TeamRepositoryFields['role_name']>;
 }
 
-type DependabotVulnResponse =
+export type DependabotVulnResponse =
 	Endpoints['GET /repos/{owner}/{repo}/dependabot/alerts']['response'];
 
 export type Alert = DependabotVulnResponse['data'][number];
+
+export type PartialAlert = Pick<
```
NovemberTang, line 71: Taking suggestions for a new name for this type

packages/repocop/src/utils.test.ts
```diff
@@ -0,0 +1,25 @@
+import type { Repository } from './types';
+import { isProduction } from './utils';
+
+describe('test', () => {
+	test('pass', () => {
```
chrislomaxjones, line 5: maybe...

```suggestion
describe('isProduction', () => {
	test('should return correct values for prod and non-prod repos', () => {
```

packages/repocop/src/types.ts
chrislomaxjones: If the `Alert` type isn't used anywhere else it could be inlined. That'd free up calling it `Alert`.

```typescript
export type Alert = Pick<
	DependabotVulnResponse['data'][number],
	'state' | 'dependency' | 'security_vulnerability' | 'created_at'
>;
```

PR #730, opened 2024-01-24
Add aws plugin and get credentials via policy arn

JuliaBrigitte:
## What does this change?
We added the steampipe aws plugin and gave a policy read-only

## Why?
We want to query aws tables


akash1810 (approved):
Works great on CODE 🎉. Let's make it multi-account in a follow-up PR.

PR #731, opened 2024-01-25
Use EFS for Steampipe

AshCorr:


akash1810:
To double-check my understanding:
- Is this an alternative to using batch mode, outputting to a file, and ingesting that file into Postgres?
- The query behind the materialised views will need to be specific, rather than generic?

akash1810 (approved):
Interested to see how this performs, regarding SQL performance, deployment, and maintenance.

PR #733, opened 2024-01-25
Allow repocop to access the snyk secret

NovemberTang:
## What does this change?

Allows repocop to read the contents of the snyk secret so it can retrieve the data that used to populate the `snyk_projects` table. Logs the group ID to the console to demonstrate it was retrieved successfully.

## Why?

So we can retrieve Snyk Projects data using repocop

## How has it been verified?

Local execution works.
Verified group ID is logged on CODE


chrislomaxjones (approved):
Looks good 👍

PR #734, opened 2024-01-26
refactor(prisma): Simplify Prisma integration

akash1810:
Ultimately, this change aims to simplify how we use CloudQuery, and Prisma locally by making them cover only the tables we'll interact with in RepoCop, or data-audit, or the SQL views.

Our Prisma schema is big (~16,000 lines). IntelliJ, with the Prisma plugin, struggles to edit the file in a response way[^1]. It being difficult to edit, makes it harder to experiment with. For example testing [multi schema support](https://www.prisma.io/docs/orm/prisma-schema/data-model/multi-schema) as part of the Steampipe investigation.

The schema file is a reflection of the tables and views in the database. Locally, we're currently collecting every available table with CloudQuery, therefore the schema file is understandably large.

This PR changes how we run CloudQuery locally, to collect exactly the tables needed by our applications (RepoCop, and data-audit), and the SQL views. The resulting schema file is ~700 lines, which IDEs are happier with.

This change does not prevent CODE and PROD from collecting more tables. Indeed, they already are as they're collecting Riff-Raff, NS1, and Fastly data.

The Prisma migrations are also edited in this PR, specifically:
- The initial migration has been recreated, removing the now redundant tables
- A number of migrations have been merged, e.g. the `repocop_github_repository_rules` table is now a single migration vs. being split across multiple

This is because the migrations need to match the schema file. If we added a new migration, it would be to drop all the extra tables. This is undesired! Therefore, I've essentially started the migrations from scratch with a new `0_init` migration.

Deployment becomes a bit tricky as a result. We'd need to:
- [ ] Drop the `_prisma_migrations` table, effectively removing the old migrations
- [ ] Apply the new migrations via [resolve](https://www.prisma.io/docs/orm/reference/prisma-cli-reference#migrate-resolve). This'll mean we don't execute any SQL, as the tables, views, and functions already exist.

## How has it been verified?
I've successfully performed the migrations on CODE.

[^1]: eslint has also struggled with such a large schema file, see https://github.com/guardian/service-catalogue/pull/296.


packages/dev-environment/dev-config/cloudquery.yaml
```diff
@@ -33,46 +33,40 @@ spec:
...
+    - github_team_repositories
+  destinations: ['postgresql']
+  spec:
+    app_auth:
+      - org: guardian
+        private_key_path: /private-key.pem
+        app_id: '${file:/app-id}'
+        installation_id: '${file:/installation-id}'
+    orgs:
+      - guardian
```
akash1810, line 52: The other GitHub tables are scoped to a couple of repositories. Collecting `github_teams` and `github_team_repositories` with a repository restriction was resulting in no data, therefore collect them separately, scoped to the organisation (`orgs`).

packages/common/prisma/migrations/20230907105726_create_table_repocop_github_repository_rules/migration.sql
akash1810, line 1: The SQL in this file is a combination of [all the migrations we have performed](https://github.com/search?q=repo%3Aguardian%2Fservice-catalogue%20path%3A*.sql%20repocop_github_repository_rules&type=code).

packages/dev-environment/dev-config/cloudquery.yaml
```diff
@@ -82,28 +73,6 @@ spec:
...
-  destinations: ['postgresql']
-  tables:
-    - snyk_dependencies
-    - snyk_groups
-    - snyk_group_members
-    - snyk_integrations
-    - snyk_organizations
-    - snyk_organization_members
-    - snyk_organization_provisions
-    - snyk_reporting_issues
```
NovemberTang, line 98: @guardian/devx-security needs to hold on to `snyk-reporting-issues` and `snyk_reporting_latest-issues`, and possibly `snyk_groups` for work that's happening this quarter

akash1810: IIUC these tables are not used yet. Would suggest one of:
- Using them, and rebasing this branch off `main`
- Merging them in, and then adding these tables at point of usage

The latter would be a good test of the updated process.

akash1810: Spotted that https://github.com/guardian/service-catalogue/pull/754 started to query the `snyk_reporting_latest_issues` table. Branch has now been rebased accordingly.

tjsilver (approved):
LGTM 👍

akash1810:
> Deployment becomes a bit tricky as a result. We'd need to:
> 
> * [x]  Drop the `_prisma_migrations` table, effectively removing the old migrations
> * [x]  Apply the new migrations via [resolve](https://www.prisma.io/docs/orm/reference/prisma-cli-reference#migrate-resolve). This'll mean we don't execute any SQL, as the tables, views, and functions already exist.

Confirming this has been done on CODE and PROD now.

PR #735, opened 2024-01-26
Opt in service catalogue RDS databases to aws-backups

michaelwmcnamara:
## What does this change?
This change adds the "devx-backup-enabled":"true" tag to the service catalogue RDS database.

## Why?
This opts the database in to AWS-backup securing it against accidental or malicious deletion of backups of this database.
See the [FAQ's](https://docs.google.com/document/d/1VDCSxYFlWs4R6g0Waa6OmmfytV60AROyHxfIGho7cLA/edit?usp=sharing) for more information.

## How has it been verified?
Deploying change to CODE database and ensuring correct tags are applied and database backups transition correctly.

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -14,7 +14,7 @@ import {
 } from '@guardian/cdk/lib/constructs/ec2';
 import { GuardianPrivateNetworks } from '@guardian/private-infrastructure-config';
 import type { App } from 'aws-cdk-lib';
-import { Duration } from 'aws-cdk-lib';
+import { Duration, Tags } from 'aws-cdk-lib';
 import {
 	InstanceClass,
 	InstanceSize,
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/service-catalogue.ts
```diff
@@ -54,6 +54,7 @@ interface ServiceCatalogueProps extends GuStackProps {
 	 * @default true
 	 */
 	rdsDeletionProtection?: boolean;
+	withBackup: true;
```
jacobwinch, line 57: We are opting-in at this top level

packages/cdk/lib/service-catalogue.ts
```diff
@@ -110,6 +111,8 @@ export class ServiceCatalogue extends GuStack {
 
 		const db = new DatabaseInstance(this, 'PostgresInstance1', dbProps);
 
+		Tags.of(db).add("devx-backup-enabled", "false");
```
jacobwinch, line 114: ...but opting-out here.

I think we need to drop one of these.

PR #736, opened 2024-01-26
Opt in service catalogue RDS databases to aws-backups 

michaelwmcnamara:
## What does this change?
This change adds the "devx-backup-enabled":"true" tag to the service catalogue RDS database.

## Why?
This opts the database in to AWS-backup securing it against accidental or malicious deletion of backups of this database.
See the [FAQ's](https://docs.google.com/document/d/1VDCSxYFlWs4R6g0Waa6OmmfytV60AROyHxfIGho7cLA/edit?usp=sharing) for more information.

## How has it been verified?
Deploying change to CODE database and ensuring correct tags are applied and database backups transition correctly.


jacobwinch (approved):
This looks good to me - I'd suggest merging on Monday due to the timing of the DB modification 👍

packages/cdk/lib/service-catalogue.ts
```diff
@@ -127,6 +127,8 @@ export class ServiceCatalogue extends GuStack {
 
 		const db = new DatabaseInstance(this, 'PostgresInstance1', dbProps);
 
+		Tags.of(db).add('devx-backup-enabled', 'true');
```
akash1810, line 130: Did you consider setting [`withBackup`](https://github.com/guardian/cdk/blob/5df089eab51cb2eff3a42b3a2aaa34f9a3464902/src/constructs/core/stack.ts#L46-L53) to `true` at the `GuStack` level? It removes needing to know the name of the tag.

akash1810: UPDATE: After a chat with @jacobwinch it sounds like the `withBackup` prop will soon be removed, in favour of a RDS construct. That is, my comment can be ignored.

PR #738, opened 2024-01-26
fix: Use correct SSM parameter for Galaxies bucket

akash1810:
## What does this change?
Whilst the SSM Parameter `/INFRA/deploy/services-api/galaxies-bucket-name` exists, it's value is incorrect! Use the [same SSM Parameter used in CODE/PROD](https://github.com/guardian/service-catalogue/blob/b26cd77d01cc2caaf6d824929e8986ed824f2471/packages/cdk/lib/cloudquery/index.ts#L393-L401) to locate the Galaxies S3 bucket.

## How has it been verified?
- Collect Galaxies data locally on `main`
- Observe the database tables being empty
- Collect Galaxies data on this branch
- Observe the database tables having data


AshCorr (approved):
Looks good!

Bit of an odd SSM parameter!

akash1810:
> Bit of an odd SSM parameter!

Agreed! It originates from https://github.com/guardian/actions-static-site, which is a generic service.

PR #749, opened 2024-01-30
Pull project data from Snyk API

NovemberTang:
## What does this change?

Pulls basic information about Snyk projects from the Snyk Rest API. Crucially, this includes the project tags, so that we can see which repos are on snyk. For now, I'm not collecting the commit hash due to data staleness issues. (the snyk value will be totally up to date, but the information cloudquery has could be a day or even a week out of date, creating the false impression that Snyk and GitHub are out of sync).

This work swaps one workaround for a slightly more accurate workaround, until we have time to fix the CQ plugin, or build a Steampipe one.

## Why?

Currently we are using a proxy signal to tell if a repo is on snyk. This is the presence of a snyk workflow file. This is less accurate than calling the snyk projects api directly, as we do not know whether the workflow is active or not, what branches it's running on, etc

## How has it been verified?

Local execution succeeds.
CODE run succeeds.

## How to read

There's a lot of changes here, so it's probably easier to go through this PR commit by commit.


packages/repocop/src/utils.test.ts
```diff
@@ -23,3 +23,16 @@ describe('isProduction', () => {
 		expect(isProduction(nonProdRepo)).toBe(false);
 	});
 });
+
+describe('Failure on empty arrays', () => {
```
NovemberTang, line 27: This has been moved from the deleted file `query.test.ts`

packages/repocop/src/utils.ts
```diff
@@ -1,5 +1,39 @@
...
+export function toNonEmptyArray<T>(value: T[]): NonEmptyArray<T> {
+	if (value.length === 0) {
+		throw new Error(`Expected a non-empty array. Source table may be empty.`);
+	}
+	return value as NonEmptyArray<T>;
+}
+
+//TODO test this
+export class SetWithContentEquality<T> {
+	private items: T[] = [];
```
chrislomaxjones, line 16: Nice! You might want to use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) as the underlying data structure here. We wouldn't have to ensure uniqueness of elements ourselves then.

packages/repocop/src/utils.ts
```diff
@@ -1,5 +1,39 @@
...
 }
+
+export function toNonEmptyArray<T>(value: T[]): NonEmptyArray<T> {
+	if (value.length === 0) {
+		throw new Error(`Expected a non-empty array. Source table may be empty.`);
+	}
+	return value as NonEmptyArray<T>;
+}
+
+//TODO test this
```
chrislomaxjones, line 14: There's tests in this PR right?

```suggestion
//TODO test this
```

packages/repocop/src/utils.ts
```diff
@@ -1,5 +1,39 @@
...
+export class SetWithContentEquality<T> {
```
adamnfish, line 14: I'm not a mad fan of this. Can we make an immutable version at least? Alterantively, if we're going to create a type to model the behaviour we're after, can we model it directly rather than through the proxy of a set with custom equality?

packages/repocop/src/index.ts
```diff
@@ -48,10 +55,38 @@ async function writeEvaluationTable(
...
+		.flat()
+		.map((x) => x.attributes.tags)
+		.map(toGuardianSnykTags)
+		.filter((x) => !!x.repo && !!x.branch);
+
+	const tagsWithContentEquality = new SetWithContentEquality<GuardianSnykTags>(
+		(t) => `${t.repo}-${t.branch}`,
+	);
+
+	allSnykTags.forEach((t) => tagsWithContentEquality.add(t));
```
adamnfish, line 84: As I mention elsewhere, can we remove duplicates ourselves rather than go through all the machinery of mutating this custom datastructure?

packages/repocop/src/query.ts
```diff
@@ -127,3 +126,58 @@ export async function getWorkflowFiles(
...
+		},
+	};
+}
+
+export async function getSnykOrgs(config: Config): Promise<SnykOrgResponse> {
+	const getOrgsUrl = `https://api.snyk.io/api/orgs?version=${config.snykApiVersion}`;
+	const resp = await get(getOrgsUrl, snykRequestOptions(config));
+	console.log('Status code: ', resp.statusCode);
+
+	const snykOrgResponse = JSON.parse(resp.body) as SnykOrgResponse;
```
chrislomaxjones, line 147: This is just a thought for the future, but asserting the JSON is a certain shape here could lead to some tricky bugs down the line if our `SnykOrgResponse` type diverges from the Snyk API. This might be fine if the API is versioned and doesn't change in the future. However, if not, we could think about using a library like [zod]( https://zod.dev/) to validate against a schema at runtime.

packages/repocop/src/utils.ts
NovemberTang: I tried using a Set, but it didn't deduplicate properly so had to create a custom class (see section four of [this page](https://howtodoinjava.com/typescript/sets/))

packages/repocop/src/utils.ts
NovemberTang: yes you're right!

packages/repocop/src/query.ts
NovemberTang: We are using a versioned API so we **should** be okay, and this change is only supposed to be in place until we get the projects table back (either via CQ or Steampipe), but I think we should definitely bear this in mind if this ends up being a long term fix

packages/repocop/src/utils.ts
chrislomaxjones: Oh yeah that makes total sense! [Perhaps a `Map` with string keys constructed via `getKey`](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwDYEMDOa4GVgwOoCWMAFgMIQB2MwVAogI4CuKSRAngDwAqAfHAN4AoAJBgoBAG4pqcIsAC2aAFxwAsijAc0McRQDmAGji84AXjgVgAdzUaAFAEoA3CLGTpwOHtwBpYGxU7OXkVLgczPm1dPRcRBEooxgQYaDtvGD8AuCDqEONw00idAn1woWFhEgI0ADp0zLMvX38XYQBfQREUABNunIVQhxUJCAJugRFheIptOABrf0aq2vr-fvlnSeWa4Nq0XDsFtiNgzfbO4WJ0dcGVACMICCRgFAoJiunZo6XiarrmtjrM7CKC4RhQN7bXY1K5oQ7+M4dERSJCMYBwobGADaAF13iCwRC4FiaqSoblaii0Rica0Om0gA)?

packages/repocop/src/query.ts
chrislomaxjones: > only supposed to be in place until we get the projects table back

Very good point!

packages/repocop/src/utils.ts
NovemberTang: I think I've figured out a simple-ish way to deduplicate without requiring any new types at all, let me know what you think

tjsilver (approved):
Looks good to me, and very useful to tide us over! 👍

PR #750, opened 2024-01-30
Remove the deregistrationDelay on Steampipes ECS service

AshCorr:
## What does this change?

Sets the deregistrationDelay on Steampipes NLB to 0, down from the default of 5 minutes.

## Why?

The deregistration delay controls how long the NLB should wait until terminating connections to a ECS Task thats being removed. Grafana keeps a pool of connections alive, meaning the NLB will wait the maximum delay.

The task is still considered to be active and count towards the desired task total, this means that new tasks can't be started until the current task has been deregistered. For Steampipe where we limit to a maximum of 1 task it means after a deployment is made it takes 5 minutes for the old task to stop before the new task can start.


tjsilver (approved):
Looks good 👍

PR #751, opened 2024-01-30
Remove extra directory in snyk-integrator package

chrislomaxjones:
## What does this change?

Remove the `snyk-integrator` directory inside the `src` directory of the snyk integrator package.

## Why?

Unnecessary nesting.

## How has it been verified?

CI Passes.


NovemberTang (commented):
Does this still deploy and run ok? It's possible that altering the file structure has altered the structure of the files we send to riff-raff, and the lambda could be pointing at a handler that has moved.

chrislomaxjones:
> Does this still deploy and run ok? 

Just checked this on CODE and seems to run fine! The handler isn't moved as part of this PR, and I think all of the source is bundled into one `index.js` file that's uploaded via Riff Raff

NovemberTang:
> > Does this still deploy and run ok?
> 
> Just checked this on CODE and seems to run fine! The handler isn't moved as part of this PR, and I think all of the source is bundled into one `index.js` file that's uploaded via Riff Raff

oh yeah that makes sense!!

NovemberTang (approved):

PR #754, opened 2024-01-30
Count up number of alerts from Snyk

NovemberTang:
## What does this change?

Calculates whether or not a repository has Snyk vulnerabilities older than the cut-off date

## Why?

So teams can be better informed about their vulnerability obligations

## How has it been verified?

Unit tests verify expected behaviour. Deploying to CODE, we can see that it is logging when it finds repos with out of date vulnerabilities.

Example:
```
INFO	Dependabot - <REDACTED>: has 19 alerts that need addressing
INFO	Snyk - <REDACTED>: has 5 issues that need addressing
```

## Next steps

Combine alerts from both formats using a common format


tjsilver (approved):
💯

PR #759, opened 2024-01-31
feat(steampipe): Configure for multi-AWS account

akash1810:
> [!Note]
> Easiest to review [commit by commit](https://github.com/guardian/service-catalogue/pull/759/commits).

## What does this change?
Configure Steampipe to [access all the accounts](https://hub.steampipe.io/plugins/turbot/aws#multi-account-connections) in the AWS organisation by generating a config file with an entry per AWS account returned by:

```bash
aws organizations list-accounts
```

> [!Important]
> This branch requires static (permanent) IAM credentials to run locally.

It's worth noting that to run this locally, AWS credentials vended by Janus DEV is not enough, as it does not allow `sts:AssumeRole`. Therefore, one currently has to create some static credentials for an IAM User. This user should have an inline policy as below, and set to the profile `service-catalogue-DEV`:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": "sts:AssumeRole",
            "Effect": "Allow",
            "Resource": "arn:aws:iam::*:role/cloudquery-access"
        },
        {
            "Effect": "Allow",
            "Action": "organizations:ListAccounts",
            "Resource": "*"
        }
    ]
}
```

I have created a user, and credentials, manually, I'll revoke the credentials in the coming days. In the next PR, I'll update the DEV environment to use only the deployTools account, and Janus DEV credentials, simplifying things.

## How has it been verified?
- Ran locally via `npm -w dev-environment start`
- This has been deployed to CODE, and the following query has been run in Grafana:
   ```sql
   select  arn
           , account_aliases
           , organization_id
           , organization_master_account_email
           , organization_master_account_id
   from aws.aws_account;
   ```

---
Co-authored by: @AshCorr, and @JuliaBrigitte.

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -38,6 +38,7 @@ import {
 } from 'aws-cdk-lib/aws-ssm';
 import { getCentralElkLink } from 'common/src/logs';
 import { addCloudqueryEcsCluster } from './cloudquery';
+import { cloudqueryAccess, listOrgsPolicy } from './cloudquery/policies';
 import { addDataAuditLambda } from './data-audit';
 import { InteractiveMonitor } from './interactive-monitor';
 import { Repocop } from './repocop';
@@ -262,7 +263,7 @@ export class ServiceCatalogue extends GuStack {
 		new SteampipeService(this, 'steampipe', {
 			app,
 			cluster,
-			policies: [],
+			policies: [listOrgsPolicy, cloudqueryAccess('*')],
 			secrets: {},
 			accessSecurityGroup: steampipeSecurityGroup,
 			domainName: steampipeDomainName,
--- a/packages/cdk/lib/steampipe/service.ts
+++ b/packages/cdk/lib/steampipe/service.ts
@@ -6,8 +6,9 @@ import {
 import { GuCname } from '@guardian/cdk/lib/constructs/dns';
 import { GuSecurityGroup } from '@guardian/cdk/lib/constructs/ec2';
 import { Duration } from 'aws-cdk-lib';
-import { Port } from 'aws-cdk-lib/aws-ec2';
 import type { ISecurityGroup } from 'aws-cdk-lib/aws-ec2';
+import { Port } from 'aws-cdk-lib/aws-ec2';
+import type { FargateServiceProps } from 'aws-cdk-lib/aws-ecs';
 import {
 	CpuArchitecture,
 	FargateService,
@@ -17,13 +18,12 @@ import {
 	LogDrivers,
 	Secret,
 } from 'aws-cdk-lib/aws-ecs';
-import type { FargateServiceProps } from 'aws-cdk-lib/aws-ecs';
 import { PerformanceMode, ThroughputMode } from 'aws-cdk-lib/aws-efs';
 import {
 	NetworkLoadBalancer,
 	Protocol,
 } from 'aws-cdk-lib/aws-elasticloadbalancingv2';
-import { Effect, ManagedPolicy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import { Effect, PolicyStatement } from 'aws-cdk-lib/aws-iam';
 import { RetentionDays } from 'aws-cdk-lib/aws-logs';
 import { Secret as SecretsManager } from 'aws-cdk-lib/aws-secretsmanager';
 import { Images } from '../cloudquery/images';
@@ -183,7 +183,9 @@ export class SteampipeService extends FargateService {
 					'github-token',
 				),
 			},
-			command: ['steampipe service start --foreground'],
+			command: [
+				'generate-aws-plugin-config; steampipe service start --foreground',
+			],
 			logging: fireLensLogDriver,
 			portMappings: [
 				{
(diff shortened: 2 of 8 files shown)


packages/dev-environment/steampipe/generate-aws-plugin-config
akash1810, line 1: This script is heavily inspired by [Steampipe's example](https://github.com/turbot/steampipe-samples/blob/main/all/aws-organizations-scripts/generate_config_for_cross_account_roles.sh).

packages/dev-environment/steampipe/generate-aws-plugin-config
```diff
@@ -0,0 +1,73 @@
...
+
+echo "starting to generate configuration (stage=$STAGE, aws-version=$AWS_VERSION)"
+
+AWS_CONFIG_FILE=$HOME/.aws/config
+STEAMPIPE_INSTALL_DIR=$HOME/.steampipe/
+STEAMPIPE_CONFIG_FILE=${STEAMPIPE_INSTALL_DIR}/config/aws.spc
+
+# The `cloudquery-access` role has not been provisioned in the Root account,
+# Do not configure Steampipe to use it, as it will always error with `GetCallerIdentity`.
+ROOT_ACCOUNT_NAME="Guardian Web Systems"
```
akash1810, line 13: The role is provisioned into each member account from https://github.com/guardian/aws-account-setup via Riff-Raff. Riff-Raff cannot deploy into the Root account, therefore the role has not be provisioned there.

packages/cdk/lib/service-catalogue.ts
```diff
@@ -262,7 +263,7 @@ export class ServiceCatalogue extends GuStack {
 		new SteampipeService(this, 'steampipe', {
 			app,
 			cluster,
-			policies: [],
+			policies: [listOrgsPolicy, cloudqueryAccess('*')],
```
akash1810, line 266: `cloudqueryAccess` isn't a great name now; `serviceCatalogueAccess` is more accurate. However, the IAM role is called `cloudquery-access`, so we'd want to change that at the same time. This is for a separate PR.

packages/cdk/lib/steampipe/service.ts
```diff
@@ -137,8 +137,8 @@ export class SteampipeService extends FargateService {
 		);
 
 		const task = new FargateTaskDefinition(scope, `${id}TaskDefinition`, {
-			memoryLimitMiB: 512,
-			cpu: 256,
+			memoryLimitMiB: 2048,
+			cpu: 1024,
```
akash1810, line 141: Multi AWS account queries were timing out on the current resources. A 4x increase was somewhat arbitrarily chosen.

<details><summary>CPU utilisation graph</summary>
<p>

image

</p>
</details>

packages/cdk/lib/steampipe/service.ts
```diff
@@ -194,6 +196,14 @@ export class SteampipeService extends FargateService {
 			entryPoint: ['/bin/sh', '-c'],
 		});
 
+		task.taskRole.addManagedPolicy(
+			ManagedPolicy.fromManagedPolicyArn(
+				scope,
+				'SteampipeReadOnlyPolicy',
+				'arn:aws:iam::aws:policy/ReadOnlyAccess',
+			),
+		);
```
AshCorr: Do we still need this? Presumably to read resources in the deployTools account we'll be assuming the `cloudquery-access` role?

packages/dev-environment/steampipe/Dockerfile
```diff
@@ -1,6 +1,16 @@
...
-ARG GITHUB_PLUGIN_VERSION
-ARG AWS_PLUGIN_VERSION
+ARG SP_GITHUB
+ARG SP_AWS
 
-RUN steampipe plugin install steampipe github@${GITHUB_PLUGIN_VERSION} aws@${AWS_PLUGIN_VERSION}
+RUN steampipe plugin install steampipe github@${SP_GITHUB} aws@${SP_AWS}
+
+# Install awscli (root user needed)
+USER 0
```
AshCorr, line 9: I think you could use `root` here instead of a magic number! As documented in https://steampipe.io/docs/managing/containers#running-a-steampipe-dashboard-using-a-derived-image

packages/cdk/lib/steampipe/service.ts
akash1810, line 205: This has just been moved about, but I think it's no longer needed. Will remove.

JuliaBrigitte (approved):

AshCorr (approved):
Awesome 👌

packages/cdk/lib/steampipe/service.ts
AshCorr: Looks like its been removed.

akash1810: Done.

akash1810: We do not. I've [removed](https://github.com/guardian/service-catalogue/pull/759#discussion_r1474825927) them.

packages/dev-environment/steampipe/Dockerfile
akash1810: Updated in https://github.com/guardian/service-catalogue/pull/759/commits/3bdb6355a76dca5b9cd1ab18f0d8b1f2959abed0.

packages/dev-environment/steampipe/generate-aws-plugin-config
```diff
@@ -0,0 +1,73 @@
...
+
+cat << EOF >> "$AWS_CONFIG_FILE"
+[default]
+region = eu-west-1
+
+EOF
+
+cat << EOF > "$STEAMPIPE_CONFIG_FILE"
+connection "aws" {
+  plugin = "aws@${AWS_VERSION}"
```
akash1810, line 26: When installing a specific plugin version, it's crucial that the config states the version too. See https://steampipe.io/docs/managing/plugins#installing-a-specific-version.

akash1810:
> In the next PR, I'll update the DEV environment to use only the deployTools account, and Janus DEV credentials, simplifying things.

https://github.com/guardian/service-catalogue/pull/766

PR #760, opened 2024-01-31
Assess vulnerability tracker integration using results from snyk API

NovemberTang:
## What does this change?

We are able to pull the tags for each project from the Snyk API, meaning we have a more reliable way of telling whether or not a repo is on snyk or not. This means we don't have to use the proxy signal of whether or not it has a workflow file with a particular name.

## Why?

Better accuracy

## How has it been verified?

CI passes

There's about a 10% difference in the number of untracked repos, which is about what I was expecting to see

### DIFF
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -1,7 +1,6 @@
 import type {
 	github_languages,
 	github_repository_branches,
-	github_workflows,
 	repocop_github_repository_rules,
 	snyk_projects,
 	snyk_reporting_latest_issues,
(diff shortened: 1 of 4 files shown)


packages/repocop/src/rules/repository.ts
```diff
@@ -142,15 +143,39 @@ export function parseSnykTags(snyk_projects: snyk_projects) {
 	return snykTags;
 }
 
+function toGuardianSnykTags(tags: ProjectTag[]): GuardianSnykTags {
```
NovemberTang, line 146: This has just moved over from `index.ts`

chrislomaxjones (approved):
Looks good!

PR #761, opened 2024-01-31
refactor: move Ruby and Rust into common supported languages

tjsilver:
## What does this change?

Moves Ruby and Rust into common supported languages, as they are [now supported by Dependabot](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph) as well as Snyk.

## Why?

To be more accurate!

## How has it been verified?

There is a dependency graph on the [self-assessment tool](https://github.com/guardian/self-assessment/network/dependencies) which lists Rust dependencies, and I enabled dependency graph on [the fork of the ios app](https://github.com/guardian/snyk-test-ios-live/network/dependencies) and this lists the Ruby dependencies.

### DIFF
--- a/packages/repocop/src/languages.ts
+++ b/packages/repocop/src/languages.ts
@@ -23,6 +23,8 @@ const commonSupportedLanguages = [
 	'Python',
 	'Jinja', // Jinja uses Python dependencies
 	'Mako', // Mako uses Python dependencies
+	'Ruby',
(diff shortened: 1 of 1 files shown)


chrislomaxjones (approved):
Nice!

NovemberTang (approved):

PR #762, opened 2024-02-01
create unified type for alerts

NovemberTang:
## What does this change?

Pulls some common fields out of Dependabot and Snyk alerts, and creates a `RepocopVulnerability` type to put them in so we can meaningfully compare them

This change _looks_ incredibly chonky, but more than half of it is just setting up test data, and refactoring unit tests

You may find it easier to step through this PR commit by commit, otherwise the changes may seem a bit random.

## Why?

This will allow us to send consistent messaging to teams about their out of date vulnerabilities

## How has it been verified?

- [x] Deploy and test on CODE
- [x] Unit test the functions that convert from platform specific types to `RepocopVulnerability`

## Next steps

Potentially include CVE/CWE numbers so that we can deduplicate alerts?


packages/repocop/src/rules/repository.ts
```diff
@@ -474,6 +461,41 @@ export function evaluateOneRepo(
...
+		alert_issue_date: alert.created_at,
+		vulnerable_version: alert.security_vulnerability.vulnerable_version_range,
+	};
+}
+
+export function snykAlertToRepocopVulnerability(
+	alert: snyk_reporting_latest_issues,
+): RepocopVulnerability {
+	alert;
+
```
chrislomaxjones, line 483: can we remove this?

```suggestion

```

packages/repocop/src/rules/repository.ts
```diff
@@ -361,17 +364,13 @@ function getProjectIssues(
 	);
 }
 
-export function hasOldSnykAlerts(
+export function findSnykAlerts(
```
chrislomaxjones, line 367: I don't really have a suggestion for improvement, but the naming of this function led me to think it'd return a Snyk specific alert type rather than `RepocopVulnerability[]`.

packages/repocop/src/types.ts
```diff
@@ -126,16 +121,31 @@ export interface GuardianSnykTags {
...
-	packageManager?: string;
+	package: string;
+	packageManager: string;
 	publicationTime: string; //or Date?
 }
+
+export interface RepocopVulnerability {
+	source: 'Dependabot' | 'Snyk';
+	open: boolean;
+	severity: string;
```
chrislomaxjones, line 144: Could the different severities be modelled as a string union `"high" | "critical" | ...`?

packages/repocop/src/rules/example-dependabot-alerts.ts
chrislomaxjones, line 1: Nice idea to grab these examples!

How would you feel about keeping this in something like a `test-data` directory?

packages/repocop/src/rules/repository.ts
NovemberTang: yep, i have no idea how that got there!

packages/repocop/src/rules/repository.ts
NovemberTang: renamed the function. It's a little clunky, but is more accurate.

chrislomaxjones (approved):
Looks great!

PR #763, opened 2024-02-01
refactor(steampipe): Move plugin versions to `.env` file

akash1810:
## What does this change?
Move the Steampipe plugin versions to the `.env` file, for consistency, but also to later make them available as env vars in other contexts.

## How has it been verified?
Ran the workflow to build and push the image to GitHub Packages - https://github.com/guardian/service-catalogue/actions/runs/7742341313.

### DIFF
--- a/.github/workflows/ecs-publish.yml
+++ b/.github/workflows/ecs-publish.yml
@@ -22,6 +22,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Read .env file
+        run: cat .env | grep SP_ >> $GITHUB_ENV
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ecs-publish.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Read .env file
+        run: cat .env | grep SP_ >> $GITHUB_ENV
```
akash1810, line 26: `grep` to:
1. Avoid removing any comments from the file
2. Only pull out the values that relate to Steampipe

.github/workflows/ecs-publish.yml
```diff
@@ -48,3 +51,6 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            GITHUB_PLUGIN_VERSION=${{ env.SP_GITHUB }}
+            AWS_PLUGIN_VERSION=${{ env.SP_AWS }}
```
AshCorr, line 56: Worth keeping the naming consistent for Ctrl+F purposes?

AshCorr (approved):
LGTM!

.github/workflows/ecs-publish.yml
akash1810: Ah! Good point! I see about rolling this into #759, which this PR relates to.

PR #768, opened 2024-02-02
Gather more information about vulnerabilities, and deduplicate

NovemberTang:
## What does this change?

1. Adds an `isPatched` field to tell us if a vulnerability can be resolved through a direct or indirect version bump
2. Deduplicate vulnerabilities based on CVE, keeping whichever alert had the highest severity
3. Return a list of vulnerabilities along with the evaluation results for each repo.
4. Only return runtime dependencies from Dependabot

## Why?

1. To give teams an idea of how much work is involved in mitigating a vulnerability, allowing them to prioritise their remediations
2. Make the vulnerability count more accurate, so we aren't double-counting vulnerabilities across two platforms
3. So we can keep track of vulnerabilities on a departmental level
4. Picking only runtime dependencies brings us in line with our existing Snyk policy. It also makes the number of vulnerabilities less overwhelming, and provides teams with some initial focus.

## How has it been verified?

Added unit tests, verified on CODE.

Output looks something like

```json
{
    "timestamp": "YYYY-MM-DDT09:00:00.000Z",
    "level": "WARN",
    "requestId": "<REDACTED>",
    "message": "Found <REDACTED> out of date critical vulnerabilities, of which <REDACTED> are patchable"
}
```

with a corresponding log entry for high vulnerabilities

## Effect of filtering on vulnerability count

Please see [ELK](https://logs.gutools.co.uk/s/devx/goto/9d66a180-c1be-11ee-b3e0-71dbc9710958) for the actual numbers, which I haven't quoted directly as they are somewhat sensitive information

### CVE-based deduplication

- High vulnerabilities - 19% reduction
- Critical vulnerabilities - 22% reduction

### CVE-based deduplication and exclude dev dependencies

- High vulnerabilities - 36% reduction
- Critical vulnerabilities - 54% reduction

## Next steps

- `repository.ts` is getting a little crowded, so I'd like to break this file up a bit, but that can be done as part of a separate PR
- Repocop outputs a LOT of logs. You can provide a log level in the lambda config so I think this would be a nice to have just `INFO` and above, and then figure out which logs are essential, and which can be relegated to `DEBUG`
- Talk to the wider team about how we communicate these vulnerabilities to teams
- Do we have a clear telling whether or not dependabot is enabled everywhere it's supposed to be? Otherwise we may be undercounting our vulnerabilities. How do we enforce this across the org?

### DIFF
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -473,6 +473,10 @@ export function evaluateOneRepo(
 export function dependabotAlertToRepocopVulnerability(
 	alert: Alert,
 ): RepocopVulnerability {
+	const CVEs = alert.security_advisory.identifiers
+		.filter((i) => i.type === 'CVE')
+		.map((i) => i.value);
+
 	return {
 		open: alert.state === 'open',
 		source: 'Dependabot',
@@ -482,6 +486,7 @@ export function dependabotAlertToRepocopVulnerability(
 		ecosystem: alert.security_vulnerability.package.ecosystem,
 		alert_issue_date: alert.created_at,
 		isPatchable: !!alert.security_vulnerability.first_patched_version,
+		CVEs,
 	};
 }
 
@@ -500,6 +505,7 @@ export function snykAlertToRepocopVulnerability(
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- this is never null in reality
 		alert_issue_date: alert.introduced_date!,
 		isPatchable: issue.isPatchable || issue.isUpgradable || issue.isPinnable,
+		CVEs: issue.Identifiers.CVE,
 	};
 }
 
--- a/packages/repocop/src/rules/repository.ts
+++ b/packages/repocop/src/rules/repository.ts
@@ -455,8 +455,7 @@ export function deduplicateVulnerabilities(
 			return acc;
 		}, {});
 
-	const x: RepocopVulnerability[] = Object.values(dedupedWithCVEs);
-	const dedupedVulns = x.concat(withoutCVEs);
+	const dedupedVulns = Object.values(dedupedWithCVEs).concat(withoutCVEs);
 	return dedupedVulns;
 }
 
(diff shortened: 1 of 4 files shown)


packages/repocop/src/rules/repository.ts
```diff
@@ -303,7 +305,10 @@ export async function getAlertsForRepo(
 				direction: 'asc', //retrieve oldest vulnerabilities first
 			});
 
-		return alert.data;
+		const openRuntimeDependencies = alert.data.filter(
+			(a) => a.dependency.scope === 'runtime',
```
NovemberTang, line 309:
```suggestion
			(a) => a.dependency.scope !== 'development',
```

as this value could be `null` or `undefined`, we should try to catch those values

packages/repocop/src/rules/repository.ts
```diff
@@ -305,7 +305,10 @@ export async function getAlertsForRepo(
 				direction: 'asc', //retrieve oldest vulnerabilities first
 			});
 
-		return alert.data;
+		const openRuntimeDependencies = alert.data.filter(
+			(a) => a.dependency.scope !== 'development',
```
NovemberTang, line 309: the dependency scope could be `null`, or `undefined`, so only exclude specifically development ones until we have a better idea of how GH detects scope

packages/repocop/src/rules/repository.ts
```diff
@@ -422,6 +427,39 @@ export function testExperimentalRepocopFeatures(
...
+		.reduce<Record<string, RepocopVulnerability>>((acc, vuln) => {
+			const key = vuln.CVEs.join(',');
+			if (!acc[key]) {
+				acc[key] = vuln;
+			}
+			return acc;
+		}, {});
+
+	const x: RepocopVulnerability[] = Object.values(dedupedWithCVEs);
+	const dedupedVulns = x.concat(withoutCVEs);
```
chrislomaxjones, line 459: nit: could we inline `x` here?

```suggestion
	const dedupedVulns = Object.values(dedupedWithCVEs).concat(withoutCVEs);
```

chrislomaxjones (approved):
Nice one!

packages/repocop/src/rules/repository.ts
```diff
@@ -422,6 +427,39 @@ export function testExperimentalRepocopFeatures(
 	);
 }
 
+export function deduplicateVulnerabilities(
```
chrislomaxjones, line 430: Minor, but wonder if it's worth adding a comment or changing the function name to explicitly mention it's depduping by CVEs?

packages/repocop/src/rules/repository.ts
NovemberTang: whoops!

PR #778, opened 2024-02-08
Create a vulnerability digest for teams

NovemberTang:
## What does this change?

Create a digest for teams, so they can see up to 10 of their highest priority vulnerabilities that need addressing.

## Why?

Dependabot and snyk alerting is highly configurable. At the moment, we do not want to dictate how teams should go about patching their dependencies, however we do want to make them aware of production repos that have outstanding vulnerabilities, so that they can prioritise fixing them.

## How has it been verified?

Deployed to CODE

## TODO (for this PR)

- [ ] Indicate that these are only alerts for production repos, we do not surface others.
- [ ] Send alerts directly to teams rather than to a testing channel
- [ ] Filter out non-P+E teams, as they should have their own processes for vulnerability management
- [ ] Finish writing unit tests where indicated by comments

## TODO (for a different PR)

- Update recommendations repo to indicate what our obligations are around vulnerability resolution
- Link to the recommendations as a CTA in the message.


packages/repocop/src/vulnerability-digest.ts
```diff
@@ -0,0 +1,75 @@
...
+export function getTopVulns(vulnerabilities: RepocopVulnerability[]) {
+	return vulnerabilities
+		.sort(vulnSortPredicate)
+		.slice(0, 10)
+		.sort((v1, v2) => v1.fullName.localeCompare(v2.fullName));
+}
+
+function createHumanReadableVulnMessage(vuln: RepocopVulnerability): string {
+	const ecosystem =
+		vuln.ecosystem === 'maven' ? 'sbt or maven' : vuln.ecosystem;
```
tjsilver, line 37: Not sure what the distinction means here. Do we need to distinguish between sbt and maven?

NovemberTang: To me, maven implies a Java dependency and sbt implies a Scala dependency. I don't think a developer will always immediately recall that sbt uses maven dependencies under the hood, especially if they are quite junior, or new to Scala. I'm hoping that this replacement makes it clearer that this is quite likely to be a Scala vuln.

tjsilver: Ah right, that makes sense!

tjsilver (approved):
Before we roll this out it would be good to have a link to the soon-to-be recommendation/obligation page included in the digest.

PR #784, opened 2024-02-12
fix(iam): Create an explicit role w/deterministic name

akash1810:
## What does this change?
The [`cloudquery-access` role](https://github.com/guardian/aws-account-setup/blob/cfa18ed8b94b56223498f9227b0f32a1a7ea70d2/packages/cdk/lib/constructs/cloudquery-role.ts) is currently assumable by any role in the DeployTools account.

By having a deterministic IAM Role name, we can restrict the principal on the `cloudquery-access` role to tighten access.
This will be done as it's own PR, in the aws-account-setup repository - https://github.com/guardian/aws-account-setup/pull/166.

We can also set the [external id](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html) to further tighten access. Will do this in a follow-up PR.

## Why?
The `cloudquery-access` role deliberately breaks least responsibility; access should be tightly controlled.

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/bf3c8e54-75da-497b-9e87-564638f5af73), and successfully ran one of the tasks.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -392,7 +392,7 @@ spec:
         ],
         "TaskRoleArn": {
           "Fn::GetAtt": [
-            "CloudquerySourceAwsCostExplorerTaskDefinitionTaskRole41DB9C74",
+            "servicecatalogueTESTtaskAwsCostExplorer78777A06",
             "Arn",
           ],
         },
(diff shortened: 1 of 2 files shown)


AshCorr (approved):
LGTM

akash1810:
> Just one thing, might be my lack of AWS knowledge, but this role can't be assumed by ECS tasks outside of our account right?

Yep. IIUC if no [Principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html) is set, the Role is only accessible in the parent account.

PR #785, opened 2024-02-12
fix: Connect to Riff-Raff's Postgres with full ssl verification

akash1810:
## What does this change?
The commented issue has been resolved in https://github.com/guardian/deploy-tools-platform/pull/733 [^1].

[^1]: Riff-Raff's RDS is using the [same CA certificate as Service Catalogue](https://github.com/guardian/service-catalogue/blob/59cc777a6daa8927c5b7a83566c2a1904771fd10/packages/cdk/lib/service-catalogue.ts#L111-L127).


AshCorr (approved):
Fab :)

PR #793, opened 2024-02-14
Don't run DeployToolsListOrgs every minute at the start of a month

AshCorr:
## What does this change?

Updates the `minute` field to be `0` instead of `*`. This caused:

1. At 10AM the `DeployToolsListOrgs` task would be launched every minute until 11AM
2. Our stale data alerting incorrectly flagged the table as `DAILY` (despite it actually being `MONTHLY`) as the time between 1 invocation and the next was 1 minute.


chrislomaxjones (approved):
Good spot!

akash1810 (approved):
Ahhhh! Fantastic find!

img

PR #796, opened 2024-02-15
refactor(data-audit): Explicitly name the IAM Role

akash1810:
The data-audit lambda assumes the `cloudquery-access` role. This role's principal has been narrowed to a pattern. Update the IAM Role used by the lambda to match this pattern, so that it can assume the role.

See https://github.com/guardian/aws-account-setup/pull/166.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -14168,8 +14168,8 @@ spec:
     },
     "DataAudit2FEB3068": {
       "DependsOn": [
-        "DataAuditServiceRoleDefaultPolicyB6C67E52",
-        "DataAuditServiceRole35A15887",
+        "DataAuditRoleDefaultPolicyD0BF34E5",
+        "DataAuditRoleB3B90C40",
       ],
       "Properties": {
         "Code": {
@@ -14196,7 +14196,7 @@ spec:
         "MemorySize": 512,
         "Role": {
           "Fn::GetAtt": [
-            "DataAuditServiceRole35A15887",
+            "DataAuditRoleB3B90C40",
             "Arn",
           ],
         },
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/data-audit.ts
```diff
@@ -18,8 +19,46 @@ export function addDataAuditLambda(scope: GuStack, props: DataAuditProps) {
...
+		The lambda will be assuming the `cloudquery-access` role.
+		This role's principal has been narrowed to a pattern.
+
+		See https://github.com/guardian/aws-account-setup/pull/166.
+		 */
+		roleName: `service-catalogue-${app}-${stage}`,
+
+		/*
+		These managed policies do not meet AWS FSBP.
+		TODO remove these once GuCDK has improved - https://github.com/guardian/cdk/pull/2212.
```
akash1810, line 37: We could narrow down here, however fixing this upstream will have wider impact. Also, by continuing to use these managed policies, the diff in this PR is minimal.

packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
```diff
@@ -14292,31 +14292,10 @@ spec:
...
-                  "Ref": "AWS::Partition",
-                },
-                ":iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole",
-              ],
-            ],
-          },
+          "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole",
+          "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole",
         ],
+        "RoleName": "service-catalogue-data-audit-TEST",
```
akash1810, line 14298: This is the most meaningful change in this PR.

JuliaBrigitte: How likely is it that this would not be named -TEST forever?

akash1810: Not sure I understand the question...

The role name is produced via the template string `service-catalogue-${app}-${stage}`. `TEST` here is a reflection of the [tests using `TEST` as the `stage`](https://github.com/guardian/service-catalogue/blob/8ce3d494f1615080644f7ee3decced58bc78137b/packages/cdk/lib/service-catalogue.test.ts#L10).

AshCorr (approved):

PR #798, opened 2024-02-19
send notifications more teams, modify wording

NovemberTang:
## What does this change?

Sending out 15 messages only results a couple of teams with vulnerable production repos. let's widen our test group a little, especially as we don't know for sure that all those teams have anghammarad mappings. Let's double this number

Snyk API responses sometimes wrongly indicate that a vulnerability is not patchable. This is something we should address with Snyk directly, but for now, let's make the negative case more ambiguous.

## How has it been verified?

Deployed to CODE, confirmed we are now sending twice as many messages as before


tjsilver (approved):
👍

PR #805, opened 2024-02-20
Run DeployToolsListOrgs once a month, instead of once a year

AshCorr:
## What does this change?

This job was set to run on the 1st of January every year, Instead of the 1st of every month. This was causing our our table out of sync alarm to trigger.

## How has it been verified?

Before

image

After

image


akash1810 (approved):
💎

PR #808, opened 2024-02-22
Bump Clouquery Github plugin to v8

AshCorr:
## What does this change?

Bumps the Github plugin to v8.0.0

## Why?

The newest version of the Github plugin [supports syncing SBOM tables](https://github.com/cloudquery/cloudquery/pull/16796) from Github!

## How has it been verified?

Ran locally

### DIFF
--- a/.env
+++ b/.env
@@ -15,7 +15,7 @@ CQ_POSTGRES_SOURCE=3.0.7
 CQ_AWS=23.6.1
 
 # See https://hub.cloudquery.io/plugins/source/cloudquery/github/versions
-CQ_GITHUB=7.6.4
+CQ_GITHUB=8.0.0
 
 # See https://hub.cloudquery.io/plugins/source/cloudquery/fastly/versions
 CQ_FASTLY=3.0.7
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -4034,7 +4034,7 @@ spec:
             "Environment": [
               {
                 "Name": "GOMEMLIMIT",
-                "Value": "409MiB",
+                "Value": "819MiB",
               },
             ],
             "Essential": true,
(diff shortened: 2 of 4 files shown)


akash1810:
> The newest version of the Github plugin [supports syncing SBOM tables](https://github.com/cloudquery/cloudquery/pull/16796) from Github!

😍 amazing!

Unlike our AWS tasks, we don't have a "RemainingGithubData" task[^1] (due to rate limiting), so we'd need to add a new [task](https://github.com/guardian/service-catalogue/blob/71f91798a51ef170fe55e3794fe99a61947c0577/packages/cdk/lib/cloudquery/index.ts#L299-L366) to start collecting the new SBOM tables.

[^1]: `RemainingAwsData` defined [here](https://github.com/guardian/service-catalogue/blob/71f91798a51ef170fe55e3794fe99a61947c0577/packages/cdk/lib/cloudquery/index.ts#L234-L268)

AshCorr:
> Unlike our AWS tasks, we don't have a "RemainingGithubData" task[1](#user-content-fn-1-d13b14d18fb9990737a07cebe73f432e) (due to rate limiting), so we'd need to add a new [task](https://github.com/guardian/service-catalogue/blob/71f91798a51ef170fe55e3794fe99a61947c0577/packages/cdk/lib/cloudquery/index.ts#L299-L366) to start collecting the new SBOM tables.
> 
> ## Footnotes
> 1.

Ah yes, good point!

AshCorr:
Ah, I think the `GitHubRepositories` will actually collect the new sbom table as `github_repository_sboms` is a child... However how cloudquery does parent/child relationships does confuse me a little and maybe thats not true

akash1810:
> However how cloudquery does parent/child relationships does confuse me a little and maybe thats not true

Should we start using the [`skip_dependent_tables` config option](https://docs.cloudquery.io/docs/reference/source-spec#skip_dependent_tables) to simplify this?

AshCorr:
> > However how cloudquery does parent/child relationships does confuse me a little and maybe thats not true
> 
> Should we start using the [`skip_dependent_tables` config option](https://docs.cloudquery.io/docs/reference/source-spec#skip_dependent_tables) to simplify this?

Yes possibly, I'm not sure what kind of impact this will have on our rate limits though. If we have 2 tables "A", and "B" and we:

   1) Sync "A" with `skip_dependent_tables"
   2) Sync "B"

Does syncing "B" then trigger another sync for "A" because "B" depends on "A"? Maybe not if both are synced in the same task?

I think this has also highlighted a bug in our data staleness alarm. I don't think child tables are currently being checked for staleness as we rely on all tables being synced to be present in the `tables` property.

akash1810 (approved):
img

PR #809, opened 2024-02-22
Snyk works again!

AshCorr:
## What does this change?

CQ's Snyk Plugin is operational again. This mega-PR brings it back into CloudQuery, Prisma, and Repocop. Unfortunately, all of these changes need to be made at the same time, resulting in quite a large diff

We've also deleted view_snyk_project_tags as it relies on the schema of the old snyk plugin. We can bring it back, but this should be a separate PR.

## Why?

The Snyk API is a little flakey, so relying on it in real-time is a bit risky. Also, using the CQ table means we don't need to maintain any code that makes HTTP requests.

## How has it been verified?

- [ ] CI passes, unit tests preserve their behaviour
- [ ] Verify repocop works on CODE without access to Snyk secret
- [ ] CODE/PROD migrations work

## Next steps

Archive the guardian snyk cloudquery plugin


packages/common/prisma/schema.prisma
```diff
@@ -532,6 +532,40 @@ model github_languages {
   languages      String[]
 }
 
+model aws_lambda_functions {
```
NovemberTang, line 535: this has just moved up from its previous position. I don't know why. Seems harmless though

packages/common/prisma/schema.prisma
```diff
@@ -682,16 +686,6 @@ view view_running_instances {
   @@ignore
 }
 
-view view_snyk_project_tags {
```
NovemberTang, line 685: This has been removed as it relied on the schema of the old project table. If we still need it, we can bring it back as a subsequent PR.

packages/common/prisma/migrations/20240319101208_upgrade_snyk_plugin/migration.sql
```diff
@@ -0,0 +1,54 @@
...
+  - You are about to drop the column `name` on the `snyk_projects` table. All the data in the column will be lost.
+  - You are about to drop the column `org_id` on the `snyk_projects` table. All the data in the column will be lost.
+  - You are about to drop the column `origin` on the `snyk_projects` table. All the data in the column will be lost.
+  - You are about to drop the column `tags` on the `snyk_projects` table. All the data in the column will be lost.
+  - You are about to drop the `snyk_reporting_latest_issues` table. If the table is not empty, all the data it contains will be lost.
+
+*/
+
+-- DropTable
+DROP VIEW IF EXISTS "view_snyk_project_tags";
```
akash1810, line 14: We have a few [dashboards](https://metrics.gutools.co.uk/d/XgwFt8r4z/kpmg?orgId=1&editPanel=39) using this view. Are we comfortable breaking them? Or could we rebuild the view from the new tables?

packages/repocop/src/types.ts
```diff
@@ -41,6 +41,70 @@ export interface Repository extends RepositoryFields {
 	id: NonNullable<RepositoryFields['id']>;
 }
 
+export interface Coordinate {
```
akash1810, line 44: minor: I wonder if https://github.com/snyk-labs/snyk-api-ts-client might publish these types?

chrislomaxjones (approved):
Looks great! Had some minor suggestions below.

packages/repocop/src/query.ts
```diff
@@ -90,70 +86,32 @@ export async function getStacks(
...
 	client: PrismaClient,
-): Promise<snyk_reporting_latest_issues[]> {
-	return await client.snyk_reporting_latest_issues.findMany({});
+): Promise<SnykIssue[]> {
+	return (await client.snyk_issues.findMany({})).map((i) => {
+		return {
+			id: i.id,
+			attributes: i.attributes as unknown as SnykIssue['attributes'],
+			relationships: i.relationships as unknown as SnykIssue['relationships'],
+		};
```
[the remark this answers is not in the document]
tjsilver: Will do this in a follow-up PR.

PR #811, opened 2024-02-23
Refactor snyk integrator for maximum reusability

NovemberTang:
## What does this change?

- Move a lot of PR generation code to `common`, so that it can be reused when we generate PRs to integrate scala projects with dependabot
- Refactor the code so that it's more flexible, and can be used to generate an arbitrary PR that changes 1 file in 1 commit, and adds it to an arbitrary project board

The diff here is mostly changes to the lockfile, the code changes represent about 200 lines

## Why?

We will be rolling out a workflow file across all scala repos that submits scala dependencies to GitHub, so dependabot can keep an eye on its vulnerabilities. We'll be doing this in much the same way as the Snyk integrator, so want to make this code a little more reusable.

## How has it been verified?

- I've created a dependency-graph-integrator project in this repo, and verified that its PR generation works. PR generation for snyk integrator also still works. I've removed dependency-graph-integrator from this PR for now to keep it a reasonable size, and will put it back in a follow-up PR.
- Unit tests and CI still pass.
- Ran snyk integrator PROD from this commit, and you can see it is still generating the correct PR #813

### DIFF
--- a/packages/common/src/pull-requests.ts
+++ b/packages/common/src/pull-requests.ts
@@ -72,9 +72,7 @@ export async function getExistingPullRequest(
 	const found = pulls.filter((pull) => isGithubAuthor(pull, author));
 
 	if (found.length > 1) {
-		console.warn(
-			'More than one Snyk integrator PR found on repository - choosing the first.',
-		);
+		console.warn(`More than one PR found on ${repoName} - choosing the first.`);
 	}
 
 	return found[0];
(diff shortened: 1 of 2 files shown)


packages/common/src/pull-requests.ts
```diff
@@ -0,0 +1,138 @@
+import { randomBytes } from 'crypto';
```
NovemberTang, line 1: This file has just been moved from `packages/snyk-integrator/src/pull-requests.ts`. I'm not sure why GitHub hasn't picked that up

chrislomaxjones (approved):
Much neater! had two non-blocking comments.

PR #816, opened 2024-02-26
Create new dependency-graph-integrator project

NovemberTang:
## What does this change? Why?

- Creates a new package called dependency-graph-integrator that generates the YAML file and PR text required to send a repository's sbt dependencies to GitHub's dependency submission API. 
- Most of this PR is boilerplate. All the essential logic already exists in `common`, after being moved in #811 .
- The biggest design choice I made as part of this PR was the naming of the YAML file. Using `sbt-dependency-graph.yaml` means we are committing to one yaml file per language/build tool. This potentially means more files _if_ we decide to support more languages, but means less wrangling of existing files in that case, we simply add a new file and don't have to modify the existing one.
- Currently, it only runs locally. To keep changes manageable, AWS infrastructure and CI/CD will be set up in the next PR. As this is not currently running on PROD, and has no interaction with the GitHub API, the PR author and the board number have deliberately been left with placeholder values as they do not exist yet.
- The README is duplicates large parts of snyk-integrator's as I expect we will be retiring that project next quarter, but the context is important to hold on to.

## How has it been verified?

Added a unit test to verify that the YAML returns what we expect. Local execution produces the expected results

### DIFF
--- a/packages/dependency-graph-integrator/README.md
+++ b/packages/dependency-graph-integrator/README.md
@@ -1,0 +1,39 @@
+## Dependency graph integrator integrator
+
+### What does it do?
+
+The snyk integrator takes a repo name as JSON, like so:
+
+```json
+{
+	"name": "service-catalogue"
+}
+```
+
+It uses this input to generates a workflow file that allows us to submit sbt dependencies to GitHub for vulnerability monitoring.
+
+After creating the YAML, it raises a pull request on the named repository for teams to review. An automated solution saves teams time and effort expended into understanding how to configure the action, and means repos are integrated with Dependabot more quickly, and with fewer mistakes.
+
+### Why?
+
+#### Why track dependencies?
+
+The supply chain is an increasingly common attack vector for malicious actors. Tracking the dependencies we use - and resolving the vulnerabilities they introduce - allows us to minimise the probability of such an attack occurring, and the impact of such an attack.
+
+#### Why do we need this tool?
+
+Dependency management is a complex problem, and no tool is a complete solution. Dependabot (GitHub's in-house dependency management tool) works well for some languages, but not all. Crucially for P&E, Dependabot does not support scanning build.sbt files. This tool solves that problem.
+
+### Running on non-production environments
+
+In non-production environments, such as CODE, or when running locally, the dependency graph integrator will not create a PR. Instead, it will print the contents of the YAML file to the console. This allows developers to test the the core logic (creating the yaml file), without unintended side effects.
+
+#### CODE
+
+The lambda can safely be invoked in the CODE environment, as not only will it not attempt to create a PR, but there are no GitHub credentials available to it, attempting to instantiate a GitHub client will fail.
+
+The format of input to the lambda on the CODE environment is that of an SNS message. It's not trivial to construct, so some sample events have been provided in the Test tab of the lambda in the AWS console.
+
+#### DEV
+
+The lambda can be invoked locally by running `npm run start -w dependency-graph-integrator` from the root of the repo, or `npm run start` from the root of the snyk-integrator package. The input can be configured by modifying [this file](./src/run-locally.ts)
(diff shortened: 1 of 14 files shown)


packages/dependency-graph-integrator/README.md
```diff
@@ -0,0 +1,39 @@
+## Dependency graph integrator integrator
+
+### What does it do?
+
+The snyk integrator takes a repo name as JSON, like so:
```
tjsilver, line 5: dependency graph rather than snyk?

packages/dependency-graph-integrator/src/index.ts
```diff
@@ -0,0 +1,36 @@
+import type { SNSHandler } from 'aws-lambda';
+import { parseEvent } from 'common/functions';
+import {
+	createPrAndAddToProject,
+	generateBranchName,
+} from 'common/src/pull-requests';
+import type { DependencyGraphIntegratorEvent } from 'common/src/types';
+import type { Config } from './config';
+import { getConfig } from './config';
+import { createYaml, generatePrBody } from './snyk-integrator';
```
tjsilver, line 10: should we rename snyk-integrator, or put these in common?

packages/dependency-graph-integrator/src/snyk-integrator.test.ts
```diff
@@ -0,0 +1,24 @@
+import { createYaml } from './snyk-integrator';
```
tjsilver, line 1: again, should we move this function to common?

packages/dependency-graph-integrator/src/index.ts
NovemberTang: ooo good catch, they definitely need renaming as they are specific to this lambda

tjsilver (commented):
I've been thinking that maybe it would be good to keep a generic dependency graph submission config in `guardian/.github` like the Snyk one,  and refer to it in this one that gets deployed to many repos. That way we could add other language dependency submission actions to that as we bring them on board. It also means we can update the config in one place to exclude dev dependencies if we decide to in the future, or upgrade the version, rather than have to update every single yaml file created. WDYT?

NovemberTang:
> I've been thinking that maybe it would be good to keep a generic dependency graph submission config in `guardian/.github` like the Snyk one

I think this makes a lot of sense! I don't think it should block this PR from being merged though as 95% of this code would stay the same in that case. the only change we'd need to make here is to the generateYaml function, once the new workflow has been set up.

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -0,0 +1,68 @@
+import { markdownChecklist } from 'common/src/string';
+import { h2, p, tsMarkdown } from 'ts-markdown';
+import { stringify } from 'yaml';
+
+export function createYaml(prBranch: string): string {
+	const dependencyGraphWorkflowJson = {
+		name: 'Update Dependency Graph',
```
tjsilver, line 7: Should this refer specifically to SBT here as if we add other yaml files to the same repo, for example, the actions will have different names?

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -0,0 +1,68 @@
...
+		on: {
+			push: {
+				branches: ['main', prBranch],
+			},
+			workflow_dispatch: {}, //There isn't an elegant way to do this in TypeScript, so we'll remove the {} at the end
+		},
+		jobs: {
+			'dependency-graph': {
+				name: 'Update Dependency Graph',
```
tjsilver, line 16: As above

packages/dependency-graph-integrator/src/file-generator.ts
NovemberTang: Good shout! fixed!

packages/dependency-graph-integrator/src/file-generator.ts
NovemberTang: I don't think its needed here as it shows up as <ACTION_NAME>/<JOB_NAME>, so having sbt in the action name should be enough, but have amended the action name higher up

chrislomaxjones (approved):
Nice!

PR #817, opened 2024-02-26
Refactor vulnerability digest code

NovemberTang:
## What does this change?

The vulnerability digest flow looks something like this

1. Get all vulnerabilities for a repo from Snyk and Dependabot
2. Format them into a common type
3. Deduplicate the vulnerabilities
4. Return them as part of repocop's evaluation result.
5. Assemble the vulnerabilities into per-team vulnerability digests
6. Send the digests

This flow can be broken into two sections
Evaluation: Steps 1 - 4
Remediation/Action: Steps 5 - 6

Half of the code relating to steps 5 & 6 currently lives in the evaluation section of the code, and the other half lives in the root of the package. This PR moves it all into `remediation/vuln-digest`.

## Why?

Code organised in a more logical way is more readable and maintainable.

I've also renamed the `rules` folder `evaluation`, as it makes it clearer that the scope of the code in that folder should be to collect information. All code relating to specific actions should be in `remediation`

## How has it been verified?

We are just moving code around, so CI passing should be enough


tjsilver (approved):
This makes things a lot clearer!

PR #824, opened 2024-02-27
Check for branch protection before raising a PR

NovemberTang:
## What does this change?

Any part of the service catalogue using `createPrAndAddToProject` will now check that the repo has sufficient branch protections before raising a PR.

## Why?

These PRs need to be reviewed by teams as they have been raised by a bot. We make requests to GitHub to get this information. We can't rely on repocop to provide this information for two reasons:
1) Branch protection data could be out of date, especially if the job has failed recently
2) CloudQuery does not collect data about rulesets or custom properties

## How has it been verified?

This has been tested on PROD. #823 was generated as a result, as the repo does not have branch protection, but does have the ruleset applied as a consequence of its `production_status` property

### DIFF
--- a/packages/common/src/pull-requests.ts
+++ b/packages/common/src/pull-requests.ts
@@ -105,10 +105,12 @@ async function hasBranchProtection(
 
 		console.log(protection.data);
 
-		return (
+		const hasProtection =
 			protection.data.enabled === true &&
-			!!protection.data.required_pull_request_reviews
-		);
+			!!protection.data.required_pull_request_reviews;
+
+		console.log('has branch protection:', hasProtection);
+		return hasProtection;
 	} catch (e) {
 		console.error(e);
 		return false;
@@ -138,7 +140,12 @@ async function productionCustomProperty(
 			property.value === 'production',
 	);
 
-	return !!result;
+	console.log('Found property:', result);
+
+	const hasProductionStatus = !!result;
+	console.log('has production status:', hasProductionStatus);
+
+	return hasProductionStatus;
 }
 
 export async function createPrAndAddToProject(
--- a/packages/common/src/pull-requests.ts
+++ b/packages/common/src/pull-requests.ts
@@ -96,18 +96,23 @@ async function hasBranchProtection(
 	octokit: Octokit,
 	repoName: string,
 ): Promise<boolean> {
-	const protection = await octokit.rest.repos.getBranchProtection({
-		owner: 'guardian',
-		repo: repoName,
-		branch: 'main',
-	});
+	try {
+		const protection = await octokit.rest.repos.getBranchProtection({
+			owner: 'guardian',
+			repo: repoName,
+			branch: 'main',
+		});
 
-	console.log(protection.data);
+		console.log(protection.data);
 
-	return (
-		protection.data.enabled === true &&
-		!!protection.data.required_pull_request_reviews
-	);
+		return (
+			protection.data.enabled === true &&
+			!!protection.data.required_pull_request_reviews
+		);
+	} catch (e) {
+		console.error(e);
+		return false;
+	}
 }
 
 async function productionCustomProperty(
--- a/packages/common/src/pull-requests.ts
+++ b/packages/common/src/pull-requests.ts
@@ -112,7 +112,7 @@ async function hasBranchProtection(
 		console.log('has branch protection:', hasProtection);
 		return hasProtection;
 	} catch (e) {
-		console.error(e);
+		console.warn(e);
 		return false;
 	}
 }
@@ -132,7 +132,7 @@ async function productionCustomProperty(
 		},
 	);
 
-	console.log(allResults.data);
+	console.debug(allResults.data);
 
 	const result = allResults.data.find(
 		(property) =>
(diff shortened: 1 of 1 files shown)


packages/common/src/pull-requests.ts
```diff
@@ -90,47 +158,58 @@ export async function createPrAndAddToProject(
...
-			await addPrToProject(stage, repoName, boardNumber, author);
-			console.log('Updated project board');
-		} else {
-			console.log(
-				`Existing pull request found. Skipping creating a new one.`,
-				existingPullRequest.html_url,
-			);
-		}
-	} else {
-		console.log(`Testing generation of ${fileName} for ${repoName}`);
```
NovemberTang, line 129: this now exists as `logPr()`

packages/common/src/pull-requests.ts
```diff
@@ -78,6 +78,74 @@ export async function getExistingPullRequest(
...
+			repo: repoName,
+			headers: {
+				'X-GitHub-Api-Version': '2022-11-28',
+			},
+		},
+	);
+
+	console.debug(allResults.data);
+
+	const result = allResults.data.find(
```
NovemberTang, line 137: repos that contain the custom property `production_status : production` have a branch protection ruleset applied to them automatically

packages/common/src/pull-requests.ts
```diff
@@ -78,6 +78,74 @@ export async function getExistingPullRequest(
...
+			owner: 'guardian',
+			repo: repoName,
+			branch: 'main',
+		});
+
+		console.log(protection.data);
+
+		const hasProtection =
+			protection.data.enabled === true &&
+			!!protection.data.required_pull_request_reviews;
```
akash1810, line 110: Any reason to not get this from the database? There's a `protected` column in the [github_repository_branches table](https://hub.cloudquery.io/plugins/source/cloudquery/github/latest/tables/github_repository_branches).

packages/common/src/pull-requests.ts
```diff
@@ -78,6 +78,74 @@ export async function getExistingPullRequest(
...
+		return false;
+	}
+}
+
+async function productionCustomProperty(
+	octokit: Octokit,
+	repoName: string,
+): Promise<boolean> {
+	const allResults = await octokit.request(
+		'GET /repos/{owner}/{repo}/properties/values',
```
akash1810, line 125: Similar to above, any reason to not get this from the database? There's a `custom_properties` field in the [github_repositories table](https://hub.cloudquery.io/plugins/source/cloudquery/github/latest/tables/github_repositories).

packages/common/src/pull-requests.ts
NovemberTang: Defence in depth. It's possible that this is overkill but my reasoning is:

Ideally whoever is invoking this function would have made sure that the repo they're referring to has branch protection, but there's no guarantee of that. There's also no guarantee that the branch protection they have includes a review requirement. Basically, I'm trying to trust repocop as little as possible. (FWIW when repocop starts sending messages to this lambda I do plan to have it filter out unprotected repos, but someone using this a year from now might forget).

An alternative approach could be establishing a requirement to bundle prisma into any lambda that uses this function, and calling out to the DB at runtime, but I think that'll increase the setup complexity, hugely increase the zip size, make inline lambda editing impossible, is a bit tricky to document, so may not be worth it to grab two pieces of information.

Another (minor) point is that it's possible that a repo's branch protection has been changed between CQ running, and this lambda being invoked, and this code handles that edge case.

packages/common/src/pull-requests.ts
```diff
@@ -78,6 +78,74 @@ export async function getExistingPullRequest(
...
 }
 
+function logPr(
+	fileName: string,
+	repoName: string,
+	fileContents: string,
+	prTitle: string,
+	prBody: string,
+) {
+	console.log(`Testing generation of ${fileName} for ${repoName}`);
```
tjsilver, line 88: Does this line make sense for all uses of `logPr`?

NovemberTang: I thought that it did. Where do you think it doesn't make sense?

packages/common/src/pull-requests.ts
akash1810: Hmm, I'm not quite sure I understand how we're expecting this lambda to work... Are you suggesting it's manually invoked by a person every time? If so, I'd suggest we don't need AWS infrastructure here; a script that we run locally would simplify things.

I'd suggest we move to a setup where the lambda is invoked either on a schedule, or on an event (e.g. the CloudQuery task completing). This removes human error, and answers:

>  a repo's branch protection has been changed between CQ running, and this lambda being invoked

If the lambda does need to be invoked by a human from time to time, could we script it? The script would perform the pre-flight checks of ensuring branch protection is enabled, etc.

> make inline lambda editing impossible

I'd challenge the use-case for this. Editing the lambda function in place removes any audit trail!

As I said though, I'm not sure I understand the role/context of this lambda - happy to chat about it.

NovemberTang:
Reasons to close:
We're not currently doing anything dangerous. This PR is anticipating future "dangerous" PRs that we might raise, but we can avoid that by just not raising them.

To keep functions relatively single-purpose, it's probably best for the protection check to happen somewhere else.

PR #825, opened 2024-02-27
bring back dependabot vulnerabilities

NovemberTang:
## What does this change?

Bring back dependabot alerts

## Why?

We stopped collecting dependabot alerts while we were deciding what to do about yarn dependencies, as github does not differentiate between development and runtime deps. We decided to take no action, and encourage people to move away from yarn, in line with departmental recommendations.

## How has it been verified?

CI passes.
deploy to CODE and sense check the number of vulnerabilities


tjsilver (approved):
👍

chrislomaxjones (approved):

PR #826, opened 2024-02-27
refactor: use central dependency submission workflow

tjsilver:
## What does this change?

Following on from #816, this changes the dependency submission workflow from the third party sbt submission one to our new [central workflow](https://github.com/guardian/.github/blob/main/.github/workflows/dependency-graph.yml) in the `dependency-graph-integrator`.

## Why?

This is so that we can add other languages'/build tools' dependency submission workflows to our central one, and update it in one place.

## How has it been verified?
TBC

### DIFF
--- a/packages/dependency-graph-integrator/src/file-generator.test.ts
+++ b/packages/dependency-graph-integrator/src/file-generator.test.ts
@@ -14,7 +14,7 @@ jobs:
   dependency-graph:
     steps:
       - uses: actions/checkout@v4
-      - uses: scalacenter/sbt-dependency-submission@v2
+      - uses: guardian/.github/.github/workflows/dependency-graph.yml@main
     permissions:
       contents: write
 `;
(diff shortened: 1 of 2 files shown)


packages/dependency-graph-integrator/src/file-generator.test.ts
```diff
@@ -14,7 +14,7 @@ jobs:
   dependency-graph:
     steps:
       - uses: actions/checkout@v4
-      - uses: scalacenter/sbt-dependency-submission@v2
+      - uses: guardian/.github/.github/workflows/dependency-graph.yml@main
```
akash1810, line 17: We shouldn't use `main`. https://github.com/actions/toolkit/blob/master/docs/action-versioning.md provides a good justification as to why. TL;DR `main` could have breaking changes, and there isn't deterministic.

Whilst we can trust workflows we author, to a degree, it's a good habit to use tagged versions.

tjsilver: We had a chat about this IRL and agreed that it might be harder to keep repos up-to-date with the latest workflow if we use the SHA, say if repos haven't configured dependency updates in Dependabot - @akash1810 are you agreed that we can use `main` for now to ensure that repos are all up-to-date, considering it's a workflow that we author ourselves?

tjsilver:
I'm going to close this for now - and will resurrect if the team decides we should.

PR #830, opened 2024-03-01
log digests no matter what

NovemberTang:
## What does this change?

Always log digests

## Why?

So we can see what messages have been sent to teams on the day digests are sent

## How has it been verified?

CI passes


chrislomaxjones (approved):
👍

PR #831, opened 2024-03-04
Add Github team slug to `view_repo_ownership`

chrislomaxjones:
## What does this change?

This PR adds a new Prisma migration that drops and recreates the `view_repo_ownership` view.

The migration contains some breaking schema changes:
- Add column `github_team_slug`.
- Add column `short_repo_name`.
- Drop column `repo_name`.
- Renamed the `full_name` column to `full_repo_name`.

As some of these changes are breaking, this required changes in the Recocop application code, which are included as part of this PR.

> [!NOTE]
> Applying the migration without deploying the new Repocop application will break Repocop (as field names have changed). This shouldn't be a problem in reality, as Repocop runs on a schedule such that we should be able to migrate the database and deploy the new application between invocations.

## Why?

Making the team slug available in this view should ease sending notifications to Anghammarad, as we previously had to join this data with another table to retrieve team slugs.

## How has it been verified?

The dev migration was successfully applied locally. On approval, we'll also test this migration on CODE.

### DIFF
--- a/packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
+++ b/packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
@@ -21,7 +21,7 @@ function getOwningRepos(
 
 	const resultsOwnedByTeam = reposOwnedByTeam
 		.map((repo) => {
-			return results.find((result) => result.fullName === repo.full_name);
+			return results.find((result) => result.fullName === repo.full_repo_name);
 		})
 		.filter((result): result is EvaluationResult => result !== undefined);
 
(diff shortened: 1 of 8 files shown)


tjsilver (approved):
👍

packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
```diff
@@ -21,7 +21,7 @@ function getOwningRepos(
 
 	const resultsOwnedByTeam = reposOwnedByTeam
 		.map((repo) => {
-			return results.find((result) => result.fullName === repo.full_name);
+			return results.find((result) => result.fullName === repo.full_repo_name);
 		})
 		.filter((result): result is EvaluationResult => result !== undefined);
```
akash1810, line 26: minor, and not related to this PR - I think we can simplify this code into a single `filter`.

PR #838, opened 2024-03-07
fix: add Snyk API key ARN to setup script

tjsilver:
Co-authored-by: @NovemberTang 

## What does this change?

Adds a step to fetch the Snyk API key ARN and add this as an environment variable in the setup script.

## Why?

This is required to successfuly run  Repocop locally, since the environment variable was added to Repocop in #733.

## How has it been verified?

Ran the setup script after this change and observed repocop running without error.

## Note: 
A future piece of work is required to use only one SNYK env var.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -82,6 +82,8 @@ setup_environment() {
 
   SNYK_TOKEN=$(aws secretsmanager get-secret-value --secret-id /CODE/deploy/service-catalogue/snyk-credentials  --profile "$PROFILE" --region "$REGION" | jq -r '.SecretString' | jq -r '."api-key"' | tr -d "'")
 
+  SNYK_API_KEY_ARN=$(aws secretsmanager list-secrets --profile "$PROFILE" --region "$REGION" --query 'SecretList[*]['ARN']' --output text | grep /CODE/deploy/service-catalogue/snyk-credentials)
+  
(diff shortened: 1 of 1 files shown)


scripts/setup.sh
```diff
@@ -82,6 +82,8 @@ setup_environment() {
 
   SNYK_TOKEN=$(aws secretsmanager get-secret-value --secret-id /CODE/deploy/service-catalogue/snyk-credentials  --profile "$PROFILE" --region "$REGION" | jq -r '.SecretString' | jq -r '."api-key"' | tr -d "'")
 
+  SNYK_API_KEY_ARN=$(aws secretsmanager list-secrets --profile "$PROFILE" --region "$REGION" --query 'SecretList[*]['ARN']' --output text | grep /CODE/deploy/service-catalogue/snyk-credentials)
```
akash1810, line 85: Do we need both `SNYK_TOKEN` above and `SNYK_API_KEY_ARN` in this script? They both seem to be pulling from the same SSM Parameter 🤔.

tjsilver: You're right. The `SNYK_TOKEN` is used by other lambdas so I was going on the assumption that to use just one parameter across the board would require a larger refactor - this was just to get Repocop running locally again, and perhaps the refactor could be done separately? Correct me if I'm wrong though and you see a quick fix!

akash1810 (approved):

PR #839, opened 2024-03-11
remove redundant table from repocop

NovemberTang:
## What does this change?

All the repo ownership information we need comes from `view_repo_ownership`, so we no longer need to load the `TeamRepository` table

## Why?

Efficiency. Fewer DB calls, less memory usage

## How has it been verified?

CI passing should be sufficient to indicate repocop's behaviour hasn't changed


NovemberTang:
> Does the table also need to be removed from the prisma schema and cloudquery?

Cloudquery: I think it's still a useful table to keep in cloudquery as the information is generally handy to developers across the department. For example - it contains information about all levels of permissions, not just admins.

Prisma: The prisma schema/client is a manageable size at the moment, so I'm not too worried about the impact of one table. Plus dealing with prisma is a bit of pain right now so it may be easier to come back to this when things are a bit easier to work with. Either way, I think this could be done in a standalone PR.

akash1810:
> Plus dealing with prisma is a bit of pain right now so it may be easier to come back to this when things are a bit easier to work with.

Tell me more 👂🏽.

chrislomaxjones (approved):
Makes sense! Happy that we tackle any Prisma-related work separately.

tjsilver (approved):
👍

PR #840, opened 2024-03-11
Provision infra for the Dependency Graph Integrator

chrislomaxjones:
## What does this change?

This change adds additional CDK to create some new resources:
- A lambda that will run the dependency-graph-integrator script.
- A SNS topic from which Repocop can publish messages, and a lambda subscription so the dependency graph integrator lambda is triggered with these messages.
- In `PROD`/`TEST`, a secret for Github app credentials.

It also modifies our CI script and Riff Raff content directories to add a new lambda artifact, which we're already producing.

## Why?

So that we can begin running the `dependency-graph-integrator` in CODE and PROD stages, as opposed to just locally.

## How has it been verified?

CI passes and the snapshot appears to match what I'd roughly expect. I've also deployed the CFN and lambda changes to CODE, which passed successfully and created the new resources in AWS.

I tested the new CODE SNS topic and lambda by publishing messages via this command (using a suitable `$TOPIC_ARN`), and observing the lambda being invoked and logging to Cloudwatch as expected .

```bash
MESSAGE='{ "name": "test-repocop-prs" }'
aws sns publish \
    --topic-arn "$TOPIC_ARN" \
    --region eu-west-1 \
    --profile deployTools \
    --message "$MESSAGE" | jq .
```

The one thing I haven't been able to test, due to the lambda being stage-aware, is that the secret is successfully created and made available to the lambda. It should be okay to test this in prod, as the lambda doesn't consume it yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,8 @@ jobs:
               - packages/data-audit/dist/data-audit.zip
             snyk-integrator:
               - packages/snyk-integrator/dist/snyk-integrator.zip
+            dependency-graph-integrator:
+                - packages/dependency-graph-integrator/dist/dependency-graph-integrator.zip
             github-actions-usage:
               - packages/github-actions-usage/dist/github-actions-usage.zip
     env:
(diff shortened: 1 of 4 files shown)


NovemberTang (approved):
great! lgtm!

PR #841, opened 2024-03-11
Add `noEmit` option to the root TypeScript config

chrislomaxjones:
## What does this change?

Specify the `noEmit` TypeScript compiler option in the root tsconfig. This means we can also remove it from the `typecheck` package script.

## Why?

If I mistakenly run `npm run tsc` it won't emit JS files across the entire repository.

## How has it been verified?

CI passes, and running `npm run typecheck` doesn't emit JS locally.


AshCorr (approved):
Nice 👌

PR #852, opened 2024-03-18
Upload Prisma migrations to S3 bucket on deploy

chrislomaxjones:
## What does this change?

This PR adds a new artifact to the RiffRaff deployment, which gets directly copied into the account dist bucket. To achieve this, a change is made to the Riff Raff file (synthed in the CDK, requiring a change to `bin/cdk.ts`).

The steps are:
1. The CI script produces an additional `prisma.zip` file containing the contents of the Prisma directory.
2. This is added as a content directory to Riff-Raff, so it gets synced to the Riff Raff artifacts bucket.
3. When a deploy is triggered, this `prisma.zip` file is copied from the RiffRaff artifacts bucket to account dist bucket.
  
Subsequent work will then focus on running a migration within a task, triggered by a `prisma.zip` being PUT into the bucket.

## Why?

This is the first PR for a series of steps that will allow us to run automated Prisma migrations every time we deploy the service-catalogue, preventing the need for manual connections to RDS from a laptop for this purpose.

## How has it been verified?

Testing by deploying to CODE and verifying that the prisma.zip file was successfully uploaded to the account dist bucket by RiffRaff.

### DIFF
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -36,7 +36,7 @@ deployments.set('service-catalogue-prisma-migrations', {
 	contentDirectory: 'prisma',
 	app: 'prisma-migrate-task',
 	parameters: {
-		cacheControl: 'max-age=0',
+		cacheControl: 'no-store',
 		publicReadAcl: false,
 	},
 	regions: new Set([region]),
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -31,18 +31,10 @@ const riffRaff = new RiffRaffYamlFile(app);
 
 const deployments = riffRaff.riffRaffYaml.deployments;
 
-/**
- * All cfn-based dependencies should be applied before this s3 deployment
- */
-const dependencies = [...deployments.entries()]
-	.filter(([, { type }]) => type === 'cloud-formation')
-	.map(([key]) => key);
-
 deployments.set('service-catalogue-prisma-migrations', {
 	type: 'aws-s3',
 	contentDirectory: 'prisma',
 	app: 'prisma-migrate-task',
-	dependencies,
 	parameters: {
 		cacheControl: 'max-age=0',
 		publicReadAcl: false,
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -1,23 +1,54 @@
 import 'source-map-support/register';
-import { GuRoot } from '@guardian/cdk/lib/constructs/root';
-import { Duration } from 'aws-cdk-lib';
+import { RiffRaffYamlFile } from '@guardian/cdk/lib/riff-raff-yaml-file';
+import { App, Duration } from 'aws-cdk-lib';
 import { Schedule } from 'aws-cdk-lib/aws-events';
 import { ServiceCatalogue } from '../lib/service-catalogue';
 
-const app = new GuRoot();
+const app = new App();
+
+const stack = 'deploy';
+const region = 'eu-west-1';
 
 new ServiceCatalogue(app, 'ServiceCatalogue-PROD', {
-	stack: 'deploy',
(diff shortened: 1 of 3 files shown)


packages/cdk/bin/cdk.ts
```diff
@@ -1,23 +1,60 @@
...
+
+deployments.set('prisma-migrations', {
+	type: 'aws-s3',
+	contentDirectory: 'prisma',
+	app: 'prisma-migrate-task',
+	dependencies,
+	parameters: {
+		bucketSsmKeyStageParam: {
+			CODE: '/CODE/deploy/prisma-migrate-task-bucket',
+			PROD: '/PROD/deploy/prisma-migrate-task-bucket',
```
akash1810, line 49: Could we create this with IaC too?

packages/cdk/bin/cdk.ts
```diff
@@ -1,23 +1,60 @@
...
+		},
+		prefixStage: false,
+		prefixStack: false,
+		publicReadAcl: false,
+		cacheControl: 'public, max-age=315360000, immutable',
```
akash1810, line 54: Why do the files need to be public?

packages/cdk/lib/prisma-migrate-task.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface PrismaMigrateTaskProps {}
+
+export function addPrismaMigrateTask(
+	scope: GuStack,
+	// eslint-disable-next-line no-empty-pattern -- This will be required later
+	{}: PrismaMigrateTaskProps,
+) {
+	const app = 'prisma-migrate-task';
+	new GuS3Bucket(scope, `prisma-bucket`, { app });
```
akash1810, line 12: Did we consider using the account's artifact bucket, at `/account/services/artifact.bucket`?

chrislomaxjones: Hah, I didn't realise this bucket existed. Might rework this PR a little bit in that case!

packages/cdk/bin/cdk.ts
chrislomaxjones: Screenshot 2024-03-18 at 15 41 01

I got an error when I didn't pass in the `publicReadAcl` and `cacheControl` parameters.

packages/cdk/bin/cdk.ts
chrislomaxjones: Have opted to use the deploy dist bucket, so these parameters are no longer required.

packages/cdk/lib/prisma-migrate-task.ts
chrislomaxjones: I've now reworked the PR to use this bucket, re-review welcome!

packages/cdk/bin/cdk.ts
```diff
@@ -1,23 +1,46 @@
...
+const riffRaff = new RiffRaffYamlFile(app);
+
+const deployments = riffRaff.riffRaffYaml.deployments;
+
+deployments.set('service-catalogue-prisma-migrations', {
+	type: 'aws-s3',
+	contentDirectory: 'prisma',
+	app: 'prisma-migrate-task',
+	parameters: {
+		cacheControl: 'max-age=0',
```
akash1810, line 39: [This property](https://riffraff.gutools.co.uk/docs/magenta-lib/types#awss3) accepts any value for the [cache-control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). A [`no-store`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#use_cases) value might be simpler here?

akash1810 (approved):

packages/cdk/bin/cdk.ts
chrislomaxjones: Nice one, have changed it!

PR #859, opened 2024-03-19
Add linting rule to prefer `.flatMap(...)` over `.map(...).flat()`

akash1810:
## What does this change?
Adds https://github.com/sindresorhus/eslint-plugin-unicorn to the linting rules, and configure the linter to force usage of `.flatMap` as a terser version of mapping then flattening.

`eslint-plugin-unicorn` has a number of rules, which we might want to enable in future, and/or add it to the [departmental eslint configuration](https://github.com/guardian/csnx/tree/main/libs/%40guardian/eslint-config).

## Why?
Terser code.

## How has it been verified?
N/A.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "esbuild-jest": "^0.5.0",
         "eslint": "^8.57.0",
         "eslint-plugin-prettier": "^5.1.3",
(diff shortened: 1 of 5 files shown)


NovemberTang (approved):
lgtm!

chrislomaxjones (approved):
Nice idea 👍

PR #860, opened 2024-03-20
Run service catalogue in Multi-AZ on prod

JuliaBrigitte:
## What does this change?
Run the service catalogue db as a Multi-AZ on prod

## Why?
This is the standard approach for PROD and we saw we did not at the moment when doing the reservations. So we reserved a Multi-AZ db instance now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13801,11 +13801,6 @@
         "@types/got": "^9.6.12"
       }
     },
-    "packages/reservations-checker": {
-      "name": "reservation-checker",
-      "version": "1.0.0",
-      "extraneous": true
-    },
     "packages/snyk-integrator": {
       "version": "1.0.0",
       "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -11143,10 +11143,6 @@
         "node": ">=0.10.0"
       }
     },
-    "node_modules/reservation-checker": {
-      "resolved": "packages/reservations-checker",
-      "link": true
-    },
     "node_modules/resolve": {
       "version": "1.22.8",
       "dev": true,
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -10682,6 +10682,10 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/reservation-checker": {
+      "resolved": "packages/reservations-checker",
+      "link": true
+    },
```
akash1810, line 10688: Not sure these changes relate to this PR?

packages/cdk/lib/service-catalogue.ts
```diff
@@ -101,11 +102,7 @@ export class ServiceCatalogue extends GuStack {
 			storageEncrypted: true,
 			securityGroups: [dbSecurityGroup],
 			deletionProtection: rdsDeletionProtection,
-
-			/*
-			This certificate supports automatic rotation.
-			See https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.RegionCertificateAuthorities
-			 */
```
akash1810, line 108: Any reason we're dropping this comment?

package-lock.json
```diff
@@ -13722,6 +13722,11 @@
         "@types/got": "^9.6.12"
       }
     },
+    "packages/reservations-checker": {
+      "name": "reservation-checker",
+      "version": "1.0.0",
+      "extraneous": true
+    },
```
akash1810, line 13729: Don't think these are needed for this change?

```suggestion

```

akash1810:
The `package-lock.json` has conflicts following some recent Dependabot PRs. Rather than resolving the conflicts, should we drop the edits to that file within this branch entirely? I don't think they're needed, technically.

package-lock.json
JuliaBrigitte: tried to remove package-lock.json from branch

akash1810 (approved):

PR #863, opened 2024-03-21
Upgrade version of CloudQuery GitHub plugin to 8.1.3

NovemberTang:
## What does this change?

```
[*] Changed the `github_repositories` table
  [+] Added column `custom_properties`
  [+] Added column `web_commit_signoff_required`

[*] Changed the `github_team_repositories` table
  [+] Added column `custom_properties`
  [+] Added column `web_commit_signoff_required`

[*] Changed the `github_workflows` table
  [+] Added column `contents_as_json`
```

## Why?

The `custom_properties` column will eventually replace our use of topics to indicate production status.

having github workflows contents_as_json makes them easier to parse, and means we can more safely check for usages of particular workflows, automatically update versions, etc

## How has it been verified?

Verified local behaviour. Will test on CODE once this change has been approved, before merging


chrislomaxjones (approved):
👍

akash1810:
> Will test on CODE once this change has been approved, before merging

I think PROD has been broken since this change. Curious to hear how testing on CODE went, as to my knowledge there isn't a significant difference between the two, and I'm at a loss as to what the cause might be otherwise.

See also https://github.com/guardian/service-catalogue/pull/889.

PR #864, opened 2024-03-21
feat: create repocop vulnerabilities table

tjsilver:
## What does this change?

This PR represents the smallest possible change to introduce the `repocop_vulnerabilities` table. Further PRs will be raised to optimise repocop's usage of this table.

- Adds  a new `repocop_vulnerabilities` prisma model
- Maps from the `RepocopVulnerability` type to this table and adds the owner (new field)
- Writes the data  to the `repocop_vulnerabilities` table in repocop
- Refactors the `RepocopVulnerability` type  to uses lower/snake case to make the mapping simpler.

## Why?

So we can view our vulnerabilities in a dashboard. 

## Next steps

- Refactor the code to use the prisma type only (remove the `RepocopVulnerability` type)
- Refactor the vulnerability digest to use the data from this table

## How has it been verified?

Code migration has been performed and the table is visible here

https://metrics.code.dev-gutools.co.uk/goto/BJ-Y1CJSg?orgId=1

### DIFF
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -102,8 +102,8 @@ export async function main() {
 		severityPredicate,
 	);
 
-	const highPatchable = high.filter((x) => x.isPatchable).length;
-	const criticalPatchable = critical.filter((x) => x.isPatchable).length;
+	const highPatchable = high.filter((x) => x.is_patchable).length;
+	const criticalPatchable = critical.filter((x) => x.is_patchable).length;
 
 	console.warn(
 		`Found ${high.length} out of date high vulnerabilities, of which ${highPatchable} are patchable`,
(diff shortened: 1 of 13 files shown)


packages/repocop/src/index.ts
```diff
@@ -49,6 +51,21 @@ async function writeEvaluationTable(
 	console.log('Finished writing to table');
 }
 
+async function writeVulnerabilitiesTable(
+	vulnerabilities: repocop_vulnerabilities[],
```
chrislomaxjones, line 55: The `as unknown as repocop_vulnerabilities[]` had me a little confused on L136, but I think that's because the type there has the additional `repo_owner` property, so could be avoided by changing the signature here to:

```suggestion
	vulnerabilities: Array<RepocopVulnerability & { repo_owner: string }>,
```

So below it'd become

```typescript
const vulnerabilities = evaluationResults
	.flatMap((result) => result.vulnerabilities)
	.flatMap((vuln) => combineVulnWithOwners(vuln, repoOwners));
await writeVulnerabilitiesTable(vulnerabilities, prisma);
```

chrislomaxjones (approved):
Great stuff!

PR #866, opened 2024-03-25
Add task definition for performing Prisma migrations

chrislomaxjones:


packages/cdk/lib/service-catalogue.ts
```diff
@@ -227,5 +247,13 @@ export class ServiceCatalogue extends GuStack {
 			db,
 			dbAccess: applicationToPostgresSecurityGroup,
 		});
+
+		addPrismaMigrateTask(this, {
+			loggingStreamName,
+			logShippingPolicy,
+			db,
+			dbAccess: applicationToPostgresSecurityGroup,
+			cluster: cloudqueryCluster,
```
chrislomaxjones, line 256: I chose to reuse the existing CloudQuery ECS cluster here, but we could always create a separate cluster for running this single task as well.

NovemberTang (approved):
🚀 🚀 🚀

PR #867, opened 2024-03-25
Copy parts of the CQ DB from CODE

NovemberTang:
## What does this change?

Introduces an alternative method of retrieving CQ data, by copying information directly from the CQ database

Currently this only pulls snyk and galaxies data, as these are the smallest datasets. More work may be required to narrow down the list of AWS/GitHub tables we are using.

## Why?

- Remove the need to store API keys etc on dev machines
- Faster startup
- Save money on CloudQuery fees by reducing the number of rows we sync

Currently the process for pulling accurate data is

1. Run the CQ docker container to retrieve all the data
2. Run the dev migration, removing all the data
3. Run the CQ docker container again.

At the end point of this work, the process will be

1. Run `npm -w dev-environment start`
2. That's it.

## How has it been verified?

Local execution works, the expected tables show up


chrislomaxjones (approved):
Looks great!

PR #874, opened 2024-03-26
Add Table Options to AWS Source to only ingest ACTIVE SecurityHub findings

louishather:
## What does this change?

This change adds table_options to the AWS source to ingest only the records which are ACTIVE. 

## Why?

Right now aws_securityhub_findings has 51000 active records and 914000 archived records. This is extremely slow to query, taking about 19.5 seconds. This number is also unlikely to decrease over time. The 914050 archived records (representing a little under 95% of entries aren't particularly interesting, as there will be a more up to date ACTIVE record, if the resource(s) in question still exist.

## Cost saving estimates
This table runs once a day. Ignoring the free rows quota, using the price of $15 per million rows, and the row counts stated above, this change will save us  about $14 a day, which over the course of a month, represents a saving of over $400 a month, or 27.5 million rows ($5000 over the course of a year).

NB these numbers only apply to `PROD`, but we expect to see similar results on a smaller scale across `DEV` and `CODE` environments
Its possible there are other tables where similar efficiency savings can be made.

## How has this been verified

CODE row count before: 968,157
CODE row count after: 51,732 (95% reduction)

Time to query before: 19.5 seconds
Time to query after: 2.5 seconds

Co-authored-by: @akash1810 @louishather @NovemberTang

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -742,7 +742,11 @@ spec:
         role_arn: arn:aws:iam::000000000015:role/cloudquery-access
     table_options:
       aws_securityhub_findings:
-        record_state: ACTIVE
+        get_findings:
+          - filters:
+              record_state:
+                - comparison: EQUALS
+                  value: ACTIVE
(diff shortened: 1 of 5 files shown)


adamnfish (approved):
Looks great!

PR #875, opened 2024-03-26
copy github data directly from DB

NovemberTang:
## What does this change?

Copy github data directly from the cloudquery DB.
The copy function uses wildcarding, meaning we're collecting a couple more tables than we were before, but only about 5 or 6. Until we're more specific about which tables we want, we should be careful when introspecting the database.

For now, the only data we are not gathering is AWS data. These tables tend to me much larger, and theres like 600 of them. This means we need to be really careful about what data we pull, and the wildcarding approach used so far might be a bit too greedy, and will be refined in the next PR

## Why?

- Save money on CQ rows
- Faster startup times

## How has it been verified?

Tables show up in DEV environment

### DIFF
--- a/packages/dev-environment/dev-config/cloudquery.yaml
+++ b/packages/dev-environment/dev-config/cloudquery.yaml
@@ -1,53 +1,3 @@
----
-kind: source
-spec:
-  # Source spec section
-  name: 'github'
-  path: 'cloudquery/github'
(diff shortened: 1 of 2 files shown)


chrislomaxjones (approved):
Makes sense!

Agree with your comment about the wildcarding, let's see how it fares once AWS tables are being pulled.

PR #876, opened 2024-03-26
Sync some AWS tables using pg_dump

NovemberTang:
### Note
This PR is more easily digestible in [split mode](https://github.com/guardian/service-catalogue/pull/876/files?diff=split&w=1)

## What does this change?

- Check for the `libpq` package before attempting to install it. Less noisy output, quicker execution if `libpq` is already installed
- Split up `copy` into two methods. One for wildcarding, one for pulling single tables. Better performance.
- Perform db migration before copying data over, so views show up from the beginning. Improves QOL - there is no need to pull data twice.
- Parallelise `copy` operations. Improves setup performance
- Copy some AWS tables over via `copy`. Much faster than waiting for cloudquery, grabs data from multiple accounts, no sync fees.

This script runs in about a minute: `npm run start -w dev-environment  7.59s user 5.29s system 21% cpu 1:00.08 total`

It pulls all the data required for repocop to run without needing to wait to collect the remaining CQ tables. I'm not sure what we're using them for, but I've left them in for now.

## What does this not change

Some amazon tables have been left for the cloudquery container to collect. This is because they throw errors when we try to transfer the data from the database directly. More investigation is required to work out the root cause, but this shouldn't stop us from moving forward with the ones that work, as this PR still provides a significant performance benefit compared to before. 

## How has it been verified?

- Ran locally, verified the tables look how we expect.
- Ran repocop locally, produced expected (though much richer) output

### DIFF
--- a/packages/dev-environment/script/start
+++ b/packages/dev-environment/script/start
@@ -107,8 +107,6 @@ fi
 
 step "Baselining DB"
 echo -e "${yellow}Please note that this will drop all tables in the local database${clear}"
-echo "Waiting 10 seconds before proceeding..." # Give the user a chance to cancel.
-sleep 10
 npm -w cli start migrate -- --stage DEV --confirm
 
 step "Copying data from CODE DB to local DB"
--- a/packages/dev-environment/script/start
+++ b/packages/dev-environment/script/start
@@ -81,6 +81,9 @@ get_db_password
 
 source "$MAIN_ENV_FILE"
 
+step "Baselining DB"
+npm -w cli start migrate -- --stage DEV --confirm
+
 copy "snyk"
 copy "galaxies"
 copy "github_team"
--- a/packages/dev-environment/script/start
+++ b/packages/dev-environment/script/start
@@ -9,10 +9,20 @@ MAIN_ENV_FILE=$ROOT_DIR/.env
 LOCAL_ENV_FILE=$HOME/.gu/service_catalogue/.env.local
 clear='\033[0m'
 cyan='\033[0;36m'
+yellow="\033[1;33m"
 
 step(){
   ((STEP_COUNT++))
   echo -e "\n${cyan}Step ${STEP_COUNT}${clear}: $1"
+}
+
+install_brew_formula() {
+    if ! brew list "$1" &>/dev/null; then
+        step "Installing $1"
+        brew install "$1"
+    else
+        echo "$1 is already installed"
+    fi
 }
 
 add_to_path ()
@@ -27,12 +37,12 @@ add_to_path ()
 
 setup_libpq() {
     step "Setting up libpq"
-    brew install libpq
+    install_brew_formula "libpq"
     add_to_path "/opt/homebrew/opt/libpq/bin"
 }
 
 get_db_password() {
-    step "Getting DB password"
+    step "Getting DB credentials"
     SECRET_NAME=$(aws secretsmanager list-secrets \
     --profile deployTools --region eu-west-1 \
     --filters Key=tag-value,Values=CODE  | jq '.SecretList[].Name' | grep Postgres | tr -d '"')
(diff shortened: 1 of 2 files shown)


packages/dev-environment/script/start
```diff
@@ -9,12 +9,22 @@ MAIN_ENV_FILE=$ROOT_DIR/.env
...
 clear='\033[0m'
 cyan='\033[0;36m'
+yellow="\033[1;33m"
 
 step(){
   ((STEP_COUNT++))
   echo -e "\n${cyan}Step ${STEP_COUNT}${clear}: $1"
 }
 
+install_brew_formula() {
```
akash1810, line 19: Could we list homebrew formulae in a [`Brewfile`](https://github.com/Homebrew/homebrew-bundle)? This'll allow us to install them all via `brew bundle`.

Here's an example - https://github.com/guardian/grid/blob/main/Brewfile.

packages/dev-environment/script/start
```diff
@@ -27,12 +37,12 @@ add_to_path ()
 
 setup_libpq() {
     step "Setting up libpq"
-    brew install libpq
+    install_brew_formula "libpq"
     add_to_path "/opt/homebrew/opt/libpq/bin"
```
akash1810, line 41: Appreciate this isn't added in this PR, however, does Homebrew not add `libpq` to the PATH by default?

packages/dev-environment/script/start
```diff
@@ -76,16 +105,27 @@ else
   exit 1
 fi
 
-setup_libpq
-get_db_password
+step "Baselining DB"
+echo -e "${yellow}Please note that this will drop all tables in the local database${clear}"
+echo "Waiting 10 seconds before proceeding..." # Give the user a chance to cancel.
+sleep 10
```
akash1810, line 111: Could this be a flag instead? The [migration CLI tool](https://github.com/guardian/service-catalogue/blob/9449204f5c285a7188f7c6e548b4d4b0e1dcfe63/packages/cli/src/index.ts#L105-L109) has a `--confirm` flag, for example.

packages/dev-environment/script/start
akash1810, line 1: Is this file significantly complex enough now to move away from bash, and towards TypeScript? Perhaps merging it with the [CLI package](https://github.com/guardian/service-catalogue/tree/9449204f5c285a7188f7c6e548b4d4b0e1dcfe63/packages/cli)?

packages/dev-environment/script/start
```diff
@@ -68,6 +93,10 @@ else
   echo "AWS credentials are valid"
 fi
 
+setup_libpq
+get_db_password
+source "$MAIN_ENV_FILE"
```
akash1810, line 98: Does this handle comments in the `.env` file as we'd expect?

packages/dev-environment/script/start
NovemberTang: Potentially this overcomplicates things if we've just got one package to install?

packages/dev-environment/script/start
NovemberTang: I'm not sure I understand what you mean? What potential bugs are you anticipating?

packages/dev-environment/script/start
NovemberTang: removed!

packages/dev-environment/script/start
NovemberTang: I've just binned it for now, we can bring it back later if it's something we feel like we need

packages/dev-environment/script/start
NovemberTang: I think for now it's **just about** manageable, especially as this PR has had chunks cut out of it. But if it becomes significantly more complicated than this it'd be worth moving it.

packages/dev-environment/script/start
akash1810: I misunderstood what this is doing - `source` is setting local variables, rather than environment variables.

If we use the `.env` file file to set environment variables, we'd need to strip the comments first:

```console
➜ export $(cat .env | xargs)
export: not valid in this context: ----
````

chrislomaxjones (approved):
Makes sense to me 👍

PR #877, opened 2024-03-27
Stop hand-rolling RepocopVulnerability type

NovemberTang:
## What does this change?

Remove the need to hand-roll a `RepocopVulnerability` type. Derive one from the `repocop_vulnerabilities` interface provided by prisma

## Why?

Maintainability. This is part of a larger piece of work to simplify vulnerability calculations in repocop

## How has it been verified?

All unit tests still pass, local execution of repocop produces the same output as CODE


chrislomaxjones (approved):
Neat 👍

PR #878, opened 2024-03-27
pull all dependabot vulnerabilities up front

NovemberTang:
## What does this change?

Pull dependabot alerts for production repos up front.

## Why?

This brings data loading behaviour for dependabot in line with the rest of repocop. We've generally preferred loading all our data up front, and failing fast if there are any issues. This also pushes `await` statements to the edge of the lambda, making repocop more testable overall.

The function to pull dependabot alerts has been moved to `query.ts`, as to an outside observer, it's behaviour isn't really that different to if we were pulling this information from the CloudQuery database.

## How has it been verified?

Repocop runs as expected. The number of vulnerabilites detected has remained constant.

### DIFF
--- a/packages/repocop/src/query.ts
+++ b/packages/repocop/src/query.ts
@@ -150,6 +150,8 @@ async function getAlertsForRepo(
 			`Dependabot - ${name}: Could not get alerts. Dependabot may not be enabled.`,
 		);
 		console.debug(error);
+		// Return undefined if dependabot is not enabled, to distinguish from
+		// the scenario where it is enabled, but there are no alerts
 		return undefined;
 	}
 }
--- a/packages/repocop/src/query.ts
+++ b/packages/repocop/src/query.ts
@@ -165,11 +165,9 @@ export async function getDependabotVulnerabilities(
 			repos.map(async (repo) => {
 				const alerts = await getAlertsForRepo(octokit, repo.name);
 				if (alerts) {
-					return alerts
-						.filter((a) => a.state === 'open')
-						.map((a) =>
-							dependabotAlertToRepocopVulnerability(repo.full_name, a),
-						);
+					return alerts.map((a) =>
+						dependabotAlertToRepocopVulnerability(repo.full_name, a),
+					);
 				}
 				return [];
 			}),
--- a/packages/repocop/src/query.ts
+++ b/packages/repocop/src/query.ts
@@ -4,9 +4,14 @@ import type {
 	PrismaClient,
 	view_repo_ownership,
 } from '@prisma/client';
+import type { Octokit } from 'octokit';
+import { dependabotAlertToRepocopVulnerability } from './evaluation/repository';
 import type {
+	Alert,
 	AwsCloudFormationStack,
+	DependabotVulnResponse,
 	NonEmptyArray,
+	RepocopVulnerability,
 	Repository,
 	SnykIssue,
 	SnykProject,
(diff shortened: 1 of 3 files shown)


packages/repocop/src/query.ts
```diff
@@ -113,3 +118,65 @@ export async function getRepositoryLanguages(
...
+	repos: Repository[],
+	octokit: Octokit,
+) {
+	const dependabotVulnerabilities: RepocopVulnerability[] = (
+		await Promise.all(
+			repos.map(async (repo) => {
+				const alerts = await getAlertsForRepo(octokit, repo.name);
+				if (alerts) {
+					return alerts
+						.filter((a) => a.state === 'open')
```
akash1810, line 167: The `getAlertsForRepo` is already filtering for `state = 'open'`, so we shouldn't need this additional filter.

packages/repocop/src/query.ts
```diff
@@ -113,3 +118,65 @@ export async function getRepositoryLanguages(
...
+		const openRuntimeDependencies = alert.data.filter(
+			(a) => a.dependency.scope !== 'development',
+		);
+		return openRuntimeDependencies;
+	} catch (error) {
+		console.debug(
+			`Dependabot - ${name}: Could not get alerts. Dependabot may not be enabled.`,
+		);
+		console.debug(error);
+		return undefined;
```
akash1810, line 153: Returning the empty array here, instead of `undefined` would simplify the type signature of this function, and the caller of this function.

packages/repocop/src/query.ts
```diff
@@ -113,3 +118,65 @@ export async function getRepositoryLanguages(
...
+					return alerts
+						.filter((a) => a.state === 'open')
+						.map((a) =>
+							dependabotAlertToRepocopVulnerability(repo.full_name, a),
+						);
+				}
+				return [];
+			}),
+		)
+	).flat();
```
akash1810, line 175: I think this could be a `flatMap` for terseness?

```ts
await Promise.all(repos.flatMap(async repo => ...));
```

packages/repocop/src/query.ts
```diff
@@ -113,3 +118,65 @@ export async function getRepositoryLanguages(
...
+
+export async function getDependabotVulnerabilities(
+	repos: Repository[],
+	octokit: Octokit,
+) {
+	const dependabotVulnerabilities: RepocopVulnerability[] = (
+		await Promise.all(
+			repos.map(async (repo) => {
+				const alerts = await getAlertsForRepo(octokit, repo.name);
+				if (alerts) {
```
akash1810, line 165: If we update the return type of `getAlertsForRepo`, this branch would be redundant.

packages/repocop/src/query.ts
```diff
@@ -113,3 +118,65 @@ export async function getRepositoryLanguages(
...
+
+//Octokit Queries
+
+async function getAlertsForRepo(
+	octokit: Octokit,
+	name: string,
+): Promise<Alert[] | undefined> {
+	if (name.startsWith('guardian/')) {
+		name = name.replace('guardian/', '');
+	}
```
akash1810, line 130: Why is this mutation needed?

akash1810 (commented):
Made a couple of comments on the original code.

packages/repocop/src/query.ts
NovemberTang: I did try that. But no matter where i put the flatmap, it always returns a `[][]` and I need to flatten it again at the end anyway. Bizarre.

packages/repocop/src/query.ts
NovemberTang: This function isn't new, it's just moved, could we fix this in another PR?

packages/repocop/src/query.ts
NovemberTang: Same as above, this function's just been moved over from another file

packages/repocop/src/query.ts
NovemberTang: It was like this before I moved it. I assume it's defensive programming, in case someone passes in a full repo name by accident. Octokit only takes short repo names, and the owner is passed in as a separate parameter.

packages/repocop/src/query.ts
NovemberTang: for some reason, using flatMap doesnt really work. The linter still thinks we are returning a `[][]`, and repocop fails to recognise any vulnerabilities from dependabot. A `.map()` and then a `.flat()` at the end, however, gives us the output we expect.

packages/repocop/src/query.ts
NovemberTang: (the original reasoning for this is that we wanted to maintain a distinction between a repo that had no vulnerabilities (`[]`) differently to a repo that did not have dependency graph enabled (`undefined`) for a future piece of work. I can document this in a comment to make the reasoning clearer.

packages/repocop/src/query.ts
NovemberTang: (resolving as discussed above, using `undefined` to highlight the case where dependabot is not enabled)

tjsilver (approved):
👍

PR #879, opened 2024-03-27
Explicitly version `node` and `prisma` in the migrate image

chrislomaxjones:
## What does this change?

- Use the `node:20.11.1-alpine` image, which matches the `.nvmrc`.
- Use `prisma@5.11.0`, which matches the rest of the service-catalogue (specifically `packages/common/package.json`).

Thanks @akash1810 for the heads up.

## Why?

The versions of software used in the image should be the same as those used in the rest of the repo.

## How has it been verified?

...


chrislomaxjones:
This will require a subsequent PR bumping the commit hash, such as https://github.com/guardian/service-catalogue/pull/883.

containers/prisma-migrate/Dockerfile
```diff
@@ -9,7 +9,8 @@ RUN apk --no-cache add aws-cli
 # Install the Prisma NPM package
 # This will allow us to use the prisma migrate deploy command
 RUN npm init -y
-RUN npm install prisma
+# TODO: Keep the version automatically in sync with common/package.json
+RUN npm install prisma@5.11.0
```
akash1810, line 13: Suspect this'll get annoying very quickly as [Dependabot](https://github.com/guardian/service-catalogue/blob/main/.github/dependabot.yml) updates Prisma every week...

akash1810 (approved):
We should address the TODO ASAP.

containers/prisma-migrate/Dockerfile
chrislomaxjones: Good point 👍 

Simplest route might be to just `COPY ./packages/common/package.json ...` and `RUN npm install`. A multi-stage build could be used to drop all of the dependencies we don't care about.

PR #881, opened 2024-03-27
allow snyk integrator to support kotlin and ruby

NovemberTang:
## What does this change?

Allows PRs to be raised for repos that contain Kotlin or Ruby

## Why?

Our action supports these languages


chrislomaxjones (approved):
Great!

PR #884, opened 2024-03-27
Move to Arm64 lambdas 

akash1810:
## What does this change?
Change the architecture of each lambda to Arm64.

## Why?
Arm based lambdas are cheaper and more performant. See https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/.

## How has it been verified?
- [Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/8df7ce4d-c5bf-42af-941e-e86d88a1bfd2)
- Successfully ran each lambda

## Statistics
Here are some statistics after invoking the x86_64 and arm64 versions of each lambda once with the same input.

Lambda | Architecture | Duration (ms) | Init Duration (ms) | Max Memory Used (MB)
-- | -- | -- | -- | --
data-audit | arm64 | 43,625.24 | 544.92 | 226
data-audit | x86_64 | 40,958.23 | 471.65 | 217
dependency-graph-integrator | arm64 | 11.61 | 404.20 | 91
dependency-graph-integrator | x86_64 | 15.16 | 386.19 | 87
github-actions-usage | arm64 | 15,342.05 | 294.96 | 188
github-actions-usage | x86_64 | 12,884.00 | 293.36 | 190
interactive-monitor | arm64 | 2,022.39 | 56.95 | 201
interactive-monitor | x86_64 | 1,970.55 | 833.22 | 191
repocop | arm64 | 27,607.03 | 909.56 | 683
repocop | x86_64 | 24,273.92 | 950.93 | 640
snyk-integrator | arm64 | 16.76 | 397.81 | 91
snyk-integrator | x86_64 | 14.90 | 359.59 | 87

As mentioned in the linked post:

> For functions using the Arm/Graviton2 architecture, duration charges are 20 percent lower than the current pricing for x86.

In [this estimate](https://calculator.aws/#/estimate?id=d904ec352da22ae5f7e0d70a665589b3367b3b5a)[^1] of 100,000 requests per month, of 50,000ms duration, and 1,024MB memory, Arm64 costs $61.33 and x86_64 $76.67.

That is, whilst the statistics do not suggest a dramatic change in performance, we will see a reduction in the bill.

[^1]: URL will expire in 1 year.

### DIFF
--- a/packages/cdk/lib/service-catalogue.test.ts
+++ b/packages/cdk/lib/service-catalogue.test.ts
@@ -1,5 +1,6 @@
 import { App } from 'aws-cdk-lib';
 import { Template } from 'aws-cdk-lib/assertions';
+import { CfnFunction } from 'aws-cdk-lib/aws-lambda';
 import { ServiceCatalogue } from './service-catalogue';
 
 describe('The ServiceCatalogue stack', () => {
@@ -11,5 +12,27 @@ describe('The ServiceCatalogue stack', () => {
 		});
 		const template = Template.fromStack(stack);
 		expect(template.toJSON()).toMatchSnapshot();
+	});
+
+	it('only uses arm64 lambdas', () => {
+		const app = new App();
+		const stack = new ServiceCatalogue(app, 'ServiceCatalogue', {
+			stack: 'deploy',
+			stage: 'TEST',
+		});
+
+		const lambdas = stack.node
+			.findAll()
+			.filter((child): child is CfnFunction => child instanceof CfnFunction);
+
+		const architectures = new Set(
+			lambdas.flatMap((lambda) => lambda.architectures),
+		);
+
+		// Only 1 architecture is used...
+		expect(architectures.size).toEqual(1);
+
+		// ...and it's arm64
+		expect(architectures.has('arm64')).toEqual(true);
 	});
 });
--- a/packages/common/prisma/schema.prisma
+++ b/packages/common/prisma/schema.prisma
@@ -1,7 +1,7 @@
 generator client {
   provider        = "prisma-client-js"
   previewFeatures = ["views"]
-  binaryTargets   = ["native", "rhel-openssl-3.0.x"]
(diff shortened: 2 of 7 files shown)


packages/common/prisma/schema.prisma
```diff
@@ -1,7 +1,7 @@
 generator client {
   provider        = "prisma-client-js"
   previewFeatures = ["views"]
-  binaryTargets   = ["native", "rhel-openssl-3.0.x"]
+  binaryTargets   = ["native", "linux-arm64-openssl-3.0.x"]
```
akash1810, line 4: See https://www.prisma.io/docs/orm/reference/prisma-schema-reference#binarytargets-options.

packages/cdk/lib/service-catalogue.test.ts
```diff
@@ -12,4 +13,26 @@ describe('The ServiceCatalogue stack', () => {
 		const template = Template.fromStack(stack);
 		expect(template.toJSON()).toMatchSnapshot();
 	});
+
+	it('only uses arm64 lambdas', () => {
```
akash1810, line 17: This test should help enforce usage of arm64 based lambdas.

NovemberTang (commented):
Having reviewed the table, it looks like our arm-based lambdas generally take longer to run, and have a higher memory usage. How does this balance out with a cheaper unit price?

akash1810:
> Having reviewed the table, it looks like our arm-based lambdas generally take longer to run, and have a higher memory usage. How does this balance out with a cheaper unit price?

With the same duration, and memory usage Arm64 is 20% cheaper than x86_64.

The "worst" change is to the github-actions-usage lambda, which sees a 19.08% duration increase. Given these lambdas don't have need to complete in a given timeframe, I think this change is acceptable. Whilst it'll take longer, the Arm64 version will cost ~16% less[^1].

[^1]: ~80% of 20%.

NovemberTang (approved):

PR #891, opened 2024-03-28
fix(cloudquery): Remove erroneous configuration

akash1810:
## What does this change?
Since the version update in https://github.com/guardian/service-catalogue/pull/809, the `snyk_reporting_issues` property is not valid, as observed in the logs:

> jsonschema: '/table_options' does not validate with https://github.com/cloudquery/cloudquery/plugins/source/snyk/client/spec/spec#/$ref/properties/table_options/$ref/additionalProperties: additionalProperties 'snyk_reporting_issues' not allowed

It has been replaced with [`created_after`](https://hub.cloudquery.io/plugins/source/cloudquery/snyk/latest/docs?search=sny#overview-snyk-table-options-spec) which takes an ISO 8601 date string.

Achieving the same "only collect issues from 30 days ago" with the new configuration options is tricky, as the ISO 8601 string will be set at build time, not runtime...

Additionally, to avoid alarms, and querying of stale data, remove the tables that are no longer [collected by CloudQuery](https://hub.cloudquery.io/plugins/source/cloudquery/snyk/latest/tables).

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/6a0a87c8-6b1e-4982-a92d-850757f85520), ran the `SnykAll` task, and observed the absence of the above log.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -13636,9 +13636,6 @@ spec:
     - postgresql
   spec:
     api_key: \${SNYK_API_KEY}
-    table_options:
-      snyk_reporting_issues:
-        period: 30d
 ' > /source.yaml;printf 'kind: destination
 spec:
   name: postgresql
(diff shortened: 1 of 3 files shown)


NovemberTang (approved):
🙌

PR #896, opened 2024-04-03
ci: Add test reporter

akash1810:
## What does this change?
Improve the DX of a failing test using https://jestjs.io/docs/configuration#github-actions-reporter to annotate a pull request with details of a failing test.

#### When a unit test fails
image

#### When the CDK snapshots have not been updated
image

## Why?
The build log is fairly busy, and can be difficult to isolate which test is failing. Annotating the failing test(s) improves the DX.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -11,6 +11,7 @@ const transform = {
 module.exports = {
 	verbose: true,
 	testEnvironment: 'node',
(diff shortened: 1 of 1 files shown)


AshCorr (approved):
Ooh, satisfyingly easy to integrate with GHA!

PR #906, opened 2024-04-05
Add `aws-otel-collector` cloudquery sidecar to collect OTLP traces and forward them to XRay

AshCorr:
## What does this change?
Adds a new sidecar container to all our tasks. This sidecar runs [`aws-otel-collector`](https://github.com/aws-observability/aws-otel-collector) which collects Telemetry traces from CloudQuery and forwards them to AWS XRay[^1].

The `aws-otel-collector` sidecar also supports collecting container metrics such as CPU usage and Memory usage, but @NovemberTang found that its possible to enable `Container Insights` at a cluster level and AWS will collect roughly this information for us. So for now I've disabled the `aws-otel-collectors` ability to collect metrics.

## Why?
Cloudquery is capable of [generating traces](https://docs.cloudquery.io/docs/advanced-topics/monitoring) to let developers understand whats happening under the hood. This is good for debugging and fine tuning job resources.

This also lets us explore XRay a bit and understand how much it might cost for us to roll out tracing across the department and whether we need to explore other alternative tracing providers.

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/81a93c5e-93a1-4cf3-a105-d8d92145766b), and after the `DeployToolsListOrgs` is been manually run, we can see traces in the AWS console:

image

And logs in Central ELK:

image

[^1]: See https://github.com/aws-observability/aws-otel-collector/blob/main/deployment-template/ecs/aws-otel-ec2-sidecar-deployment-cfn.yaml.

### DIFF
--- a/packages/cdk/lib/cloudquery/task.ts
+++ b/packages/cdk/lib/cloudquery/task.ts
@@ -1,6 +1,6 @@
 import type { AppIdentity, GuStack } from '@guardian/cdk/lib/constructs/core';
 import type { GuSecurityGroup } from '@guardian/cdk/lib/constructs/ec2';
-import { Tags } from 'aws-cdk-lib';
+import { Duration, Tags } from 'aws-cdk-lib';
 import type { ISecurityGroup } from 'aws-cdk-lib/aws-ec2';
 import {
 	ContainerDependencyCondition,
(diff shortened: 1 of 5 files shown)


packages/cdk/lib/cloudquery/task.ts
```diff
@@ -224,6 +228,26 @@ export class ScheduledCloudqueryTask extends ScheduledFargateTask {
 			logging: fireLensLogDriver,
 		});
 
+		const otel = task.addContainer(`${id}AWSOTELCollector`, {
+			image: Images.otelCollector,
+			command: ['--config=/etc/ecs/ecs-xray.yaml'],
```
akash1810, line 233: See https://github.com/aws-observability/aws-otel-collector/blob/0d65499593b4a0a794d41cd6c08baadd4004ec1a/config/ecs/ecs-xray.yaml for the content of this file.

akash1810 (approved):

PR #907, opened 2024-04-05
Enable Container Insights

AshCorr:
## What does this change?

Enables Container Insights for ECS.

## Why?

Container Insights tracks container metrics such as CPU Utilization & Memory Utilization in Cloudwatch allowing us to more easily identify jobs that have too many or too few resources assigned to them.

A slight concern with this PR is how expensive custom metrics are in Cloudwatch, with each custom metric costing 0.30$ per month it can be quite scary to have ~10 custom metrics per task. However, as I understand metrics are charged by the hour and only if they're actively being pushed to, this means for most tasks that are run daily the metric will only be "active" 1 hour every day, therefore costing 0.30/24 = 0.0125$ per month per metric.

We have 26 tasks, so it roughly works out at 3.25$ per month for these metrics assuming my above understanding is correct.

## How has it been verified?

Tested in CODE.

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -18132,6 +18132,12 @@ spec:
     },
     "servicecatalogueCluster5FC34DC5": {
       "Properties": {
+        "ClusterSettings": [
+          {
+            "Name": "containerInsights",
+            "Value": "disabled",
+          },
+        ],
         "Tags": [
           {
             "Key": "gu:cdk:version",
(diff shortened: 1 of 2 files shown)


akash1810 (approved):
minor: The implementation of this stage-aware property differs from the others, which are set at the [top level](https://github.com/guardian/service-catalogue/blob/main/packages/cdk/bin/cdk.ts). Worth aligning approaches?

AshCorr:
> minor: The implementation of this stage-aware property differs from the others, which are set at the [top level](https://github.com/guardian/service-catalogue/blob/main/packages/cdk/bin/cdk.ts). Worth aligning approaches?

Switched it to always be on. Had it set to PROD only before I realized that Cloudquery Metrics are only billed per hour of being pushed to.

PR #908, opened 2024-04-08
Link to dependency vulnerability dashboard

NovemberTang:
## What does this change?

Provide a link for teams to see more information about their vulnerabilities

## Why?

So teams can see all their vulnerabilities at a glance, not just the top 10.

## How has it been verified?

Clicking on the link works

### DIFF
--- a/packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
+++ b/packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
@@ -85,6 +85,17 @@ export function isFirstOrThirdTuesdayOfMonth(date: Date) {
 	return isTuesday && (inFirstWeek || inThirdWeek);
 }
 
+function createGrafanaAction(teamSlug: string): Action {
+	const url =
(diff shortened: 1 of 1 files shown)


packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
```diff
@@ -85,6 +85,17 @@ export function isFirstOrThirdTuesdayOfMonth(date: Date) {
 	return isTuesday && (inFirstWeek || inThirdWeek);
 }
 
+function createGrafanaAction(teamSlug: string): Action {
+	const url =
```
akash1810, line 89: Worth using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) here to make the URL easier to modify and reason about?

```ts
const params = new URLSearchParams({
  name: "Bruce Wayne",
  alias: "Batman"
});

const queryString = params.toString(); // name=Bruce%20Wayne&alias=Batman
```

PR #909, opened 2024-04-08
New view view_aws_resources for tagging OKR baseline

JuliaBrigitte:
## What does this change?
Add a view to service-catalogue for easier querying of resources and their tags

## Why?
We want more visibility on tagged resources

## How has it been verified?
Tried locally


akash1810 (approved):
💎

akash1810:
Worth linking to the CloudQuery provided query that inspired this?

PR #917, opened 2024-04-09
Add new `aws_resources()` function which returns a dedupped list of AWS Resources tracked by cloudquery

AshCorr:
## What does this change?

 - **Drops the `view_aws_resources` view in favour of a function.**

Views can be a bit problematic for CloudQuery and Postgres. The structure of any tables a view depends on cannot be changed, this means that you have to delete the view before you can change the underlying tables.

Since our view references most AWS Cloudquery tables this means that we won't be able to update any of our tables without deleting our view. This can also cause CloudQuery to behave unpredictably as it will try to update table schemas whenever a plugin is updated.

Functions on the other hand don't have this limitation. Due to how much custom code a function can contain its impossible for Postgres to do integrity checks on tables used by a function.

Using a function in a query is pretty straight forward. The only change users need to do is append `()` to the function name in a query, for example

```
SELECT * FROM aws_resources()
```
instead of
```
SELECT * FROM aws_resources
```

- **Track all resources, not just taggable ones**

The previous view ignored tables that didn't have a `tags` column, this meant that untaggable resources couldn't be found in the view. This was a quirk of how we tried to dedupe rows.

This function improves on that by introducing a `taggable` column. If a given resource has no tables with a `tags` column, this column is set to false, otherwise its set to true.

 - **This function looks for any table that has an Account ID and ARN column and aggregates all its rows with similar tables.**

ARNs are deduplicated so that each resource only has 1 row in the output. This is important as an AWS resource can be described over multiple different tables, for example S3 buckets have `aws_s3_buckets`, `aws_s3_bucket_policies`, and `aws_s3_bucket_lifecycles` all of which can contain a record for the same ARN.

## How has it been verified?

Tested in DEV and CODE

### DIFF
--- a/packages/common/prisma/migrations/20240409101800_function_aws_resource/migration.sql
+++ b/packages/common/prisma/migrations/20240409101800_function_aws_resource/migration.sql
@@ -1,3 +1,6 @@
+-- Cloudquery cant update the schema of tables referenced by views. 
+-- As this view references any Cloudquery table containing an ARN column It can make Cloudquery behave strangely.
+-- This migration replaces this view with a function which doesn't have the same restriction.
 DROP VIEW view_aws_resources;
 
 -- Custom aggregate function that combines two JSONB objects together. Uses the inbuilt `jsonb_concat` function.
@@ -66,7 +69,7 @@ BEGIN
                                                      AND table_name = cloudquery_table.table_name)
                                           THEN 'request_account_id'
                                       ELSE 'NULL' END;
-            -- Check if table has an region column, or default to NULL
+            -- Check if table has an region column, or default to unavailable
             region := CASE
                           WHEN EXISTS (SELECT 1
                                        FROM information_schema.columns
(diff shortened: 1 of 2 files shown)


packages/common/prisma/migrations/20240409101800_function_aws_resource/migration.sql
```diff
@@ -0,0 +1,138 @@
...
+
+            IF NOT (str_sql = ''::TEXT) THEN
+                str_sql = str_sql || ' UNION ALL ';
+            END IF;
+
+            -- Fetch rows of table and append to function output
+            str_sql = str_sql || FORMAT(E'
+                    SELECT
+                        %L as cq_table,
+                        COALESCE(%s, SPLIT_PART(arn, \':\', 5)) AS account_id,
```
akash1810, line 102: Not sure how exactly, but having unit tests for the manipulation of the `arn` field would be useful.

akash1810 (approved):
Couple of non-blocking comments, but LGTM.

PR #922, opened 2024-04-10
Add aws_securityhub_findings table to prisma schema, and make more performant

NovemberTang:
## What does this change?

Adds aws_securityhub_findings to the prisma schema
Creates indexes for the table, for faster searching

## Why?

Running this query

```sql
SELECT title
     , resources
     , aws_account_name
     , severity
     , workflow
FROM aws_securityhub_findings
WHERE aws_account_name = '<MY_ACCOUNT>'
  AND workflow ->> 'Status' != 'SUPPRESSED'
ORDER BY severity ->> 'Normalized' DESC
                LIMIT 50
```

previously took 22 seconds. When using this table as a source for grafana, this results in intolerable waiting times when graphs are loading. By creating a migration that includes adding indexes, we can see the query time reduced to 220 ms, a 100x improvement

## How has it been verified?

The migration has already been applied to CODE, and the query benchmarked

### DIFF
--- a/packages/common/prisma/migrations/20240410101502_aws_securityhub_findings/migration.sql
+++ b/packages/common/prisma/migrations/20240410101502_aws_securityhub_findings/migration.sql
@@ -54,8 +54,8 @@ CREATE TABLE IF NOT EXISTS "aws_securityhub_findings" (
     CONSTRAINT "aws_securityhub_findings_cqpk" PRIMARY KEY ("request_account_id","request_region","aws_account_id","created_at","description","generator_id","id","product_arn","schema_version","title","updated_at","region")
 );
 -- CreateIndex
-CREATE UNIQUE INDEX "aws_securityhub_findings__cq_id_key" ON "aws_securityhub_findings"("_cq_id");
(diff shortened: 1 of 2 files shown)


packages/common/prisma/migrations/20240410101502_aws_securityhub_findings/migration.sql
```diff
@@ -0,0 +1,61 @@
...
+    "vulnerabilities" JSONB,
+    "workflow" JSONB,
+    "workflow_state" TEXT,
+
+    CONSTRAINT "aws_securityhub_findings_cqpk" PRIMARY KEY ("request_account_id","request_region","aws_account_id","created_at","description","generator_id","id","product_arn","schema_version","title","updated_at","region")
+);
+-- CreateIndex
+CREATE UNIQUE INDEX IF NOT EXISTS "aws_securityhub_findings__cq_id_key" ON "aws_securityhub_findings"("_cq_id");
+
+CREATE INDEX IF NOT EXISTS idx_aws_account_name ON aws_securityhub_findings ((aws_account_name));
```
NovemberTang, line 59: These three lines are the important bit, everything else is pretty much a no-op

adamnfish (approved):
Nice one. We'll take all the other parts for granted, but the indexes LGTM.

PR #923, opened 2024-04-10
Fix bugs in aws_resource function

AshCorr:
## What does this change?

Missed some things in https://github.com/guardian/service-catalogue/pull/917


akash1810 (approved):
Could we add more detail to the PR description? Flying a little blind in the review as unsure what's being fixed and why...

AshCorr:
> Could we add more detail to the PR description? Flying a little blind in the review as unsure what's being fixed and why...

Updated!

PR #924, opened 2024-04-11
Refactor and cleanup aws_resources() and create a materialized view

AshCorr:
### Refactors to `aws_resources()`

I've gutted a lot of the behaviour from aws_resources() in an effort to make it more readable and less complex.

- `aws_resources()` has been renamed to `aws_resources_raw()` and now returns the raw rows from all tables containing an ARN column. Deduping and grouping is now handled by consumers of `aws_resources_raw()` instead of the function itself. This vastly improves readability as there is no longer a need to build a giant query string.

- A lot of the ARN manipulation code to extract properties of the ARN have been moved to their own functions, again reducing the complexity of the `aws_resources()` function. Adds 5 new functions: `arn_to_partition(arn)`, `arn_to_service(arn)`, `arn_to_region(arn)`, `arn_to_account_id(arn)`, `arn_to_resource_type(arn)`
- Code to check if a table has a column has also been extracted to `table_has_column`, reducing duplication of code.

### Create a new materialized view `aws_resources`

This new view groups the results of `aws_resources_raw()` so that ARNs are __mostly__ deduplicated. Similar to the old behaviour certain ARNS that span multiple accounts/regions are still duplicated.

This new view is also a materialized view, materialized views differ from regular views in that they cache/store the results of their underlying query. If 100 people query `aws_resources` it will only make one call to `aws_resources_raw`, hopefully massively reducing the time for dashboards to load when relying on this data.

This has the downside that the view has to be refreshed in order to get up to date data. This is done with the query `REFRESH MATERIALIZED VIEW aws_resources`. This PR doesn't handle refreshing the view regularly yet. Theres a couple of ways I can think of to do this:

  - Using the [pg_cron](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL_pg_cron.html) extension to run `REFRESH MATERIALIZED VIEW aws_resources` every few minutes. I don't like the idea of our database being a process manager though.
  - Add another sidecar container to AWS Cloudquery jobs that run `REFRESH MATERIALIZED VIEW`. This has the benefit of only refreshing the view when the underlying data changes, but it does make our Cloudquery tasks yet more complicated.

### Adds testing of custom functions

Slightly hacky, but I figured since we apply all our migrations in CI we could also run some code in our migrations that test the behaviour of the custom functions we create. This is done by creating a temporary `__TEST()` function that runs all our tests, executing it, and then dropping it at the end of the migration.

## How has it been verified?

Tested Locally

### DIFF
--- a/packages/common/prisma/migrations/20240409101800_function_aws_resource/migration.sql
+++ b/packages/common/prisma/migrations/20240409101800_function_aws_resource/migration.sql
@@ -1,16 +1,17 @@
 -- Cloudquery cant update the schema of tables referenced by views. 
 -- As this view references any Cloudquery table containing an ARN column It can make Cloudquery behave strangely.
 -- This migration replaces this view with a function which doesn't have the same restriction.
-DROP VIEW view_aws_resources;
+DROP VIEW IF EXISTS view_aws_resources;
 
 -- Custom aggregate function that combines two JSONB objects together. Uses the inbuilt `jsonb_concat` function.
 -- Useful for when grouping many rows together that share a JSONB column
-CREATE AGGREGATE jsonb_aggregate (jsonb)(
+CREATE OR REPLACE AGGREGATE jsonb_aggregate (jsonb)(
     SFUNC = jsonb_concat,
     STYPE = jsonb
 );
 
 -- Aggregate rows from all AWS Cloudquery tables
+DROP FUNCTION IF EXISTS aws_resources();
 CREATE FUNCTION aws_resources()
     RETURNS TABLE
             (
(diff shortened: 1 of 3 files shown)


JuliaBrigitte:
How do we sanity check the data provided by this view?

AshCorr:
> How do we sanity check the data provided by this view?

I can't really think of a great way of doing this. My hunch is that this data doesn't need to be 100% accurate.

I think its important that we look at our main services like EC2, ECS, VPC and make sure that those are accurate. Maybe we can take some examples from each service and make sure that they show up and have the correct tags.

Theres just so many Cloudquery tables that we would need to check to make sure they match our expected schema, at very least they need to have a text column called ARN, and a JSON column called tags, and we've already found some that don't match

JuliaBrigitte (approved):
Reviewed in code

PR #929, opened 2024-04-15
Skip `aws_wellarchitected` tables

AshCorr:
## What does this change?

Skips the `aws_wellarchitected` tables.

## Why?

We don't use wellarchitected so the data from these tables is not very useful. They also contain a lot of AWS managed resources that are untagged which throws off our tagging baseline.


JuliaBrigitte (approved):
Are there other table we can drop?

AshCorr:
> Are there other table we can drop?

Definitely. I think its outside the scope of this PR though, we're dropping this table in particular specifically because its impacting us.

PR #930, opened 2024-04-15
feat: Remove Snyk vulnerability IDs where possible in favour of CVEs.

NovemberTang:
Co-authored-by: @tjsilver 

## What does this change?

If a CVE-based vulnerability ID is available, pass only that to the `cves` field of the `RepocopVulnerability` interface. Otherwise, fall back to the original list of IDs.

## Why?

Snyk usually gives a vulnerability an internal Snyk ID, regardless of whether or not there is already a CVE.
This means there are lots of scenarios where a dependabot issue looks like this `[CVE-1234]`, and its corresponding Snyk issue could look like this `[CVE-1234, SNYK-5678]`, meaning that it does not get deduplicated.
Now that we are filtering out the Snyk ID wherever possible, we are able to deduplicate CVEs more accurately

## How has it been verified?

- Unit tests have been added to verify the correct behaviour.
- Repocop has been run on CODE, where we saw the number of vulnerabilities drop by 11%

### DIFF
--- a/packages/repocop/src/evaluation/repository.test.ts
+++ b/packages/repocop/src/evaluation/repository.test.ts
@@ -21,6 +21,7 @@ import {
 	hasDependencyTracking,
 	hasOldAlerts,
 	snykAlertToRepocopVulnerability,
+	snykVulnIdFilter,
 } from './repository';
 
 function evaluateRepoTestHelper(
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
Paired on the writing and testing of this and all working as expected.

PR #932, opened 2024-04-16
feat: add action to link to team vuln dashboard

tjsilver:
## What does this change?

- Adds a new action to the vulnerability digest Anghammarad notification which links to the Grafana dashboard filtered to the team slug.
- Moves the action creation into the `createDigest`  function
- Adds an `actions` field to the VulnerabilityDigest type.

## Why?

Teams will be able to view all their Critical and High vulnerabilities in one place.

## How has it been verified?

All unit tests pass and the shape of the digest produced is as expected.

### DIFF
--- a/packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
+++ b/packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
@@ -100,7 +100,6 @@ describe('createDigest', () => {
 		expect(
 			createDigest(team, [ownershipRecord], [resultWithVuln]),
 		).toStrictEqual({
-			teamName,
 			teamSlug,
 			subject: `Vulnerability Digest for ${teamName}`,
 			message: String.raw`Found 1 vulnerabilities across 1 repositories.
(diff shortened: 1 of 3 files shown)


packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
```diff
@@ -109,6 +109,16 @@ Note: DevX only aggregates vulnerability information for repositories with a pro
 [guardian/repo](https://github.com/guardian/repo) contains a [HIGH vulnerability](example.com).
 Introduced via **leftpad** on Fri Jan 01 2021, from pip.
 This vulnerability is patchable.`,
+			actions: [
+				{
+					cta: `View vulnerability dashboard for ${teamName} on Grafana`,
+					url: `https://metrics.gutools.co.uk/d/fdib3p8l85jwgd?var-repo_owner=${teamSlug}`,
+				},
+				{
+					cta: "See 'Prioritise the vulnerabilities' in these docs for obligations",
```
NovemberTang, line 118: Side note, no change required for this this PR: I wonder if we could just state the obligations in the digest, rather than make people click around to find them?

NovemberTang (approved):
Looks great!

packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
tjsilver: Yeah I was wondering that too - wasn't sure what the original decision was but it makes more sense as we could state it concisely I think!

PR #934, opened 2024-04-17
Enable `readonlyRootFilesystem` in our ECS Task config

AshCorr:
## What does this change?

 - Enable readonlyRootFilesystem on our ECS Tasks. This resolves 52 FSBP ECS.5 violations by locking down the container filesystem to prevent unexpected writes.
 - Adds ephemeral mounts for Internal Cloudquery data and User config to be written to. Functionally, each of these mounts replaces a directory in the file system that would otherwise be read-only.
 - Creates a custom Dockerfile that already contains the required AWS certificates, and a workflow to release the image, similar to `prisma-migrate-image.yml`

## How has it been verified?

Tested in CODE against AWS and GitHub jobs. Both run without issue

### DIFF
--- a/packages/cdk/lib/cloudquery/config.ts
+++ b/packages/cdk/lib/cloudquery/config.ts
@@ -142,7 +142,6 @@ export function githubSourceConfig(
 						org: 'guardian',
 
 						// For simplicity, read all configuration from disk.
-						//These values are duplicated. Can we use a variable for them?
 						private_key_path: `${serviceCatalogueConfigDirectory}/github-private-key`,
 						app_id:
 							'${' +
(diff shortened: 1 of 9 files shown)


packages/cdk/lib/cloudquery/config.ts
```diff
@@ -142,9 +142,10 @@ export function githubSourceConfig(
 						org: 'guardian',
 
 						// For simplicity, read all configuration from disk.
-						private_key_path: '/github-private-key',
-						app_id: '${file:/github-app-id}',
-						installation_id: '${file:/github-installation-id}',
+						//These values are duplicated. Can we use a variable for them?
+						private_key_path: '/data/github-private-key',
```
tjsilver, line 146: Does this need to be changed to `usr/share/cloudquery`?

tjsilver (approved):
Great work!

PR #935, opened 2024-04-17
Bump AWS Source plugin to v26.0.0

AshCorr:
## What does this change?

Bumps AWS Source plugin to v26.0.0

## Why?

This version includes a bug fix to various tables where the `tags` column was not in the correct format.

- Change tag column to a map {} rather than an array of objects [{}] for tables aws_directoryservice_directories, aws_ec2_traffic_mirror_filters, aws_ec2_traffic_mirror_sessions, aws_ec2_traffic_mirror_targets, aws_emr_notebook_executions, aws_emr_studios, aws_transfer_agreements, aws_transfer_users, aws_transfer_certificates, aws_transfer_connectors, aws_transfer_profiles, aws_transfer_workflows

In v25.0.0 a number of AWS tables were moved to premium. We should keep an eye on our row usage to make sure it doesn't increase drastically.

### DIFF
--- a/.env
+++ b/.env
@@ -12,7 +12,7 @@ CQ_POSTGRES_DESTINATION=7.2.0
 CQ_POSTGRES_SOURCE=3.0.7
 
 # See https://hub.cloudquery.io/plugins/source/cloudquery/aws/versions
-CQ_AWS=23.6.1
+CQ_AWS=26.0.0
 
 # See https://hub.cloudquery.io/plugins/source/cloudquery/github/versions
 CQ_GITHUB=8.1.3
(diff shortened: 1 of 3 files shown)


NovemberTang (commented):
We're not using any of these tables for anything right? So no migration required?

akash1810 (approved):
We're updating across 3 major versions. I think only [v24.0.0](https://hub.cloudquery.io/plugins/source/cloudquery/aws/v24.0.0/versions) includes significant changes for us[^1].

[^1]: [v25.0.0](https://hub.cloudquery.io/plugins/source/cloudquery/aws/v25.0.0/versions) moved some tables to the paid tier.

AshCorr:
> We're not using any of these tables for anything right? So no migration required?

Nope, should be fine.

PR #937, opened 2024-04-22
Make singleton container readonly

NovemberTang:
## What does this change?

Create a custom docker image for the singleton container, so we don't need to install awscli and jq.

## Why?

This is part of the work to resolve AWS FSBP ECS.5 (No write access to root file system by default

## How has it been verified?

Tested on CODE to verify expected behaviour for one job, and multiple concurrent ones.

## What next

Create a reusable workflow for creating container images, as we now have 3 of these. This will be done in a follow-up PR

### DIFF
--- a/containers/singleton/Dockerfile
+++ b/containers/singleton/Dockerfile
@@ -1,3 +1,3 @@
-FROM amazonlinux:latest
+FROM amazonlinux:2.0.20240412.0
 
 RUN yum install -y -q aws-cli jq
--- a/containers/singleton/Dockerfile
+++ b/containers/singleton/Dockerfile
@@ -1,0 +1,3 @@
+FROM amazonlinux:latest
+
+RUN yum install -y -q aws-cli jq
--- a/.github/workflows/singleton.yml
+++ b/.github/workflows/singleton.yml
@@ -8,7 +8,6 @@ on:
   push:
     branches:
       - main
-      - nt/more-readonly
     paths:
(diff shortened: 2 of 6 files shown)


containers/singleton/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM amazonlinux:latest
```
akash1810, line 1: Should we pin to a version here? `latest` in the previous version was always up-to-date as the image is pulled on each task invocation. With this pre-baked image, `latest` will be the version that exists at the time the image is created. By pinning, we become more explicit.

containers/singleton/Dockerfile
akash1810, line 1: Did we consider using the [AWS CLI image](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-docker.html) plus some `awk` as a replacement for `jq`?

[I was meant to experiment with this previously...](https://github.com/guardian/service-catalogue/pull/484/files#r1397153766)

NovemberTang: We talked about this, but I think it represents an extra level of scope creep on a ticket that's already crept up massively, and could probably be a separate investigation.

containers/singleton/Dockerfile
NovemberTang: hmmm good shout, I hadn't considered that

tjsilver (approved):
Looks good!

PR #944, opened 2024-04-23
Create reusable workflow for pushing docker images

NovemberTang:
## What does this change?

Creates a shared workflow for pushing container images.
Creates a `stable` tag that gets added if an image was pushed from the main branch. This does not follow our best practice of using the long SHA tag, but I am choosing to not follow this practice for two reasons:

1. We do not have a convenient way to automatically update the SHA when a new one is released, similar to dependabot. Hopefully this is a feature we can introduce in the future, but for now, this avoids the scenario of a developer who doesn't understand the process well releasing a new image version, and forgetting to update the version of the image in CDK.
2. A stable tag can only be released on merge to main, after all CI checks have passed, **and** a reviewer has approved the changes. So it's quite unlikely that the underlying version will change under our feet.

## Why?

There's a lot of boilerplate involved in pushing a docker image. Creating a shared workflow with a couple of parameters makes this more DRY

## How has it been verified?

- CI checks passed
- `stable` docker images exist.
- CQ containers still run successfully

## What's next

A follow up PR will be required to make sure that the workflow is pointing to the correct version. I think in this case we want it pointing to `main`, rather than a hash so that every PR doesn't require a follow-up PR to change the version of `release-image` being used, ad infinitum. For now, I'm setting it to a fixed version so that the actions don't break when the underlying branch is deleted.

Once we've finished resolving the securityhub alerts, we should come back to this and think about how to improve the versioning story. We also may be able to move off of amazonlinux and onto alpine, reducing the singleton image size significantly.


tjsilver (approved):
I paired with @NovemberTang to review, all makes sense!

PR #945, opened 2024-04-23
use the main branch of release-image instead of a tagged version

NovemberTang:
## What does this change?

Switches from a named release to a branch

## Why?

I think versioned releases make sense if we are using a workflow from another repo. However, as this action is only used by the service catalogue, I think that doing this would create unnecessary complication, in exchange for very little security benefit. In order to update a versioned workflow, the steps would be

1. Create a branch with the desired change,
2. Test the change referencing `branch-name` instead of `vX.Y.Z`
3. Remember to put it back at the end
4. Create a PR, and merge it to main
5. Create a new version via the CLI or GitHub UI
6. Create a second PR, advancing the version referenced in the action to the new release.

Using the relative workflow path, the steps are
1. Create a branch with the desired change
2. Test it
3. Create a PR, and merge it to main.

### DIFF
--- a/.github/workflows/cq-image.yml
+++ b/.github/workflows/cq-image.yml
@@ -44,7 +44,7 @@ jobs:
 
   build-and-push:
     needs: CQ_CLI
-    uses: guardian/service-catalogue/.github/workflows/release-image.yml@v0.0.1
+    uses: guardian/service-catalogue/.github/workflows/release-image.yml@main
     with:
       IMAGE_NAME: cloudquery
       BUILD_ARGS: CQ_CLI=${{ needs.CQ_CLI.outputs.CQ_CLI }}
--- a/.github/workflows/cq-image.yml
+++ b/.github/workflows/cq-image.yml
@@ -44,7 +44,7 @@ jobs:
 
   build-and-push:
     needs: CQ_CLI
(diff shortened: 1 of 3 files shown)


.github/workflows/cq-image.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
   build-and-push:
     needs: CQ_CLI
-    uses: guardian/service-catalogue/.github/workflows/release-image.yml@v0.0.1
+    uses: guardian/service-catalogue/.github/workflows/release-image.yml@main
```
akash1810, line 47: Could we use a relative path instead, so that it tracks against the current branch? This'll have the added benefit of being able to test changes to the `release-image.yml` workflow before they reach `main`.

An example of this pattern can be seen in [guardian/dotcom-rendering](https://github.com/guardian/dotcom-rendering/blob/6aca8727af206402b7122d988dd6633d9f8b70b7/.github/workflows/ar-ci.yml#L29).

More examples can be found via:

```sql
select    * 
from      view_github_actions 
where     action_name like './.github/%';
```

NovemberTang: I did not know that we could do this! thanks!

AshCorr (approved):

PR #947, opened 2024-04-24
Make logging container read-only

NovemberTang:
Co-authored-by: @AshCorr 

## What does this change?

Make the logging container read-only
Add a mount point at the init directory, now that it won't be wiped when the volume is mounted, thanks to this PR in [devx-logs](https://github.com/guardian/devx-logs/pull/40)

## Why?

So we no longer fail a check that requires our root container filesystems to be read only

## How has it been verified?

Tested on CODE, verified that logging is working as expected


tjsilver (approved):
👍

PR #948, opened 2024-04-24
make OTEL container readonly

NovemberTang:
## What does this change?

Forbids the OTEL container job from writing to the root filesystem

## Why?

Security! Hopefully this is the last general-purposed container for CQ jobs that has write access to the filesystem

## How has it been verified?
Screenshot 2024-04-24 at 16 34 45

Ran the OrgWideEc2 job on CODE, verified that traces still show up

Deployed to CODE, and since then, the number of failing controls has dropped by almost half (the remaining half are PROD task definitions, and a couple of specialist containers)


AshCorr (approved):
🎉 🎉 🎉

PR #949, opened 2024-04-25
Create a new "Obligatron" project, which uses 

AshCorr:
## What does this change?

Create a new "Obligatron" project, a step function consisting of 3 lambdas.

- **FindAllTeams**: Fetches all of the `teamId`s from `galaxies_teams` and outputs it to the next step
- **Obligation(Obligation ID)**: Accepts a teamId, runs some business logic using that team ID to check compliance,
- **SaveObligationResults**: 

## Why?

As part of Ops KR1 we've been trying to source tagging data from all AWS resources so that we can measure the departments adherance to the tagging obligation.

This has been going well, but what I've been struggling with is how to transform the data we get into something useable. We're probably going to want to exclude certain resources under certain conditions, for example if we know a resource is managed by AWS we would want to exclude it.

Now we could do this by modifying our SQL function


akash1810:
I think this has been superseded by https://github.com/guardian/service-catalogue/pull/989, and can be closed?

PR #952, opened 2024-04-25
Create object to hold our SLAs

NovemberTang:
## What does this change?

Create a centralised SLA object for repocop, so that the SLAS for each level of vulnerability are clearly visible and configurable. This significantly simplifies the function used to evaluate whether or not a given vulnerability exceeds its SLA.

I've also adjusted the SLA values to reflect Infosec's recommendations. Further tuning may be required to account for weekends, etc.

## Why?

Visibility, reusability.

## How has it been verified?

Ran locally without issue, all CI tests still pass


packages/repocop/src/types.ts
```diff
@@ -163,10 +163,21 @@ export interface SnykProject {
...
+	medium: undefined,
+	low: undefined,
+	unknown: undefined,
+};
+
 export type RepocopVulnerability = Omit<
 	repocop_vulnerabilities,
-	'id' | 'repo_owner'
->;
+	'id' | 'repo_owner' | 'severity'
```
NovemberTang, line 177: Here, we are replacing a `string` type with a union type of the only available values

packages/repocop/src/evaluation/repository.ts
```diff
@@ -217,25 +219,16 @@ function findArchivedReposWithStacks(
...
-
-	if (severity === 'critical') {
-		return date < criticalVulnCutOff;
-	} else if (severity === 'high') {
-		return date < highVulnCutOff;
-	} else {
+	if (daysToRemediate === undefined) {
 		return false;
 	}
+
```
tjsilver, line 228: Might be good to add some tests for this?

tjsilver (approved):
This is a great improvement! One non-blocking comment.

PR #957, opened 2024-04-26
Add more languages that do not pull depencies for repocop to ignore

Reettaphant:
Adding more languages that are not used to pull in dependencies into repocop ignore list so that these repositories are not falsely flagged as not being covered by vulnerability scanning.


tjsilver (commented):
💯

PR #965, opened 2024-04-30
feat: Add Grafana Scenes

akash1810:
## What does this change, and why?
Currently, we're creating Grafana dashboards by hand. Whilst this works, there are a few notable limitations:
- Code sharing is difficult
- Gaining an audit trail is difficult
- Getting a review of a change is not possible (unless pairing)

[Grafana Scenes](https://grafana.com/developers/scenes) is a way to create dashboards (technically, [apps](https://grafana.com/developers/plugin-tools/introduction/plugin-types-usage))  in code. It brings all the benefits of VCS to dashboards. Scenes are implemented as a React app.

## How has it been verified?
TBD.

## TODO
- [ ] Initialise project
- [ ] Add Guardian config (tsconfig, eslint, prettier, etc)
- [ ] Solve the local DX - the Scene should have access to the DEV Postgres database
- [ ] Build project in CI
- [ ] Package project for Riff-Raff
- [ ] Add Scene to CODE/PROD Grafana

PR #967, opened 2024-04-30
Refactor DEV environment 

akash1810:
The current DEV environment has a few limitations, including:
- Installing `libpq` via Homebrew. This package includes the `pg_dump` command. Homebrew doesn't place `pg_dump` on the `PATH`, so one needs to manually add it (albeit a one time thing, if added to `~/.zshrc`)
- CloudQuery runs at the same time as the `pg_dump` from CODE happens. If both are operating on the same table, a race condition is created.

## What does this change?
This change removes the requirement on `libpq`. It is replaces with a new container - `db-copy`. This container:
1. Accepts a list of tables to copy from CODE
2. Loops over the tables, and uses `pg_dump` to create `.sql` files in the `sql` directory
3. Exits

The postgres container is configured to start only once `db-copy` has completed successfully. The `.sql` files created by the `db-copy` container are mounted to `/docker-entrypoint-initdb.d`, which is a special directory:

> After the entrypoint calls `initdb` to create the default `postgres` user and database, it will run any `*.sql` files, run any executable `*.sh` scripts, and source any non-executable `*.sh` scripts found in that directory to do further initialization before starting the service.
> – https://hub.docker.com/_/postgres#:~:text=Initialization%20scripts

That is, the files created by `db-copy` are executed by `postgres` on launch.

## Why?
Simplifying the local DX. This should also make it easier to progress with https://github.com/guardian/service-catalogue/pull/965, as the `start` script is a little easier to reason about.

## How has it been verified?
Ran the script to start the DEV environment:

```sh
npm -w dev-environment start
```

Confirmed that all containers wait until the `db-copy` one has completed. Once all containers have started, one can browse the database via Grafana.


.env
```diff
@@ -37,8 +37,9 @@ STAGE=DEV
 DATABASE_USER=postgres
 DATABASE_PASSWORD=not_at_all_secret
 DATABASE_HOSTNAME=localhost
+DATABASE_DB=postgres
```
NovemberTang, line 40: minor: could we call this something like `DATABASE_ENGINE`, It's current name is a bit of a tautology, and could be a little clearer

akash1810: No. This env var sets `POSTGRES_DB` on the Postgres container, which is defined as:

> This optional environment variable can be used to define a different name for the default database that is created when the image is first started. If it is not specified, then the value of `POSTGRES_USER` will be used.
> – https://hub.docker.com/_/postgres

I'm not even sure the database engine will be anything other than `postgres`.

packages/dev-environment/docker-compose.yaml
```diff
@@ -1,15 +1,56 @@
...
+        - aws_organizations_account_parents
+        - aws_organizations_organizational_units
+        - aws_ec2_instances
+        - aws_cloudformation_stacks
   postgres:
     image: postgres:14.6
     ports:
       - '${DATABASE_PORT}:${DATABASE_PORT}'
     environment:
-      POSTGRES_DB: ${DATABASE_HOSTNAME}
```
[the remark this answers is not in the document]
akash1810: Ah, we had hard-coded the postgres db in the connection string [here](https://github.com/guardian/service-catalogue/blob/5f60acf6bcb7c6df308d64ba46254541c32e23ee/.env#L41) and [here](https://github.com/guardian/service-catalogue/blob/5f60acf6bcb7c6df308d64ba46254541c32e23ee/packages/dev-environment/config/cloudquery.yaml#L49).

We would have had two databases running in the server:
- "localhost" created by the Postgres container on launch
- "postgres" created by CloudQuery (presumably)

Now we'll have one database: "postgres".

.env
marsavar: Yeah, it's the terminal part of the Postgres URI scheme which defines the database name, so perhaps naming the variable `DATABASE_NAME` would be clearer.

containers/db-copy/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -e
+
+if [[ -z "$TABLES" ]]; then
+  echo "TABLES environment variable is not set. Nothing to do."
+  exit 0
+fi
+
+# TODO Use RDS IAM Authentication in favour of the root credentials
+getCodeDatabaseUrl() {
```
marsavar, line 11: I believe the convention is to use snake case for bash functions (see Google [style guide](https://google.github.io/styleguide/shellguide.html#function-names))

containers/db-copy/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+  SECRET_NAME=$(aws secretsmanager list-secrets --filters Key=tag-value,Values=CODE | jq -r '.SecretList[].Name' | grep Postgres)
+
+  SECRET_STRING=$(aws secretsmanager get-secret-value --secret-id "$SECRET_NAME" | jq '.SecretString | fromjson')
+
+  CODE_DATABASE_USER=$(echo "$SECRET_STRING" | jq -r '.username')
+  CODE_DATABASE_PASSWORD=$(echo "$SECRET_STRING" | jq -r '.password')
+  CODE_DATABASE_HOSTNAME=$(echo "$SECRET_STRING" | jq -r '.host')
+  CODE_DATABASE_PORT=$(echo "$SECRET_STRING" | jq -r '.port')
+
+  CODE_DATABASE_URL=postgres://${CODE_DATABASE_USER}:${CODE_DATABASE_PASSWORD}@${CODE_DATABASE_HOSTNAME}:${CODE_DATABASE_PORT}/postgres
```
marsavar, line 21: The the last part of the URI scheme (the DB name) should also be a variable as per changes in this PR, right?

packages/dev-environment/docker-compose.yaml
```diff
@@ -1,15 +1,56 @@
...
+        - github_team_repositories
+        - github_teams
+        - github_repositories
+        - github_repository_branches
+        - github_workflows
+        - github_languages
+        - aws_organizations_account_parents
+        - aws_organizations_organizational_units
+        - aws_ec2_instances
+        - aws_cloudformation_stacks
```
marsavar, line 33: Am I right in thinking that if new migrations are added that import new tables (e.g. aws security groups) this will also need to be updated?
Still new to this project so trying to confirm that I'm understanding this right. :-)

packages/dev-environment/script/start
```diff
@@ -9,101 +9,53 @@ MAIN_ENV_FILE=$ROOT_DIR/.env
...
-insert_into_db() {
-      PGPASSWORD=${DATABASE_PASSWORD}  psql \
-    -U postgres -h "${DATABASE_HOSTNAME}" -p 5432 -d "${DATABASE_NAME}" -f "$1.sql" -q
-    rm "$1.sql"
+startContainers() {
+  step "Launching Docker containers"
 
-    echo "Transfer of $1 data complete ✅"
+  if (docker info) 2>&1 >/dev/null; then
+    docker-compose -f "${DIR}/../docker-compose.yaml" --env-file "$MAIN_ENV_FILE" --env-file "$LOCAL_ENV_FILE" up -d --build
```
marsavar, line 37: `docker-compose` was officially deprecated last year, and `compose` is now a subcommand of Docker (`docker compose`). Maybe not for this PR, but I think it's definitely worth moving to the newer version.

marsavar (commented):
Looks like a good improvement!

containers/db-copy/entrypoint.sh
akash1810: Snake case is so much longer... but also life is too short to debate this 😄. Will update.

NovemberTang (approved):
Looks great!

.env
akash1810: Interestingly, the shape of the AWS Secrets Manager resource is:

```json
{
  "password": "REDACTED",
  "engine": "REDACTED",
  "port": "REDACTED",
  "dbInstanceIdentifier": "REDACTED",
  "host": "REDACTED",
  "username": "REDACTED"
}
```

That is `name` is missing. I think this is because we've not explicitly set the [`databaseName` property](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#databasename), and so it's using the [default](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-dbname) of "postgres".

Relatedly, the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-engine) confirm `engine` is static.

akash1810: Addressed in https://github.com/guardian/service-catalogue/pull/967/commits/02a00ad5f610efa8dcf5021eff622c6b7670ce4a.

containers/db-copy/entrypoint.sh
akash1810: Addressed in https://github.com/guardian/service-catalogue/pull/967/commits/02a00ad5f610efa8dcf5021eff622c6b7670ce4a.

packages/dev-environment/script/start
akash1810: Ah, good spot! Lets do this in a follow-up PR.

packages/dev-environment/docker-compose.yaml
akash1810: Correct. We need to document it somewhere, but there are two methods to collect data:
1. Updating this list will provide data from CODE. It will be stale, but from all accounts
2. Updating the DEV CloudQuery configuration. It will be fresh, but from one account

Ideally the DEV database, and by extension Prisma, will have data that is used by other applications (e.g. RepoCop).

containers/db-copy/entrypoint.sh
akash1810: Addressed in https://github.com/guardian/service-catalogue/pull/967/commits/37fe826e4470fa7c2a4320512ee8573763907113.

PR #968, opened 2024-05-01
QOL: sort vulnerability URLs to show more useful ones first

NovemberTang:
## What does this change?

Prefer snyk or github advisories, if they exist. Vulnerability digests only link out to the first URL in the list, so lets make it one that's the most likely to be helpful

## Why?

The URLs we prioritise are aimed at developers rather than security researchers, and are generally a bit easier to understand.

## How has it been verified?

Added unit tests

### DIFF
--- a/packages/repocop/src/evaluation/repository.test.ts
+++ b/packages/repocop/src/evaluation/repository.test.ts
@@ -846,11 +846,12 @@ describe('NO RULE - Snyk vulnerabilities', () => {
 });
 
 describe('NO RULE - Vulnerabilities from Dependabot', () => {
+	const fullName = 'guardian/myrepo';
+	const result: RepocopVulnerability[] = example.map((alert) =>
+		dependabotAlertToRepocopVulnerability(fullName, alert),
+	);
+
 	test('Should be parseable into a common format', () => {
-		const fullName = 'guardian/myrepo';
-		const result: RepocopVulnerability[] = example.map((alert) =>
-			dependabotAlertToRepocopVulnerability(fullName, alert),
-		);
 		const expected1: RepocopVulnerability = {
 			full_name: fullName,
 			source: 'Dependabot',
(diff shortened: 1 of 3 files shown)


packages/repocop/src/evaluation/repository.test.ts
```diff
@@ -889,6 +890,10 @@ describe('NO RULE - Vulnerabilities from Dependabot', () => {
 
 		expect(result).toStrictEqual([expected1, expected2]);
 	});
+	test('Should display the most useful URLs first', () => {
+		expect(result[0]?.urls[0]).toContain('snyk.io');
+		expect(result[0]?.urls[1]).toContain('github.com');
+	});
```
akash1810, line 896: Rather than array index comparisons, could we compare the complete array? Something like:

```ts
const actual = [1, 3, 4, 2];
const expected = [1, 2, 3, 4];

expect(actual).toEqual(expected);
```

Jest's [`toEqual`](https://jestjs.io/docs/expect#toequalvalue) does a strict comparison (`===`), which for arrays includes order checks.

NovemberTang: Updated the test to reflect this. I've just grabbed an early slice of the array as I wanted to indicate that don't really care about the ordering of the other URLs

packages/repocop/src/test-data/example-dependabot-alerts.ts
```diff
@@ -77,10 +77,10 @@ export const example: Alert[] = [
 					url: 'https://usn.ubuntu.com/3559-1/',
 				},
 				{
-					url: 'https://www.djangoproject.com/weblog/2018/feb/01/security-releases/',
+					url: 'http://www.securitytracker.com/id/1040422',
```
tjsilver, line 80: Is this meant to be 'http'? Is it a dummy URL?

NovemberTang: ah, yeah it doesn't really matter if the URL works or not, I think securitytracker.com isn't even registered any more.

marsavar (approved):

PR #520, opened 2023-11-24
fix: Resolve compilation errors

akash1810:
## What does this change?
RepoCop isn't compiling due to a duplicate import. This wasn't caught in CI as we were only running `tsc` in a workspace if they defined a `typecheck` script (`npm run typecheck --workspaces --if-present`).

When you try to run RepoCop locally on `main` we get:

```console
➜ npm -w repocop start

> repocop@1.0.0 start
> APP=repocop ts-node src/run-locally.ts

/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:859
    return new TSError(diagnosticText, diagnosticCodes, diagnostics);
           ^
TSError: ⨯ Unable to compile TypeScript:
src/index.ts:3:2 - error TS2300: Duplicate identifier 'PrismaClient'.

3  PrismaClient,
   ~~~~~~~~~~~~
src/index.ts:4:2 - error TS2300: Duplicate identifier 'PrismaClient'.

4  PrismaClient,
   ~~~~~~~~~~~~

    at createTSError (/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:859:12)
    at reportTSError (/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:863:19)
    at getOutput (/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:1077:36)
    at Object.compile (/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:1433:41)
    at Module.m._compile (/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:1617:30)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/akash_askoolum/code/service-catalogue/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Function.Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19) {
  diagnosticCodes: [ 2300, 2300 ]
}
npm ERR! Lifecycle script `start` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: repocop@1.0.0 
npm ERR!   at location: /Users/akash_askoolum/code/service-catalogue/packages/repocop 
```

This change does two things:
[...]

PR #969, opened 2024-05-02
fix: Compile each project individually

akash1810:
The `common` package has a compile error:

```
TS2742: The inferred type of createPullRequest cannot be named without a reference to `octokit-plugin-create-pull-request/node_modules/@octokit/types`. This is likely not portable. A type annotation is necessary.
```

It looks like compiling at the root isn't granular enough, as the error wasn't reported during builds.

This change resolves the compilation error, and updates to compiling each project individually.

This was discovered as part of #965.

## How has it been verified?
See the build logs.


package.json
```diff
@@ -6,7 +6,8 @@
     "test": "npm run test --workspaces --if-present",
     "test-update": "jest -u",
     "synth": "npm run synth --workspace=cdk",
-    "typecheck": "tsc",
+    "pretypecheck": "npm -w common run typecheck",
+    "typecheck": "npm run typecheck --workspaces",
```
akash1810, line 10: Its worth noting we were previously compiling each project individually (see https://github.com/guardian/service-catalogue/pull/520). The difference now though is we're not using the `--if-present` flag, so the concerns mentioned in https://github.com/guardian/service-catalogue/pull/520 do not hold.

package.json
```diff
@@ -6,7 +6,8 @@
     "test": "npm run test --workspaces --if-present",
     "test-update": "jest -u",
     "synth": "npm run synth --workspace=cdk",
-    "typecheck": "tsc",
+    "pretypecheck": "npm -w common run typecheck",
```
akash1810, line 9: `pretypecheck` will always run before `typecheck` see https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts. We need to compile the shared package first, as other projects depend on it, and it doesn't look like TypeScript understands this dependency.

packages/common/package.json
```diff
@@ -4,7 +4,8 @@
 	"type": "module",
 	"scripts": {
 		"postinstall": "prisma generate",
-		"test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects common"
+		"test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects common",
+		"typecheck": "tsc"
```
akash1810, line 8: When compiling the `common` package, write the resulting files to disk (i.e. no `--noEmit` flag), so that it can be used by the other projects during their compile.

packages/common/src/pull-requests.ts
```diff
@@ -46,6 +52,8 @@ export async function createPullRequest(
 			files,
 		})),
 	});
+
+	return data?.data.html_url;
```
akash1810, line 56: This is the main change here, and resolves the compilation error in the PR description.

packages/dev-environment/package.json
```diff
@@ -3,7 +3,8 @@
 	"version": "0.0.0",
 	"scripts": {
 		"start": "./script/start",
-		"test": "find . | egrep '.yml|.yaml' | xargs yamllint"
+		"test": "find . | egrep '.yml|.yaml' | xargs yamllint",
+		"typecheck": "echo 'Nothing to typecheck'"
```
akash1810, line 7: This package has no TypeScript code, so there's nothing to do.

packages/diagrams/package.json
```diff
@@ -4,7 +4,8 @@
 	"description": "Diagrams as code",
 	"type": "module",
 	"scripts": {
-		"generate": "mmdc -i diagram.mmd -o output.svg "
+		"generate": "mmdc -i diagram.mmd -o output.svg ",
+		"typecheck": "echo 'Nothing to typecheck'"
```
akash1810, line 8: This package has no TypeScript code, so there's nothing to do.

packages/common/src/pull-requests.ts
```diff
@@ -46,6 +52,8 @@ export async function createPullRequest(
 			files,
 		})),
 	});
+
+	return response?.data.html_url;
```
akash1810, line 56: This is the main change here, and resolves the compilation error in the PR description. The return type of `composeCreatePullRequest` is a bit tricky to infer.

We're (currently) only using the `html_url` property, so update `createPullRequest` to return that, and therefore the return type of `createPullRequest` becomes `Promise<string | undefined>`.

packages/common/src/pull-requests.ts
```diff
@@ -23,18 +23,24 @@ export function generateBranchName(prefix: string) {
...
 }
 
+/**
+ * Creates or updates a pull request, and return its URL.
+ * On error, an exception is thrown, or undefined is returned.
+ */
 export async function createPullRequest(
 	octokit: Octokit,
-	{
+	props: CreatePullRequestOptions,
```
akash1810, line 32: This change moves the object destructing out of the function signature, in an attempt to improve readability.

NovemberTang (commented):
Is it worth doing a test run of snyk-integrator to make sure it still behaves as expected?

akash1810:
> Is it worth doing a test run of snyk-integrator to make sure it still behaves as expected?

Absolutely! How? Is `createPullRequest` called on CODE, as IIUC CODE doesn't raise PRs? Is there a switch to toggle to change this?

NovemberTang:
> > Is it worth doing a test run of snyk-integrator to make sure it still behaves as expected?
> 
> Absolutely! How? Is `createPullRequest` called on CODE, as IIUC CODE doesn't raise PRs? Is there a switch to toggle to change this?


Yeah there's a couple of ways to do this. You can either:

- Temporarily modify the Stage env var in the console to from CODE to PROD, bypassing the flag
- Deploy your branch to PROD
- Remove the stage awareness in the source code, build it, and paste it into the text editor in the console

akash1810:
> * Deploy your branch to PROD

[I've done this](https://riffraff.gutools.co.uk/deployment/view/702ec16a-7140-4e40-baec-1e28908da1de), with a resulting PR of https://github.com/guardian/service-catalogue/pull/976.

@NovemberTang and I have been discussing the snyk-integrator over chat. Here's a summary for others:

In general masquerading as PROD should be avoided as there are potentially unknown side-effects. For example, imagine there are multiple stage aware branches in the application, we'll trigger all of them when only wanting to test one. Testing on PROD isn't ideal, for similar reasons.

Mutating the lambda within the AWS console should also be avoided, as it removes the audit trail, and introduces elements of human error. Updating the lambda with a traditional deployment means all the CI checks have passed, increasing confidence.

In this specific scenario, only one piece of the application is stage aware, and that is the part we are wanting to test. Whilst there are opportunities to refactor parts of the code to create more pure functions, it is most pragmatic to test on PROD, as this PR unblocks others. We will make the refactors in follow up PRs.

NovemberTang (approved):

PR #986, opened 2024-05-07
Make prisma container readonly

NovemberTang:
## What does this change?

Makes all parts of the prisma container readonly, except for the ./prisma directory, where all files are now loaded

## Why?

Security!

## How has it been verified?

The change has been deployed to CODE and tested to make sure no errors occur while the script is being run


marsavar (approved):
Looks good!

PR #987, opened 2024-05-09
Apply migrations to Dev environment

AshCorr:
## What does this change?

Apply Prisma migrations to dev environment when running the `start` script.

## Why?

We want to be able to use the functions contained in our prisma migrations for developing new things locally!

## How has it been verified?

Ran locally.

### DIFF
--- a/packages/dev-environment/script/start
+++ b/packages/dev-environment/script/start
@@ -15,6 +15,20 @@ step() {
   echo -e "\n${cyan}Step ${STEP_COUNT}${clear}: $1"
 }
 
+create_sql_migrations() {
+  step "Copying Prisma Migrations"
+
+  SQL_DIR=$ROOT_DIR/packages/dev-environment/sql
+
+  rm -r $SQL_DIR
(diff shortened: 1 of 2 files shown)


packages/dev-environment/script/start
```diff
@@ -15,6 +15,20 @@ step() {
...
+  step "Copying Prisma Migrations"
+
+  SQL_DIR=$ROOT_DIR/packages/dev-environment/sql
+
+  rm -r $SQL_DIR
+  mkdir -p $SQL_DIR
+
+  MIGRATIONS=$(ls -d $ROOT_DIR/packages/common/prisma/migrations/*/)
+  for MIGRATION in $MIGRATIONS; do
+    cp $MIGRATION/migration.sql $SQL_DIR/$(basename $MIGRATION).sql;
```
AshCorr, line 28: Annoyingly the Docker image doesn't support having .sql files in nested folders, otherwise we could just mount the migrations dir into the container.

akash1810 (approved):

PR #988, opened 2024-05-09
aws_resources materialised view correct taggable flad for iam policies

JuliaBrigitte:
## What does this change?
The materialised view aws_resources showed iam policies as tageable even though they are not, this migration corrects this.

This migration changes the view aws_resources but without populating it.
This can not be done after the migration was applied with
select refresh_aws_resources()

## SQL Changes
Changes to the sql of the previous migration:
function arn_taggable to return false for aws managed iam policies
aws_resources_raw uses arn_taggable (sadly I had to include the whole function not just the change)
I also added refresh_aws_resources() with I had planned to use in a lamda a db trigger, but kept.
I now create the materialised view without data in case I will add further migrations during the quarter since I feel data dumping should not happen in the migrations

## Why?
The aws_resources view powers our [OKR Dashboard DevX Ops OKRs Q1 2024](https://metrics.gutools.co.uk/d/adig3ssv1yio0c/8c14ffdd-c03a-51ad-90fc-1acb28ab7836?orgId=1) for this quarter.

We will move to repocop style lambda powered tables but need the view for this quarter.

## How has it been verified?
Ran the sql locally

### DIFF
--- a/packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
+++ b/packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
@@ -5,7 +5,7 @@
 -- This migration replaces this view with a function which doesn't have the same restriction.
 -- But will need a cron job to keep the materialized view up to date.
 
--- Decide if the resource is taggable when taggable column is present
+-- Check for aws managed iam policies and return false for taggable in that case
 CREATE OR REPLACE FUNCTION is_resource_taggable(arn text, taggable bool) RETURNS boolean AS $$
 BEGIN
     RETURN CASE WHEN arn LIKE '%arn:aws:iam::aws:policy%' THEN 'false'
--- a/packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
+++ b/packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
@@ -5,24 +5,13 @@
 -- This migration replaces this view with a function which doesn't have the same restriction.
 -- But will need a cron job to keep the materialized view up to date.
 
--- Decide if taggable when taggable column is present
-CREATE OR REPLACE FUNCTION arn_taggable(arn text, taggable bool) RETURNS boolean AS $$
+-- Decide if the resource is taggable when taggable column is present
+CREATE OR REPLACE FUNCTION is_resource_taggable(arn text, taggable bool) RETURNS boolean AS $$
 BEGIN
     RETURN CASE WHEN arn LIKE '%arn:aws:iam::aws:policy%' THEN 'false'
                 ELSE taggable END;
 END
 $$ LANGUAGE plpgsql;
-
--- Function to refresh the materialized view aws_resources
-CREATE OR REPLACE FUNCTION refresh_aws_resources()
-    RETURNS TEXT AS
-$$
-BEGIN
-    REFRESH MATERIALIZED VIEW aws_resources WITH DATA;
-    RETURN 'Triggered refresh of materialized view aws_resources';
-END;
-$$
-    LANGUAGE plpgsql;
 
 -- Aggregate rows from all AWS Cloudquery tables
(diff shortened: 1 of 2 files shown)


packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
```diff
@@ -0,0 +1,120 @@
...
+-- Decide if taggable when taggable column is present
+CREATE OR REPLACE FUNCTION arn_taggable(arn text, taggable bool) RETURNS boolean AS $$
+BEGIN
+    RETURN CASE WHEN arn LIKE '%arn:aws:iam::aws:policy%' THEN 'false'
+                ELSE taggable END;
+END
+$$ LANGUAGE plpgsql;
+
+-- Function to refresh the materialized view aws_resources
+CREATE OR REPLACE FUNCTION refresh_aws_resources()
```
akash1810, line 17: It looks like this function isn't in use? Indeed the PR description describes future plans to use it. Should we therefore remove it from this PR, as it's an unrelated change?[^1]

The PR describes the future usage of this function would be via a lambda. Given the small size of the function, might be be possible for the lambda to execute `REFRESH MATERIALIZED VIEW` directly?

[^1]: https://github.com/guardian/recommendations/blob/main/pull-requests.md#small

packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
```diff
@@ -0,0 +1,120 @@
+-- Update materialized view aws_resources creation to correct ARN taggable column for iam policies
+
+-- Cloudquery can't update the schema of tables referenced by views.
+-- As the view aws_resources references any Cloudquery table containing an ARN column it can make Cloudquery behave strangely.
+-- This migration replaces this view with a function which doesn't have the same restriction.
+-- But will need a cron job to keep the materialized view up to date.
+
+-- Decide if taggable when taggable column is present
+CREATE OR REPLACE FUNCTION arn_taggable(arn text, taggable bool) RETURNS boolean AS $$
```
akash1810, line 9: This function is a little confusing; the comment doesn't match the body, and the name doesn't quite match either. Might `is_managed_policy` be a clearer function name?

Alternatively, could this be inlined to the case statement below?

```sql
 -- Check if table has an tags column, and not an AWS managed IAM policy
taggable := CASE WHEN table_has_column(cloudquery_table.table_name, 'tags') THEN
  CASE WHEN arn LIKE 'arn:aws:iam::aws:policy%' THEN 'false' ELSE 'true'
  ELSE 'false' 
END;

tags := CASE WHEN taggable = 'true' THEN 'tags' ELSE '''{}''::jsonb' END;
```

JuliaBrigitte: I tried to inline but that got me in the quotation hell (which was why I made it a function in the first place). So I went for a rename to make it clearer.

packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
JuliaBrigitte: I deleted it.

packages/common/prisma/migrations/20240508195000_refresh_aws_resources/migration.sql
```diff
@@ -0,0 +1,109 @@
+-- Update materialized view aws_resources creation to correct ARN taggable column for iam policies
+
+-- Cloudquery can't update the schema of tables referenced by views.
+-- As the view aws_resources references any Cloudquery table containing an ARN column it can make Cloudquery behave strangely.
+-- This migration replaces this view with a function which doesn't have the same restriction.
+-- But will need a cron job to keep the materialized view up to date.
+
+-- Decide if the resource is taggable when taggable column is present
```
akash1810, line 8: minor: this comment doesn't match the function it's against

akash1810 (approved):
Looks good 👍🏽.

PR #989, opened 2024-05-09
Create "obligatron" lambda

AshCorr:
## What does this change?

Creates a new "obligatron" lambda which evaluates service-catalogue data and decides if its compliant with obligations.

## Why?

This lays some of the groundwork for writing the business logic for some of the obligations the Ops team is working on (and maybe Security and Reliability too?).

Eventually this lambda should provide a framework for fetching data from Service Catalogue and saving it to a results table in a generic way so that when writing a new obligation the developer only has to worry about parsing the data, and not establish a DB connection or saving the results to a table.

In the short term this lambda will evaluate the Ops TAGGING obligation by fetching data from the aws_resources view, filtering out AWS managed resources, and validating them against our tagging schema.

## How has it been verified?

Ran locally.

### DIFF
--- a/packages/cdk/lib/obligatron.ts
+++ b/packages/cdk/lib/obligatron.ts
@@ -30,6 +30,8 @@ export class Obligatron {
 				QUERY_LOGGING: 'false', // Set this to 'true' to enable SQL query logging
 			},
 			timeout: Duration.minutes(5),
+			// Unfortunately Prisma doesn't support streaming data from Postgres at the moment https://github.com/prisma/prisma/issues/5055
+			// This means that all rows need to be loaded into memory at the same time whenever a query is ran hence the high memory requirement.
 			memorySize: 4096,
 		});
 
--- a/packages/cdk/lib/obligatron.ts
+++ b/packages/cdk/lib/obligatron.ts
@@ -1,0 +1,38 @@
+import type { GuStack } from '@guardian/cdk/lib/constructs/core';
+import type { GuSecurityGroup } from '@guardian/cdk/lib/constructs/ec2';
+import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
+import { Duration } from 'aws-cdk-lib';
+import type { IVpc } from 'aws-cdk-lib/aws-ec2';
+import { Architecture, Runtime } from 'aws-cdk-lib/aws-lambda';
+import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
+
+type ObligatronProps = {
+	vpc: IVpc;
+	dbAccess: GuSecurityGroup;
+	db: DatabaseInstance;
+};
+
+export class Obligatron {
+	constructor(stack: GuStack, props: ObligatronProps) {
+		const { vpc, dbAccess, db } = props;
+		const app = 'obligatron';
+
+		const lambda = new GuLambdaFunction(stack, 'obligatron', {
+			app,
+			vpc,
+			architecture: Architecture.ARM_64,
+			runtime: Runtime.NODEJS_20_X,
+			securityGroups: [dbAccess],
+			fileName: `${app}.zip`,
+			handler: 'index.main',
+			environment: {
+				DATABASE_HOSTNAME: db.dbInstanceEndpointAddress,
+				QUERY_LOGGING: 'false', // Set this to 'true' to enable SQL query logging
+			},
+			timeout: Duration.minutes(5),
+			memorySize: 4096,
+		});
+
+		db.grantConnect(lambda, 'obligatron');
+	}
+}
--- a/packages/obligatron/src/obligations/index.ts
+++ b/packages/obligatron/src/obligations/index.ts
@@ -1,0 +1,33 @@
+export type ObligationResult = {
+	/**
+	 * Resource identifier. Varies depending on resource platform.
+	 *   - Github -> Slug
+	 *   - AWS -> ARN
+	 */
+	resource: string;
+
+	/**
+	 * Did this resource meet the obligation check?
+	 */
+	result: boolean;
+
+	/**
+	 * Explanation for the assessment failing.
+	 */
+	reasons: string[];
+
+	/**
+	 * Link to where the user can see more details on the resource.
+	 */
+	deep_link?: string;
+
+	/**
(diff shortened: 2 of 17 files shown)


packages/cdk/lib/obligatron.ts
```diff
@@ -0,0 +1,38 @@
...
+			runtime: Runtime.NODEJS_20_X,
+			securityGroups: [dbAccess],
+			fileName: `${app}.zip`,
+			handler: 'index.main',
+			environment: {
+				DATABASE_HOSTNAME: db.dbInstanceEndpointAddress,
+				QUERY_LOGGING: 'false', // Set this to 'true' to enable SQL query logging
+			},
+			timeout: Duration.minutes(5),
+			memorySize: 4096,
```
AshCorr, line 33: [Unfortunately Prisma doesn't support streaming rows](https://github.com/prisma/prisma/issues/5055), this means we have to load all of our query results into memory at the same time before we can process them.

In future we could potentially write our own pagination, but I want to avoid that for now. Testing in CODE this function takes less than a minute to run so even with 4GB of memory its not particularly expensive.

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -0,0 +1,56 @@
...
+	tags['gu:repo'] !== undefined;
+
+const isExemptResource = (resource: AwsResource): boolean => {
+	if (resource.resource_type === 'role') {
+		// AWS Creates roles for various services when onboarding them.
+		// Technically we can tag these but they wouldn't belong to a specific App, Stage, or Stack.
+		return resource.arn.includes('/aws-service-role/');
+	}
+
+	return false;
```
AshCorr, line 26:
```suggestion
	return ['ec2', 'iam'].contains(resource.service);
```

Exempt everything but EC2 and IAM for the purpose of keeping the Obligation simple for this quarter?

packages/obligatron/src/obligations/tagging.ts
akash1810, line 1: Should we add some tests around this logic?

akash1810 (commented):
Looks great. Shall we walk @guardian/devx-security and @guardian/devx-reliability through these changes? And plan how to migrate any current obligations to this framework?

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export async function evaluateTaggingObligation(
+	db: PrismaClient,
+): Promise<ObligationResult[]> {
+	// TODO add `where taggable = true`?
```
adamnfish, line 28: Is this TODO still valid? (it looks like this clause has already been included)

packages/cdk/lib/obligatron.ts
adamnfish: Is it worth including a comment here to this effect? I can imagine someone dropping this number in future and then finding out the long way that we need lots of memory here!

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -0,0 +1,60 @@
...
+		GROUP BY account_id, arn, service, resource_type;
+  `;
+
+	const result: ObligationResult[] = awsResources.map((resource) => {
+		const reasons = !isExemptResource(resource)
+			? REQUIRED_TAGS.filter(
+					(tag) =>
+						typeof resource.tags === 'object' &&
+						resource.tags !== null &&
+						resource.tags[tag] === undefined,
```
adamnfish, line 48: Do we need to check they aren't empty as well, in practice? No harm for now!

packages/obligatron/src/obligations/tagging.ts
AshCorr: It is! I've temporarily added `taggable = true`, but this comment needs refining.

Theres a decision that needs to be made here about what we record in our obligation results:
   A. Only record resources that fail the obligation
   B. Record everything except resources the obligation doesn't apply to. So untaggable resources should be excluded.
   C. Record everything

akash1810: I think B makes sense here.

With A, I think we'd lose some historical accuracy. For example, we wouldn't be able to say of all resources that existed on day X, how many were tagged correctly.

C could result in an amount of noise, that'd need filtering out in the dashboards.

packages/obligatron/src/obligations/index.ts
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * Associated AWS account ID (if any)
+	 */
+	aws_account_id?: string;
+
+	/**
+	 * Associated Github teams (if any)
+	 */
+	github_teams?: string[];
+};
```
akash1810, line 33: Following the ADR comments, should we combine this into a `place of origin` field, of type `jsonb` (`Record<string, unknown>` in TS land)?

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -0,0 +1,60 @@
...
+	arn: string;
+	service: string;
+	resource_type: string;
+	taggable: string;
+	tags?: Record<string, string> | null;
+}
+
+const REQUIRED_TAGS = ['Stack', 'Stage', 'App', 'gu:repo'] as const;
+
+const isExemptResource = (resource: AwsResource): boolean => {
```
akash1810, line 15: Not a blocker for this PR, I wonder if switching on the resource would make it easier to assert those rules that apply to specific resource? For example:

```ts
switch (resourceType): {
  case IAM_USER: {
    return evaluateStandardTags && evaluateIamUserTags;
  }
  case EC2_AMI: {
    return evaluateStandardTags && evaluateEc2AmiTags;
  }
  default: {
    return evaluateStandardTags;
  }
}
```

akash1810 (approved):
Couple of minor, non-blocking comments.

packages/obligatron/src/obligations/index.ts
AshCorr: Will do this in a follow up PR!

AshCorr:
Going to merge this down as is as I want to get some code merged in and not keep updating this PR. There will likely be some changes once the ADR is published and theres a good chance we'll switch to pulling data from Security Hub instead of our AWS Resources view once we've finished figuring out how we turn that on for all accounts.

PR #990, opened 2024-05-09
Fix failing auto migrate

JuliaBrigitte:
## What does this change?
Changes to path to /usr/src/app/prisma as the one that is readonly false

## Why?
automatic migration failed with the following message
download failed: s3://deploy-tools-dist/deploy/CODE/service-catalogue-prisma-migrations/prisma.zip to prisma/prisma.zip Could not create directory /usr/src/app/prisma: [Errno 30] Read-only file system: '/usr/src/app/prisma'

## Why was this difficult to diagnose
./run-prisma-migrate.sh: line 15: prisma/node_modules/.bin/prisma: not found
seemed pretty obvious to fix but no matter what I changed in run-prisma-migrate.sh nothing made a difference.
Until I became convinced my changes did not take effect and changed the messages on the log and sure enough, they did not change.
But then I found https://github.com/guardian/service-catalogue/pull/865
So it seems a new image is created but not used:
"
At the moment, each image gets pushed to GHCR tagged with the full commit sha, e.g.

ghcr.io/guardian/service-catalogue/prisma-migrate:sha-42771e43c3130daa7476d9e5b77455001533c557

This will be manually referenced within the CDK, when the task definition that uses this image is defined.
"
This seems non obvious.

How do we make it better, at the very least I would expect some mention in the READMEs and a comment in the run-prisma_migrate.sh about the need to manually reference a new image.


AshCorr (approved):

containers/prisma-migrate/run-prisma-migrate.sh
```diff
@@ -2,14 +2,22 @@
...
 export DATABASE_URL=postgres://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/postgres
 
-echo 'Running prisma migrate deploy'
-prisma/node_modules/.bin/prisma migrate deploy
\ No newline at end of file
+# Check if prisma is installed
+if ! [ -x "$(command -v ./node_modules/.bin/prisma)" ]; then
+  echo 'Prisma is not installed. Installing...'
+  npm install @prisma/cli --save-dev
+fi
```
akash1810, line 20: Curious why we need to install Prisma as it [should already be installed](https://github.com/guardian/service-catalogue/blob/778a027a4640d0f8eecd97c8e0a0ec51038b0f49/containers/prisma-migrate/Dockerfile#L13).

containers/prisma-migrate/run-prisma-migrate.sh
```diff
@@ -2,14 +2,22 @@
...
+
+echo 'Step3: Running prisma migrate deploy'
+./node_modules/.bin/prisma migrate deploy
```
akash1810, line 23: I wonder if using the [full path](https://github.com/guardian/service-catalogue/blob/778a027a4640d0f8eecd97c8e0a0ec51038b0f49/containers/prisma-migrate/Dockerfile#L3C9-L3C21) would increase certainty, removing the above `if` check?

```sh
/usr/src/app/node_modules/.bin/prisma migrate deploy
```

akash1810: We could also do the same for the S3 download above:

```sh
mkdir -p /usr/src/app/prisma
aws s3 cp "s3://$ARTIFACT_BUCKET/$PRISMA_ARTIFACT_KEY" /usr/src/app/prisma/
```

packages/cdk/lib/prisma-migrate-task.ts
```diff
@@ -146,6 +146,13 @@ export function addPrismaMigrateTask(
 		readOnly: false,
 	});
 
+	prismaTask.addMountPoints({
+		// So that we can download the prisma.zip from the artifact bucket
+		containerPath: '/usr/src/app/prisma',
```
akash1810, line 151: Are we able to remove the above mount point at `/prisma` now?

containers/prisma-migrate/run-prisma-migrate.sh
JuliaBrigitte: This was when I still thought the changes I applied would take effect.
See new section "Why was this difficult to diagnose"

packages/cdk/lib/prisma-migrate-task.ts
JuliaBrigitte: Again, I added that when I searched for the issue it did not work, because at that point I thought I remembered Natasha said, we need to add another point not change it, but this was not the problem at all.
See new section in description Why was this difficult to diagnose.

JuliaBrigitte:
Ashleigh Carr, 21 min

Thats right yea. You need to update the image version here https://github.com/guardian/service-catalogue/blob/ce521603694538cd175e14f76fe979563590dc64/packages/cdk/lib/cloudquery/images.ts#L27-L29 to point to the image built in CI

Akash Askoolum, 20 min

That's quite the long feedback loop! Are we able to shorten it by running the container locally?

You, 18 min

We do not run migrations locally at the moment but Ash has a PR about it I think.

Akash Askoolum, 14 min
, Edited
Due to some synchronisation issues, I don't think we'd be able to use the container to perform migrations in dev as part of the standard dev-environment, currently.

It might be possible, however, to run the prisma container ad-hoc against a new postgres database, as a way of testing changes quickly (vs pushing, waiting for CI, and deploying to CODE).

You, 13 min

From what I understand you would have to deploy, then change to the new image and then deploy again.

Akash Askoolum, 11 min

Indeed. That's quite a long feedback loop!

containers/db-copy/entrypoint.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/ash
```
NovemberTang, line 1: is bin/ash a valid shell?

JuliaBrigitte:
Solved in new PR

containers/db-copy/entrypoint.sh
akash1810: `ash` is the default shell of in alpine linux - https://en.wikipedia.org/wiki/Almquist_shell.

PR #997, opened 2024-05-13
fix(prisma): Correct volume permissions

akash1810:
Since https://github.com/guardian/service-catalogue/pull/986 the Prisma migration task has been failing with:

```log
download failed: s3://<REDACTED>/deploy/CODE/service-catalogue-prisma-migrations/prisma.zip to prisma/prisma.zip Could not create directory /usr/src/app/prisma: [Errno 30] Read-only file system: '/usr/src/app/prisma'
```

That is, we're unable to download a file from S3 to a read-only path.

This change:
- Mounts the `/usr/src/app/prisma` volume in the container, as a writable location
- Updates the container's entry point to download the artifact to this location

This results in the above error being resolved, and the migrations completing. Indeed the logs show this:

image

An additional change was made to better orchestrate the deployment order. See the inline comment for more detail.

## How has it been verified?
The updated container was deployed to, and ran on, CODE. See the above screenshot for the logs.

---
Co-authored-by: @JuliaBrigitte
Co-authored-by: @NovemberTang

### DIFF
--- a/containers/prisma-migrate/Dockerfile
+++ b/containers/prisma-migrate/Dockerfile
@@ -10,13 +10,13 @@ RUN apk --no-cache add aws-cli
 # This will allow us to use the prisma migrate deploy command
 RUN npm init -y
 # TODO: Keep the version automatically in sync with common/package.json
-RUN npm install prisma@5.11.0 
+RUN npm install prisma@5.11.0
 
-COPY ./containers/prisma-migrate/run-prisma-migrate.sh ./run-prisma-migrate.sh
-RUN chmod +x ./run-prisma-migrate.sh
+COPY ./containers/prisma-migrate/run-prisma-migrate.sh /usr/src/app/run-prisma-migrate.sh
+RUN chmod +x /usr/src/app/run-prisma-migrate.sh
 
 # The default entrypoint for the node image is `node ...`
 # Switch it back here so we can run a .sh script
 ENTRYPOINT [ "sh" ]
 
-CMD ["./run-prisma-migrate.sh"]
+CMD ["/usr/src/app/run-prisma-migrate.sh"]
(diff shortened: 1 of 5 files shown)


containers/prisma-migrate/Dockerfile
```diff
@@ -10,13 +10,13 @@ RUN apk --no-cache add aws-cli
 # This will allow us to use the prisma migrate deploy command
 RUN npm init -y
 # TODO: Keep the version automatically in sync with common/package.json
-RUN npm install prisma@5.11.0 
+RUN npm install prisma@5.11.0
 
-COPY ./containers/prisma-migrate/run-prisma-migrate.sh ./run-prisma-migrate.sh
-RUN chmod +x ./run-prisma-migrate.sh
+COPY ./containers/prisma-migrate/run-prisma-migrate.sh /usr/src/app/run-prisma-migrate.sh
```
akash1810, line 15: Using the full file path isn't strictly necessary as [`WORKDIR`](https://docs.docker.com/reference/dockerfile/#workdir) is set above. However, it makes it easier to understand exactly what is being done.

JuliaBrigitte (approved):

packages/cdk/lib/prisma-migrate-task.ts
```diff
@@ -133,16 +133,16 @@ export function addPrismaMigrateTask(
 	db.grantConnect(taskDefinition.taskRole);
 	artifactBucket.grantRead(taskDefinition.taskRole, prismaArtifactKey);
 
-	const prismaVolume: Volume = {
-		name: 'prisma-volume',
+	const prismaArtifactVolume: Volume = {
+		name: 'artifact-volume',
```
akash1810, line 137: Renamed to "artifact" to distinguish this from the location the prisma binary is installed; "artifact" also matches the language used in other applications.

PR #1003, opened 2024-05-14
Collect data for all AWS regions

marsavar:
## What does this change?

Removes the regions array from the AWS ingestion configuration. This makes CQ ingest data for all regions.

## Why?
We're deprecating security-hq, which flags open security groups for regions not currently ingested by CQ. We want to make sure service-catalogue can provide the same level of information.

## How has it been verified?
Tested on CODE.

Previously, the following query returned 0 rows.
```sql
SELECT * FROM aws_ec2_security_groups WHERE region = 'ap-northeast-2'
```

After deploying this branch to CODE and kicking off the `OrgWideEc2` task, it now return multiple results. Sync time also matches the time I kicked off the task.

## Notes
We might want to keep an eye of how this affects the ingestion performance in PROD once merged to main.


packages/cdk/lib/cloudquery/config.ts
```diff
@@ -1,5 +1,4 @@
 import { GuardianOrganisationalUnits } from '@guardian/private-infrastructure-config';
-import { AWS_REGIONS } from 'common/src/aws';
```
NovemberTang, line 2: This is also being used in the data-audit lambda, to check the cq numbers match what we get from the API. As part of this PR, could we remove the region restriction from [that part of the code](https://github.com/guardian/service-catalogue/blob/95dc8ee685665a7fdcde97784a16acf62040473d/packages/data-audit/src/audit/aws-lambda.ts#L85)? 
cc: @adamnfish

marsavar: Looking at [this SCP](https://github.com/guardian/infosec-platform/blob/main/policies/DenyEC2ECSLambdaAccessToNonApprovedRegions.json#L10) we block lambda creation in unapproved regions so removing that check probably won't make a difference in the data returned by the data audit lambda (unless we have some rogue lambdas that were created before the SCP was put into place years ago)

AshCorr (approved):
Looks good!

I would suggest running the All AWS tables job once this is merged to make sure that this doesn't cause the job to take too long! Shouldn't do as you mentioned most regions have no data.

tjsilver (approved):
Looks good!

PR #1023, opened 2024-05-22
Bump Cloudquery to 5.19 and enable Sync Summary

AshCorr:
## What does this change?

Bumps Cloudquery to 5.19 and enable Sync Summary for all jobs.

## Why?

Cloudquery 5.19 just [released the ability to enable sync summaries](https://github.com/cloudquery/cloudquery/issues/8690#issuecomment-2120716012). This feature sends a summary of the sync to the destination plugin in the form of a `cloudquery_sync_summary` table.

We can possibly simplify our Cloudquery job by removing the initial postgres step. This also lets us generate a reliability report for our Cloudquery tasks.

## How has it been verified?

Deployed to CODE.

### DIFF
--- a/packages/cdk/lib/cloudquery/images.ts
+++ b/packages/cdk/lib/cloudquery/images.ts
@@ -3,7 +3,7 @@ import { Versions } from './versions';
 
 export const Images = {
 	cloudquery: ContainerImage.fromRegistry(
-		`ghcr.io/guardian/service-catalogue/cloudquery:stable`,
+		`ghcr.io/guardian/service-catalogue/cloudquery:sha-2d6c95ee216c4fa9723aa3d0f8dc4e2a4636080b`,
 	),
 	devxLogs: ContainerImage.fromRegistry('ghcr.io/guardian/devx-logs:2'),
 	singletonImage: ContainerImage.fromRegistry(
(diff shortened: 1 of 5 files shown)


marsavar (approved):

packages/cdk/lib/cloudquery/images.ts
```diff
@@ -3,7 +3,7 @@ import { Versions } from './versions';
 
 export const Images = {
 	cloudquery: ContainerImage.fromRegistry(
-		`ghcr.io/guardian/service-catalogue/cloudquery:stable`,
+		`ghcr.io/guardian/service-catalogue/cloudquery:sha-2d6c95ee216c4fa9723aa3d0f8dc4e2a4636080b`,
```
akash1810, line 6: Why is this change needed?

AshCorr: Remnants from trying to test this branch in CODE, it can be reverted.

As is this PR doesn't currently work, [in the original CloudQuery PR they suggest adding a `send_sync_summary` field to the destination plugin config](https://github.com/cloudquery/cloudquery/issues/8690#issuecomment-2120716012) but that doesn't seem to be correct

PR #1034, opened 2024-05-24
fix(dev-environment): Wait for Postgres to initialise before running CloudQuery

akash1810:
## What does this change?
In the DEV environment, CloudQuery fails to connect to the Postgres database, logging:

```log
failed to init destination postgresql: rpc error: code = Internal desc = failed to init plugin: failed to initialize client: failed to get current database: failed to connect to `host=postgres user=postgres database=postgres`: dial error (dial tcp 172.26.0.2:5432: connect: connection refused)
```

This is because Postgres is still initialising when the connection is attempted. This initialisation includes creating the tables, and inserting rows copied from CODE.

In this change, a healthcheck is added to the Postgres container, passing only once the initialisation has completed. CloudQuery (and Grafana) starts afterwards.

Additionally, the configuration is updated, getting the CloudQuery container to connect to the Postgres container via the Docker network, as opposed to going via the host.

## How has it been verified?
- Start the dev-environment on `main`, observe the above error from the CloudQuery container
- Start the dev-environment on this branch, observe the CloudQuery container running

### DIFF
--- a/.env
+++ b/.env
@@ -41,6 +41,10 @@ DATABASE_NAME=postgres
 DATABASE_PORT=5432
 DATABASE_URL=postgres://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}
 
+# Locally, CloudQuery and Postgres run within Docker.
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
Great work! Tested locally on `main` and this branch and observed the error go away 🥇

PR #1035, opened 2024-05-24
feat(cloudquery): Collect more Security Hub findings

akash1810:
Remove the `severity_normalized` filter when collecting AWS Security Hub findings. We're using the Tagging Standard for the tagging obligation. These findings are reported with a severity of 0; removing the filter allows them to be collected in Service Catalogue.

The `severity_normalized` filter is also [deprecated](https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_Severity.html).

## How has it been verified?
Ran locally; it does not take significantly longer to collect this data either.


tjsilver (approved):
Looks good 👍

PR #955, opened 2024-04-25
docs: Add ADR for monitoring the tagging obligation

akash1810:
## What does this change?
Add an architecture decision record (ADR), for prosperity, describing the approaches considered to monitor the tagging obligation.

## Why?
See https://github.com/guardian/recommendations/blob/main/architecture-decision-records.md.

PR #1036, opened 2024-05-24
docs: Update ADR for tagging obligation

akash1810:
## What does this change?
Extends the ADR from https://github.com/guardian/service-catalogue/pull/955, to document the decision to use AWS Security Hub to identify tagging usage.


tjsilver (approved):
👍

PR #1048, opened 2024-05-30
feat: Automatically refresh `aws_resources` materialized view

akash1810:
> [!TIP]
> Easiest to review [commit by commit](https://github.com/guardian/service-catalogue/pull/1048/commits).

The `aws_resources` materialized view is currently informing the tagging obligation, and is being used by a few dashboards. Materialized views need to be refreshed. We're currently doing this manually, and sporadically. This means our dashboards often show stale data.

This change adds a new lambda to refresh the materialized view. The lambda is triggered on the successful completion of an AWS based task. This should result in our dashboard reflecting up to date data at all times.

> [!NOTE]
> - The "obligatron" (#989) should replace this view. That is, this lambda is likely to be a short term solution.
> - This change creates a new database user.

## How has it been verified?
I've run the project locally, with success.

After [deploying to CODE](https://riffraff.gutools.co.uk/deployment/view/5002d964-1f79-40b4-8c98-37338741b050), I've manually invoked the lambda, and invoked it via manually running an Aws task. The logs for each can be seen in the screenshot below (logs at 12:05 originate from the manual invocation, logs at 12:08 originate from the event based invocation):

image

### DIFF
--- a/packages/cdk/lib/refresh-materialized-view.ts
+++ b/packages/cdk/lib/refresh-materialized-view.ts
@@ -1,0 +1,77 @@
+import type { GuStack } from '@guardian/cdk/lib/constructs/core';
+import type { GuSecurityGroup } from '@guardian/cdk/lib/constructs/ec2';
+import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
+import { Duration } from 'aws-cdk-lib';
+import type { IVpc } from 'aws-cdk-lib/aws-ec2';
+import { ScheduledFargateTask } from 'aws-cdk-lib/aws-ecs-patterns';
+import { Rule } from 'aws-cdk-lib/aws-events';
+import { LambdaFunction } from 'aws-cdk-lib/aws-events-targets';
+import { Architecture, Runtime } from 'aws-cdk-lib/aws-lambda';
+import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
+
+interface RefreshMaterializedViewLambdaProps {
+	vpc: IVpc;
+	db: DatabaseInstance;
+	dbAccess: GuSecurityGroup;
+}
+
+export function addRefreshMaterializedViewLambda(
+	scope: GuStack,
+	props: RefreshMaterializedViewLambdaProps,
+) {
+	const app = 'refresh-materialized-view';
+
+	const { vpc, dbAccess, db } = props;
+
+	const lambda = new GuLambdaFunction(scope, 'RefreshMaterializedView', {
+		app,
+		vpc,
+		architecture: Architecture.ARM_64,
+		securityGroups: [dbAccess],
+		fileName: `${app}.zip`,
+		handler: 'index.main',
+		environment: {
+			DATABASE_HOSTNAME: db.dbInstanceEndpointAddress,
+			QUERY_LOGGING: 'false', // Set this to 'true' to enable SQL query logging
+		},
+		runtime: Runtime.NODEJS_20_X,
+		timeout: Duration.minutes(10),
+	});
+
+	// This sort of lookup is a bit fragile!
+	// TODO pass the tasks in as a prop
+	const awsTasks = scope.node.children
+		.filter((_): _ is ScheduledFargateTask => _ instanceof ScheduledFargateTask)
+		.filter((_) => _.taskDefinition.family.includes('Aws'));
+
+	if (awsTasks.length === 0) {
+		// This is only seen at synth time. It doesn't impact running infrastructure.
+		throw new Error(`Could not find any 'Aws' tasks`);
+	}
+
+	awsTasks.forEach((task) => {
+		// Invoke the lambda when the task has completed successfully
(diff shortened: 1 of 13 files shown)


packages/cdk/lib/refresh-materialized-view.ts
```diff
@@ -0,0 +1,76 @@
...
+		},
+		runtime: Runtime.NODEJS_20_X,
+		timeout: Duration.minutes(10),
+	});
+
+	// This sort of lookup is a bit fragile!
+	// TODO pass the tasks in as a prop
+	const awsTasks = scope.node.children
+		.filter((_): _ is ScheduledFargateTask => _ instanceof ScheduledFargateTask)
+		.filter((_) => _.taskDefinition.family.includes('Aws'));
```
adamnfish, line 45: Should this say "starts with" instead of "includes"?

akash1810: [Sadly not](https://github.com/guardian/service-catalogue/blob/9a267b355c71c631920da61ae96ff399235b1972/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap#L432). The tags haven't been applied at this point either[^1], so we can't use those 😢.

[^1]: Presumably because they're [implemented as Aspects](https://docs.aws.amazon.com/cdk/v2/guide/tagging.html), and the Aspects are run later in the synth lifecycle.

packages/cdk/lib/refresh-materialized-view.ts
```diff
@@ -0,0 +1,77 @@
...
+					'containers.exitCode': [{ numeric: ['=', 0] }],
+					'containers.name': [
+						task.taskDefinition.defaultContainer?.containerName,
+					],
+				},
+			},
+		});
+	});
+
+	db.grantConnect(lambda, 'refresh_materialized_view');
```
akash1810, line 76: This database user is created below in the file `packages/common/prisma/migrations/20240530093317_aws_resources_owner/migration.sql`

packages/common/prisma/migrations/20240530093317_aws_resources_owner/migration.sql
akash1810, line 1: This user differs from [the others](https://github.com/guardian/service-catalogue/blob/main/sql/dbuser.sql) in that it's treated as any other Prisma migration. Consequently, it is applied by the robots, and not manually.

packages/refresh-materialized-view/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "refresh-materialized-view",
+  "version": "0.0.0",
+  "scripts": {
+    "test": "echo \"Error: no test specified\"",
```
akash1810, line 5: There's not much to unit test here. We _could_ integration test it...

packages/refresh-materialized-view/src/index.ts
```diff
@@ -0,0 +1,14 @@
+import { getPrismaClient } from 'common/database';
+import { getConfig } from './config';
+
+export async function main(...args: unknown[]) {
+	console.log(
+		`Called with ${args.map((arg) => JSON.stringify(arg)).join(', ')}`,
+	);
```
akash1810, line 7: Purely for debugging.

packages/common/prisma/migrations/20240530093317_aws_resources_owner/migration.sql
akash1810, line 1: This user differs from [the others](https://github.com/guardian/service-catalogue/blob/main/sql/dbuser.sql) in that it's treated as any other Prisma migration. Consequently, it is applied by the robots, and not manually.

packages/common/prisma/migrations/20240530093317_aws_resources_owner/migration.sql
```diff
@@ -0,0 +1,25 @@
...
+        -- This user is used by the refresh-materialized-view lambda
+        CREATE USER refresh_materialized_view WITH LOGIN;
+
+        GRANT USAGE ON SCHEMA public to refresh_materialized_view;
+        GRANT SELECT ON ALL TABLES IN SCHEMA public TO refresh_materialized_view;
+
+        -- The rds_iam role is created by the RDS IAM extension, which is not available in DEV
+        IF EXISTS (select * from pg_roles where rolname='rds_iam') THEN
+            GRANT rds_iam TO refresh_materialized_view;
+        END IF;
```
akash1810, line 13: This `IF` is only possible inside a `DO` statement[^1].

[^1]: https://stackoverflow.com/a/8099557 used as inspiration.

AshCorr (approved):

packages/cdk/lib/refresh-materialized-view.ts
akash1810: It's a little brittle, I agree. Worth noting it being a build time issue, with an exception thrown if `awsTasks` is empty, provides a bit more confidence that regressions would be found before deployment. However, being more type-safe, and explicit, is never a bad thing!

akash1810:
Confirming this is working as intended on PROD:

image

PR #1057, opened 2024-05-31
Create ingestion task for IAM credential reports

marsavar:
## What does this change?

Adds a new scheduled task to ingest IAM credential reports every 4 hours.

## Why?
This data will be used to populate a Grafana dashboard that will replace Security HQ.
Currently, Security HQ updates IAM credential reports every 4 hours, so we need service catalogue to match the ingestion frequency.

## How has it been verified?
Deployed to CODE and verified the new task appears in the list of tasks. I also ran it.

### DIFF
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -284,7 +284,6 @@ export function addCloudqueryEcsCluster(
 				tables: ['aws_iam_credential_reports'],
 			}),
 			policies: [listOrgsPolicy, cloudqueryAccess('*')],
-			runAsSingleton: true,
 			memoryLimitMiB: 1024,
 		},
 	];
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/cloudquery/index.ts
```diff
@@ -275,6 +275,18 @@ export function addCloudqueryEcsCluster(
...
+		{
+			name: 'AwsOrgWideIamCredentialReports',
+			description:
+				'Collecting IAM credential reports to surface information about outdated or inactive users and access keys',
+			schedule: nonProdSchedule ?? Schedule.rate(Duration.hours(4)),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_iam_credential_reports'],
+			}),
+			policies: [listOrgsPolicy, cloudqueryAccess('*')],
+			runAsSingleton: true,
```
akash1810, line 287: The `runAsSingleton` prevents overlapping runs, useful in the case where a task has an indeterminate duration. Does collecting the IAM credential report take more close to 4 hours?! 😱

marsavar: Good shout, have removed it. It definitely does not take that long.

akash1810 (approved):
The branch needs updating, pulling in the changes from https://github.com/guardian/service-catalogue/pull/1048. These changes should require a snapshot update too, with a new lambda trigger being created.

PR #1068, opened 2024-06-10
use url parsing instead of string matching

NovemberTang:
## What does this change?

We got a CodeQL alert for unsafe URL parsing. This change checks the relevant part of the URL for the correct string, making our check a little more accurate. It shouldn't have a significant impact on code security, because URLs are always coming from Snyk or Dependabot, but this is good practice, and makes our check more precise.

## How has it been verified?

Unit tests around URL ordering continue to pass

### DIFF
--- a/packages/repocop/src/evaluation/repository.ts
+++ b/packages/repocop/src/evaluation/repository.ts
@@ -415,7 +415,6 @@ const urlSortPredicate = (maybeUrl: string) => {
 		}
 		return 0;
 	} catch {
-		//Do nothing if the url is invalid
 		console.debug(`Invalid url: ${maybeUrl}`);
 		return 0;
 	}
--- a/packages/repocop/src/evaluation/repository.ts
+++ b/packages/repocop/src/evaluation/repository.ts
@@ -401,24 +401,22 @@ export function evaluateOneRepo(
 }
 
 //create a predicate that orders a list of urls by whether they contain snyk.io first, and then github.com second
-const urlSortPredicate = (url: string) => {
+const urlSortPredicate = (maybeUrl: string) => {
 	try {
-		const parsedUrl = new URL(url);
+		const url = new URL(maybeUrl);
 
-		if (
-			parsedUrl.hostname == 'snyk.io' ||
-			parsedUrl.hostname == 'security.snyk.io'
-		) {
+		if (url.hostname === 'snyk.io' || url.hostname === 'security.snyk.io') {
 			return -2;
 		} else if (
-			parsedUrl.hostname == 'github.com' &&
-			parsedUrl.pathname.includes('advisories')
+			url.hostname === 'github.com' &&
+			url.pathname.includes('advisories')
 		) {
 			return -1;
 		}
 		return 0;
 	} catch {
 		//Do nothing if the url is invalid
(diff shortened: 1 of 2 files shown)


packages/repocop/src/evaluation/repository.ts
```diff
@@ -401,12 +402,25 @@ export function evaluateOneRepo(
...
 const urlSortPredicate = (url: string) => {
-	if (url.includes('snyk.io')) {
-		return -2;
-	} else if (url.includes('github.com') && url.includes('advisories')) {
-		return -1;
+	try {
+		const parsedUrl = new URL(url);
+
+		if (
+			parsedUrl.hostname == 'snyk.io' ||
```
akash1810, line 409: Could we use [`===`](https://www.freecodecamp.org/news/loose-vs-strict-equality-in-javascript) here for strict equality? Surprised the linter didn't catch this.

packages/repocop/src/evaluation/repository.ts
```diff
@@ -401,12 +402,25 @@ export function evaluateOneRepo(
 
 //create a predicate that orders a list of urls by whether they contain snyk.io first, and then github.com second
 const urlSortPredicate = (url: string) => {
-	if (url.includes('snyk.io')) {
-		return -2;
-	} else if (url.includes('github.com') && url.includes('advisories')) {
-		return -1;
+	try {
+		const parsedUrl = new URL(url);
```
akash1810, line 406: minor: as we don't yet know if the incoming `url` is valid might an alternative name for it be `maybeUrl`? Then this `parsedUrl` variable can be called `url`.

packages/repocop/src/evaluation/repository.ts
```diff
@@ -401,12 +402,25 @@ export function evaluateOneRepo(
...
+			return -2;
+		} else if (
+			parsedUrl.hostname == 'github.com' &&
+			parsedUrl.pathname.includes('advisories')
+		) {
+			return -1;
+		}
+		return 0;
+	} catch {
+		//Do nothing if the url is invalid
```
akash1810, line 421: Would we want to log the invalid URL for debugging?

packages/repocop/src/evaluation/repository.ts
```diff
@@ -400,13 +401,23 @@ export function evaluateOneRepo(
...
-		return -2;
-	} else if (url.includes('github.com') && url.includes('advisories')) {
-		return -1;
+const urlSortPredicate = (maybeUrl: string) => {
+	try {
+		const url = new URL(maybeUrl);
+
+		if (url.hostname === 'snyk.io' || url.hostname === 'security.snyk.io') {
+			return -2;
+		} else if (
```
akash1810, line 410: minor: As we're using early returns, the `else` keyword isn't doing anything here, and could be dropped.

akash1810:
> Unit tests around URL ordering continue to pass

I'm not overly familiar with this area of the repository, but there doesn't appear to be a test just for `urlSortPredicate`? Might we want to add some to demonstrate it's behaviour in isolation with all other logic?

akash1810 (approved):
A couple of non-blocking comments. Would suggest adding a test for `urlSortPredicate` explicitly.

PR #1076, opened 2024-06-11
Consume Security Hub tagging data in Obligatron

AshCorr:
## What does this change?

Start using our Security Hub data in Obligatron.

## Why?

This greatly reduces the complexity of our Tagging obligation by reducing our dependency on our custom `aws_resources` view and instead leave it to Amazon to figure out whats tagged and what isn't.

## How has it been verified?

Ran Locally

image

### DIFF
--- a/packages/obligatron/src/obligations/index.ts
+++ b/packages/obligatron/src/obligations/index.ts
@@ -19,5 +19,6 @@ export type ObligationResult = {
 	/**
 	 * Key-value pairs to link failing obligations to the responsible teams.
 	 */
-	contacts?: Record<string, string>;
+
+	contacts?: object;
 };
--- a/packages/obligatron/src/obligations/tagging.test.ts
+++ b/packages/obligatron/src/obligations/tagging.test.ts
@@ -141,7 +141,7 @@ describe('The tagging obligation', () => {
 		]);
 
 		await expect(evaluateTaggingObligation(client)).rejects.toEqual(
-			new Error('Invalid resource in finding 123456789012 at index 0'),
+			new Error('Invalid resource in finding 123456789012'),
 		);
 	});
 });
--- a/packages/obligatron/src/obligations/tagging.test.ts
+++ b/packages/obligatron/src/obligations/tagging.test.ts
@@ -1,30 +1,80 @@
 import type { PrismaClient } from '@prisma/client';
-import type { AwsResource } from './tagging';
 import { evaluateTaggingObligation } from './tagging';
 
-const createPrismaClientWithMockedResponse = (response: AwsResource[]) => {
+const createPrismaClientWithMockedResponse = (response: unknown[]) => {
+	const aws_securityhub_findings = {
+		findMany: () => Promise.resolve(response),
+	};
+
 	const test = {
-		$queryRaw: () => Promise.resolve(response),
+		aws_securityhub_findings,
 	} as unknown as PrismaClient;
 
 	return test;
 };
 
 describe('The tagging obligation', () => {
-	it('passes correct resource', async () => {
+	it('catches failed securityhub findings', async () => {
 		const client = createPrismaClientWithMockedResponse([
 			{
-				account_id: '123456789012',
-				arn: 'arn:aws:s3:::mybucket',
-				service: 's3',
-				resource_type: 'bucket',
-				taggable: 'true',
-				tags: {
-					Stack: 'my-stack',
-					Stage: 'prod',
-					App: 'myapp',
-					'gu:repo': 'myrepo',
-				},
+				id: '123456789012',
+				title: 'failed tagging',
+				region: 'mars-north-1',
+				aws_account_id: '123456789012',
+				resources: [
+					{
+						Id: 'arn:aws:s3:::mybucket',
+						Tags: {
+							Stack: 'my-stack',
+							Stage: 'prod',
+							App: 'myapp',
+							'gu:repo': 'myrepo',
+						},
+					},
+				],
+			},
+			{
+				id: '123456789012',
+				title: 'failed tagging',
+				region: 'mars-north-1',
+				aws_account_id: '123456789012',
+				resources: [
+					{
+						Id: 'arn:aws:s3:::mybucket',
+						Tags: {
+							Stack: 'my-stack',
+							Stage: 'prod',
+							App: 'myapp',
+						},
+					},
+				],
+			},
+		]);
+
+		const results = await evaluateTaggingObligation(client);
+
+		expect(results).toHaveLength(2);
+		expect(results[0]).toEqual({
+			resource: 'arn:aws:s3:::mybucket',
+			reason: 'failed tagging',
+			contacts: {
+				aws_account_id: '123456789012',
+				Stack: 'my-stack',
+				Stage: 'prod',
+				App: 'myapp',
+			},
+			url: 'https://mars-north-1.console.aws.amazon.com/securityhub/home?region=mars-north-1#/findings?search=RecordState%3D%255Coperator%255C%253AEQUALS%255C%253AACTIVE%26Id%3D%255Coperator%255C%253AEQUALS%255C%253A123456789012',
+		});
+	});
+
+	it('handles findings with no resources', async () => {
+		const client = createPrismaClientWithMockedResponse([
+			{
+				id: '123456789012',
+				title: 'failed tagging',
+				region: 'mars-north-1',
+				aws_account_id: '123456789012',
+				resources: [],
 			},
 		]);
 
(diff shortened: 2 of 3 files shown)


packages/obligatron/src/obligations/tagging.test.ts
```diff
@@ -1,166 +1,136 @@
...
-		expect(results[0]).toEqual({
-			resource: 'arn:aws:s3:::mybucket',
-			reason: "Resource missing 'App' tag.",
-			contacts: { aws_account: '123456789012' },
-		});
+		expect(results).toHaveLength(0);
 	});
 
-	it('catches missing Repo tags', async () => {
+	it('handles findings with incorrect amount of resources', async () => {
```
akash1810, line 86: Is this "correctness" based on AWS's model of security hub findings? [The docs](https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_AwsSecurityFinding.html) suggest `resources` is an array with 1-32 items.

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -1,68 +1,103 @@
...
-	resource_type: string;
-	taggable: string;
-	tags?: Record<string, string> | null;
+type FindingResource = {
+	Id: string;
+	Tags: Record<string, string>;
 };
 
-const REQUIRED_TAGS = ['Stack', 'Stage', 'App', 'gu:repo'] as const;
+const securityHubLink = (region: string, findingId: string) => {
```
akash1810, line 9: https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_AwsSecurityFinding.html suggests there's a `SourceUrl` field. Could that be used?

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -1,68 +1,103 @@
...
-	for (const resource of awsResources) {
-		if (isExemptResource(resource)) {
+	for (const finding of findings) {
+		const resources = finding.resources?.valueOf();
+
+		if (!Array.isArray(resources)) {
+			console.error({
+				message: `Skipping invalid SecurityHub finding, invalid 'resources' field`,
+				finding_id: finding.id,
+			});
```
akash1810, line 67: minor: Would we want to know how many times we hit this code path per execution? If so, applying a `filter` might help?

```ts
const validFindings = findings.filter(f => Array.isArray(f));
const invalidFindings = findings.filter(f => !Array.isArray(f));

console.log(`Total invalid findings: ${invalidFindings.length}`);
invalidFindings.forEach(f => console.error({
  message: `Skipping invalid SecurityHub finding, invalid 'resources' field`,
  finding_id: f.id
}));
```

Might this also allow us to avoid the `.push` mutation of `results` below, assigning it via a `.map` over `validFindings`?

packages/obligatron/src/obligations/tagging.ts
```diff
@@ -1,68 +1,103 @@
...
+		}
+
+		for (const [index, resource] of resources.entries()) {
+			// This in theory should not happen as long as AWS don't change their schema.
+			// if they do change the schema its unlikely to be just this one finding failing,
+			// so lets make sure that we crash the lambda and get a humans attention!
+			if (!isFindingResource(resource)) {
+				throw new Error(
+					`Invalid resource in finding ${finding.id} at index ${index}`,
+				);
```
akash1810, line 88: IIUC this `Error` could halt execution at any point in the for loop. Is this desired? Is there a scenario where in a list of 10 items, only index 5 is faulty? Throwing in this way means 6-10 don't get processed?

I might have mis-understood the code too!

packages/obligatron/src/obligations/tagging.test.ts
```diff
@@ -1,166 +1,136 @@
 import type { PrismaClient } from '@prisma/client';
-import type { AwsResource } from './tagging';
 import { evaluateTaggingObligation } from './tagging';
 
-const createPrismaClientWithMockedResponse = (response: AwsResource[]) => {
+const createPrismaClientWithMockedResponse = (response: unknown[]) => {
```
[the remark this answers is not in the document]
AshCorr: Yea I couldn't get Prisma's guide to work 😅 Just ended up being far simpler to manually mock the client.

packages/obligatron/src/obligations/tagging.test.ts
```diff
@@ -1,166 +1,136 @@
...
-					App: 'myapp',
-					'gu:repo': 'myrepo',
-				},
+				id: '123456789012',
+				title: 'failed tagging',
+				region: 'mars-north-1',
+				aws_account_id: '123456789012',
+				resources: [
+					{
+						Id: 'arn:aws:s3:::mybucket',
```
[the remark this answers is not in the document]
AshCorr: I think in the real world we wouldn't expect more than 1 resource per finding. As Security hub needs to support a lot of use cases the schema is such that a finding can have multiple resources, but for our particular rules this shouldn't happen (i think).

I couldn't find a specification for `Id`, I think its sensible to assume it might not be unique but I don't think its worth dedupping atleast for our usecase.

packages/obligatron/src/obligations/tagging.test.ts
```diff
@@ -1,166 +1,136 @@
...
+						},
+					},
+				],
 			},
 		]);
 
 		const results = await evaluateTaggingObligation(client);
 
-		expect(results).toHaveLength(1);
+		expect(results).toHaveLength(2);
```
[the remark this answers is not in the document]
AshCorr: Laziness 😅 

I'll add it.

packages/obligatron/src/obligations/tagging.test.ts
```diff
@@ -1,166 +1,136 @@
...
+				region: 'mars-north-1',
+				aws_account_id: '123456789012',
+				resources: [
+					{
+						Id: 'arn:aws:s3:::mybucket',
+						Tags: {
+							Stack: 'my-stack',
+							Stage: 'prod',
+							App: 'myapp',
+							'gu:repo': 'myrepo',
```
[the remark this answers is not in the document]
AshCorr: Yes but would they be relevant in terms of contacting the resource owner?

packages/obligatron/src/obligations/tagging.test.ts
AshCorr: Correctness in terms of my interpretation of the data 😅 

Findings can have up to 32 items which as mentioned above gives Security Hub flexibility  in what rules are possible, however for the tagging rules I've found its a one to one relationship between finding and resources.

packages/obligatron/src/obligations/tagging.ts
AshCorr: Unfortunately it doesn't seem to be a required field and none of the tagging findings appear to have it set!

packages/obligatron/src/obligations/tagging.ts
AshCorr: Yep this is intentional, it will kill the entire lambda not just the 1 loop. Theres a comment just above explaining my reasoning!

TLDR: in theory this should only fail if AWS change their schema and in that case I want the lambda to fail fast and alert us.

packages/obligatron/src/obligations/index.ts
```diff
@@ -19,5 +19,6 @@ export type ObligationResult = {
 	/**
 	 * Key-value pairs to link failing obligations to the responsible teams.
 	 */
-	contacts?: Record<string, string>;
+
+	contacts?: object;
```
akash1810, line 23: What's the reason for this change? IIUC we're still storing something like `Record<string, unknown>`.

packages/obligatron/src/obligations/tagging.ts
akash1810: I think we might be able to fail faster by doing this check up front, either within the SQL query itself, or as a `filter` on the dataset before looping over each item.

These changes will only have a performance impact over a certain scale. I wonder if throwing outside of the loop improves readability, and maintainability?

Not a blocker though.

akash1810 (commented):
Added a couple of non-blocking comments. Let's add [the missing test](https://github.com/guardian/service-catalogue/pull/1076/files#r1637662341) before merging?

packages/obligatron/src/obligations/tagging.ts
AshCorr: This check is dual purpose, alerting us of invalid data is one of them, but mostly its there to make Typescript happy by asserting that the unknown(ish, its a JSONB) `resources` is infact an array! 

If this were moved up it would end up looking like this 

```
const validFindings = findings.filter(f => Array.isArray(f.resources));
```

and I don't think Typescript is smart enough to recognise that this is asserting that all `validFindings` have a `resources` property that is an Array

packages/obligatron/src/obligations/index.ts
AshCorr: `Record<string, string>` couldn't really encompass all the different JSON types that exist that could be used in this JSONB column. In particular undefined, null, and arrays.

Unfortunately `Record<string, unknown>` causes type errors as prisma can't cast unknown to a JSON primitive.

`Type 'Record<string, unknown>' is not assignable to type 'InputJsonValue | JsonNull'.`

`object` is a little broad as functions technically count as objects too, but it does encompass all of the different JSON primitives and Prisma finds it acceptable. 

In reality I think if we wanted to be more specific (as unfortunately Prisma doesn't export their JSON primitive types) we would make our own type

```
type JSONPrimitive =
	| undefined
	| null
	| string
	| number
	| boolean
	| JSONPrimitive[]
	| Record<String, JSONPrimitive>;

...
contacts: Record<string, JSONPrimitive>
```

but I wanted to avoid that complexity and stick with `object` which is almost what we want.

akash1810 (approved):

PR #1078, opened 2024-06-12
Run Obligatron daily at 7AM

AshCorr:
## What does this change?

Run Obligatron daily at 7AM

## Why?

So that we can start building a historical view and get an understanding of how much data there is (once we enable Security Hub on all accounts that is!)

## How has it been verified?

Ran snapshot test.


akash1810:
This is a cron method. Did we consider an event driven method, running the lambda once the task to collect the `aws_securityhub_findings` table completes?

akash1810 (approved):
[My comment](https://github.com/guardian/service-catalogue/pull/1078#issuecomment-2164546513) isn't blocking.

PR #1079, opened 2024-06-12
Track DRIFT in Obligatron

AshCorr:
## What does this change?

Collect Drift information in Obligatron.

## Why?

We probably won't do anything with this for a while, but it'd be nice to see if we can start using Obligatron to track some potentially "core" metrics.

Eventually this data could be used for potentially justifying 2 OKRs if we ever have the capacity:
 - Reduce number of stacks without automatic drift detection from X to 0. Where X is gotten from our current data.
 - **afterwards** Reduce number of stacks that have drifted from X to O.

I imagine most Stacks will fail as we don't automatically run drift detection anywhere yet (as far as I know!). 

## How has it been verified?

Ran locally.

### DIFF
--- a/packages/obligatron/src/index.ts
+++ b/packages/obligatron/src/index.ts
@@ -3,6 +3,7 @@ import { getPrismaClient } from '../../common/src/database';
 import { getConfig } from './config';
 import type { ObligationResult } from './obligations';
 import { Obligations, stringIsObligation } from './obligations';
+import { evaluateDriftObligation } from './obligations/drift';
 import { evaluateTaggingObligation } from './obligations/tagging';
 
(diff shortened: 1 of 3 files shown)


packages/obligatron/src/index.ts
```diff
@@ -3,6 +3,7 @@ import { getPrismaClient } from '../../common/src/database';
 import { getConfig } from './config';
 import type { ObligationResult } from './obligations';
 import { Obligations, stringIsObligation } from './obligations';
+import { evaluateDriftObligation } from './obligations/drift';
```
akash1810, line 6: Is this function already in VCS? Or are there missing files in this PR?

AshCorr: Doh, forgot to add the untracked file!

PR #1085, opened 2024-06-14
fix(repocop): Prisma migration for repocop db user

akash1810:
## What does this change?
Moves the manually applied SQL script to create the repocop user, and apply permissions, to a Prisma migration, so that it gets automatically applied.

Whilst it looks like a no-op, the exexution of:

```sql
GRANT SELECT ON ALL TABLES IN SCHEMA public TO repocop;
```

Will grant `repocop` access to the `view_repo_ownership` view, which was lost in #1066.

> [!NOTE]
> - The next PR will add a test that's run in CI to catch this regression in future.
> - We have [some more SQL scripts](https://github.com/guardian/service-catalogue/blob/524e0d07211ed1cbef86884878c3bb3d4eb2ab51/sql/dbuser.sql) that have been manually applied. In future PRs we will convert those to Prisma migrations too.

## Why?
The `repocop` user requires permission on `view_repo_ownership` to run. It no longer has these permissions.

## How has it been verified?
- [Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/bd655da3-70d0-43af-aa7e-548dfac20aba)
- Successfully executed the repocop lambda (see logs below)

image

---
Co-authored-by: @AshCorr.

### DIFF
--- a/packages/common/prisma/migrations/20240614105000_repocop_user/migration.sql
+++ b/packages/common/prisma/migrations/20240614105000_repocop_user/migration.sql
@@ -1,0 +1,25 @@
+DO
+$do$
+    BEGIN
+        -- Create the `repocop` user if it doesn't exist
+        IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE  rolname = 'repocop') THEN
+            CREATE USER repocop WITH LOGIN;
+        END IF;
+
+        -- Allow repocop to read all tables in the public schema
+        -- TODO should we limit permissions to exactly the tables/views that repocop uses, for POLP?
+        GRANT USAGE ON SCHEMA public to repocop;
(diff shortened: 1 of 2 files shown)


AshCorr (approved):

akash1810:
> * The next PR will add a test that's run in CI to catch this regression in future.

https://github.com/guardian/service-catalogue/pull/1086

akash1810:
> * We have [some more SQL scripts](https://github.com/guardian/service-catalogue/blob/524e0d07211ed1cbef86884878c3bb3d4eb2ab51/sql/dbuser.sql) that have been manually applied. In future PRs we will convert those to Prisma migrations too.

https://github.com/guardian/service-catalogue/pull/1087.

PR #1086, opened 2024-06-14
ci: Perform basic SQL tests

akash1810:
The changes in #1066 created a regression, where the `repocop` database user lost access to a view.

In this change, we run some SQL in CI in an attempt to catch such regressions earlier.

## How has it been verified?
Observe CI. The build log (see below) shows we've switched user, and performed a `SELECT *` on the view successfully (the query returns 0 rows as the database is empty).

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,6 +130,9 @@ jobs:
       - name: perform a DEV database migration
         run: npm -w cli start migrate -- --stage DEV
 
+      - name: basic database tests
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -130,6 +130,9 @@ jobs:
       - name: perform a DEV database migration
         run: npm -w cli start migrate -- --stage DEV
 
+      - name: basic database tests
+        run: psql -d postgresql://postgres:not_at_all_secret@localhost:5432/postgres -v ON_ERROR_STOP=1 -f sql/ci.sql
```
akash1810, line 134: Execute a file, rather than inlining the query, to make it easier to add more tests in future.

NovemberTang (approved):
🫶

PR #1087, opened 2024-06-14
feat: Prisma migration for remaining db users

akash1810:
Follows #1085, moving the manually applied SQL scripts to create the `dataaudit`, and `github_actions_usage` users, and apply permissions, to a Prisma migration, so that it gets automatically applied.

## How has it been verified?
See CI.


AshCorr (approved):
Nice tests 👌

PR #1100, opened 2024-06-18
bump code-quality deps

tjsilver:
## What does this change?

Bumps some code quality deps - replaces this [dependabot PR](https://github.com/guardian/service-catalogue/pull/1089) which throws a plugin error.


package-lock.json
AshCorr, line 1: Seems like theres significantly more lines changed here than the original PR? Could something have gone wrong with the upgrade?

akash1810: It looks like the changes are [mostly whitespace](https://github.com/guardian/service-catalogue/pull/1100/files?diff=split&w=1), which suggests the IDE isn't following the repository's prettier or editorconfig configuration?

PR #737, opened 2024-01-26
Create architecture diagram for Service Catalogue

NovemberTang:
## What does this change?

Creates architecture diagrams for the service catalogue using mermaids.

### Why diagrams as code?
Diagrams as code are auditable, easier to work on collaboratively, and easier to keep up to date.

### Why not inline the mermaids diagram, as supported by GH flavoured markdown?
Not all text editors support github flavoured markdown. Generating the images ourselves means we can validate that we are happy with them before committing them to version control

## How has it been verified?

Architecture diagram has been generated and added to the repo

## Next steps

Regenerate images on save??

PR #1108, opened 2024-06-20
Inline mermaid diagram

akash1810:
## What does this change?
Removes the `diagrams` workspace in favour of inlining the mermaid diagram. This means we remove a dependency, and a number of transitive dependencies.

https://github.com/guardian/service-catalogue/pull/737 makes a good point:

> Not all text editors support github flavoured markdown. Generating the images ourselves means we can validate that we are happy with them before committing them to version control

It looks like there are VS Code, and IntelliJ plugins for mermaidjs - maybe we should recommend them in the README?

## Why?
This should resolve https://github.com/guardian/service-catalogue/security/dependabot/30.

## How has it been verified?
Checked the README.md file on this branch, and confirmed it still renders the diagram:

image


NovemberTang (approved):
LGTM, thanks!

PR #1125, opened 2024-06-24
Implement the dependency graph integrator lambda

tjsilver:
## What does this change?

- Implements the existing lambda using the `gu-snyk-integrator` app for now (this will eventually be replaced with its own app).
- Tidies up the paramaters to the PR generator.
- Replaces the action versions in the generated `yaml` with SHA versions.  

## Why?

Because we want to replace the Snyk integrator lambda.

## How has it been verified?

Tested locally and PR contents looked sensible.

### DIFF
--- a/packages/dependency-graph-integrator/src/file-generator.test.ts
+++ b/packages/dependency-graph-integrator/src/file-generator.test.ts
@@ -14,8 +14,10 @@ jobs:
   dependency-graph:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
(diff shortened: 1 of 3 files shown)


marsavar (approved):
Looks good. Only a couple of questions: at the moment there's no error handling for the async functions, is that a design choice? I can see fewer than 10 try/catch blocks in the entire codebase, I suppose the intention is for errors to propagate to the top level?

NovemberTang:
> Looks good. Only a couple of questions: at the moment there's no error handling for the async functions, is that a design choice? I can see fewer than 10 try/catch blocks in the entire codebase, I suppose the intention is for errors to propagate to the top level?

Yeah that's correct, we want to be notified if there's an error, and the logic is simple enough that the unabridged message isn't too difficult to understand. So the easiest thing to do has just been to let AWS tell us when the lambda fails

PR #1131, opened 2024-06-26
fix(obligatron): Ensure logs and markers appear correctly in Central ELK

akash1810:
The logs produced by Lambdas are sent to CloudWatch Logs, and then processed by [guardian/cloudwatch-logs-management](https://github.com/guardian/cloudwatch-logs-management ) to get them into Central ELK. We're currently producing JSON formatted logs, which cloudwatch-logs-management [doesn't process correctly](https://github.com/guardian/cloudwatch-logs-management/issues/326).

Revert to the text log format so the message, and markers, are searchable in Central ELK.

## How has it been verified?
Obligatron produces log lines with markers for `total`, `withQueryLogging` and `startTime`. They can be seen using [devx-logs](https://github.com/guardian/devx-logs/tree/main/cli):

```sh
devx-logs --space devx \
  --app obligatron \
  --stage CODE \
  --column message \
  --column total \
  --column withQueryLogging \
  --column startTime
```

To test these changes:
1. Run Obligatron from `main` on CODE, and observe the logs; the "Starting Obligatron" log (and others) does not appear:
   image
2. Run Obligatron from this branch [on CODE](https://riffraff.gutools.co.uk/deployment/view/75b44215-aa6e-45ba-b022-66bd0444506b), and observe the logs; they all appear:
   image

### DIFF
--- a/packages/cdk/lib/obligatron.ts
+++ b/packages/cdk/lib/obligatron.ts
@@ -5,7 +5,7 @@ import { Duration } from 'aws-cdk-lib';
 import type { IVpc } from 'aws-cdk-lib/aws-ec2';
 import { Rule, RuleTargetInput, Schedule } from 'aws-cdk-lib/aws-events';
 import { LambdaFunction } from 'aws-cdk-lib/aws-events-targets';
-import { Architecture, Runtime } from 'aws-cdk-lib/aws-lambda';
+import { Architecture, LoggingFormat, Runtime } from 'aws-cdk-lib/aws-lambda';
 import type { DatabaseInstance } from 'aws-cdk-lib/aws-rds';
 import { Obligations } from '../../obligatron/src/obligations';
 
@@ -40,6 +40,17 @@ export class Obligatron {
 				toleratedErrorPercentage: 0,
 				snsTopicName: 'devx-alerts',
 			},
+
+			/*
+			Override the default provided by GuCDK for improved compatability with https://github.com/guardian/cloudwatch-logs-management when producing log lines with markers.
+
+			Note: `logFormat` is a deprecated property, replaced with `loggingFormat`.
+			However, we can only specify one of `logFormat` or `loggingFormat`, and as GuCDK is setting `logFormat`, we have to do the same.
+			TODO - patch GuCDK.
+
+			See also: https://github.com/guardian/cloudwatch-logs-management/issues/326.
+			 */
+			logFormat: LoggingFormat.TEXT,
 		});
 
 		for (const obligation of Obligations) {
--- a/packages/common/src/logs.ts
+++ b/packages/common/src/logs.ts
@@ -37,3 +37,36 @@ export function getCentralElkLink(props: CentralElkProps): string {
 
 	return `${base}?${queryString}`;
 }
+
+interface LogLine {
+	/**
+	 * The message to log.
+	 */
+	message: string;
+
+	/**
+	 * Any additional markers to log.
+	 */
+	[key: string]: unknown;
+}
+
+/**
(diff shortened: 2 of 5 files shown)


packages/cdk/lib/obligatron.ts
```diff
@@ -40,6 +40,17 @@ export class Obligatron {
...
+			/*
+			Override the default provided by GuCDK for improved compatability with https://github.com/guardian/cloudwatch-logs-management when producing log lines with markers.
+
+			Note: `logFormat` is a deprecated property, replaced with `loggingFormat`.
+			However, we can only specify one of `logFormat` or `loggingFormat`, and as GuCDK is setting `logFormat`, we have to do the same.
+			TODO - patch GuCDK.
+
+			See also: https://github.com/guardian/cloudwatch-logs-management/issues/326.
+			 */
+			logFormat: LoggingFormat.TEXT,
```
akash1810, line 53: See https://github.com/guardian/cdk/blob/7170a7149f427190c82deb7e4d4f76c5d23e3cb7/src/constructs/lambda/lambda.ts#L127.

packages/common/src/logs.ts
```diff
@@ -37,3 +37,36 @@ export function getCentralElkLink(props: CentralElkProps): string {
...
+	[key: string]: unknown;
+}
+
+/**
+ * Produces a log message with markers compatible with https://github.com/guardian/cloudwatch-logs-management.
+ * Note: if using within AWS Lambda, the Lambda must also log in text format not JSON.
+ *
+ * @see https://github.com/guardian/cloudwatch-logs-management/issues/326
+ */
+export const logger = {
```
akash1810, line 59: This abstraction is a little early, however it creates a greater distinction between `console`. That is, usage is more intentional.

packages/obligatron/src/index.ts
```diff
@@ -41,23 +42,22 @@ export async function main(obligation: string) {
 	const config = await getConfig();
 	const startTime = new Date();
 
-	console.log({
+	logger.log({
 		message: 'Starting Obligatron',
 		obligation,
-		stage: config.stage,
```
akash1810, line 47: The `stage` marker is provided by default from https://github.com/guardian/cloudwatch-logs-management. When we also provide it, it gets [moved to `overwrittenFields`](https://github.com/guardian/cloudwatch-logs-management/blob/fda75f06982140ae0f876361f57d5a4c76fbc3a9/README.md?plain=1#L25).

image

packages/obligatron/src/index.ts
```diff
@@ -41,23 +42,22 @@ export async function main(obligation: string) {
...
 		withQueryLogging: config.withQueryLogging,
 		startTime,
```
akash1810, line 49: I'm not sure this information needs to be pulled out into markers, as this is the only log line that adds them. An alternative might be to include them in the message.

packages/cdk/lib/obligatron.ts
AshCorr: Should we revert the change in CDK? From the sounds of things we don't support JSON logging in any lamda?

AshCorr (approved):

packages/cdk/lib/obligatron.ts
akash1810: I'm tempted to fix forward here as:
- There aren't too many lambdas [logging as JSON](https://metrics.gutools.co.uk/goto/_DywtpQSg?orgId=1) today
- Logs are lost only when an object is logged, `console.log("some string")` is ingested as usual

AshCorr: Ahh, didn't realize logs still worked if you don't log an object! Yea then it makes more sense to fix forward!

PR #1132, opened 2024-06-26
Dependency submisison action now echoes SBOM snapshot

NovemberTang:
## What does this change?

Rather than asking the use to click a link to check their dependencies look correct, we just echo them during the GHA run.

## Why?

 This improves developer experience, as there's less clicking around. It also fixes an issue where trying to view the contents of a dependency snapshot did not work in the browser for private repositories.

## How has it been verified?

Tested by raising [this PR](https://github.com/guardian/janus/pull/4211) against a private repo, and verified that we are able to see the contents of the snapshot in situ.

### DIFF
--- a/packages/dependency-graph-integrator/src/file-generator.ts
+++ b/packages/dependency-graph-integrator/src/file-generator.ts
@@ -28,7 +28,7 @@ export function createYaml(prBranch: string): string {
 					{
 						name: 'Log snapshot for user validation',
 						id: 'validate',
-						run: 'cat ${{ steps.submit.outputs.snapshot-json-path }}',
+						run: 'cat ${{ steps.submit.outputs.snapshot-json-path }} | jq',
 					},
 				],
 				permissions: { contents: 'write' },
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
Great work! 👍

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -17,12 +17,19 @@ export function createYaml(prBranch: string): string {
...
 					{
 						name: 'Submit dependencies',
+						id: 'submit',
 						uses: 'scalacenter/sbt-dependency-submission@7ebd561e5280336d3d5b445a59013810ff79325e # v3.0.1',
 					},
+					{
+						name: 'Log snapshot for user validation',
+						id: 'validate',
+						run: 'cat ${{ steps.submit.outputs.snapshot-json-path }} | jq',
+					},
```
akash1810, line 32: Might this become buried in the build log? An alternative might be to create a [job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/).

NovemberTang: The logs aren't excessively long, and this is only really for debugging, especially once the PR has been merged. I think a job summary might be overkill. Though I actually wasn't aware these existed until just now!

PR #1136, opened 2024-06-26
feat: enable dependabot alerts in dependency-graph-integrator

tjsilver:
## What does this change?

Implements the enabling of Dependabot alerts in the `dependency-graph-integrator` lambda.

## Why?

The dependency graph is required to be enabled for the submission of dependencies, while Dependabot alerts are required for any vulnerabilities in packages submitted to the dependency graph to be identified. Enabling one enables the other, so this will enable dependency graph and Dependabot alerts on any repo in which we raise a PR to implement the dependency submission workflow.


packages/dependency-graph-integrator/src/repo-functions.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const OWNER = 'guardian';
+
+export function generateBranchName(prefix: string) {
+	return `${prefix}-${randomBytes(8).toString('hex')}`;
+}
+
+/**
+ * Creates or updates a pull request, and return its URL.
+ * On error, an exception is thrown, or undefined is returned.
```
marsavar, line 29: Function docs shouldn't be concerned with implementation details. I would remove the second sentence.

packages/dependency-graph-integrator/src/repo-functions.ts
```diff
@@ -0,0 +1,156 @@
...
+	console.log(`Enabling Dependabot alerts for ${repo}`);
+	const enableResponse = await octokit.request(
+		'PUT /repos/{owner}/{repo}/vulnerability-alerts',
+		{
+			owner: OWNER,
+			repo,
+			headers: ghHeaders,
+		},
+	);
+	return enableResponse.status;
```
marsavar, line 155: I don't think it's clear that this function returns a status code, as you have to go and read the source code to find out. The return type doesn't give any clues either as it's just a number.

A couple of approaches that come to mind:

1. Aliasing `StatusCode` to number (`type StatusCode = number`), and changing the return type to `Promise<StatusCode>`
2. Returning an object, such as `{ statusCode: response.status }`.

packages/dependency-graph-integrator/src/repo-functions.ts
```diff
@@ -0,0 +1,156 @@
...
+		state: 'open',
+	} satisfies PullRequestParameters);
+
+	const found = pulls.filter((pull) => isGithubAuthor(pull, author));
+
+	if (found.length > 1) {
+		console.warn(`More than one PR found on ${repoName} - choosing the first.`);
+	}
+
+	return found[0];
```
marsavar, line 87: This will cause an error if `found` is empty. `return found[0]` should be hoisted up into the `if` block above so that it becomes infallible, and the case of an empty array should be handled as well.

packages/dependency-graph-integrator/src/repo-functions.ts
```diff
@@ -0,0 +1,156 @@
...
+
+function isGithubAuthor(pull: PullRequest, author: string) {
+	return pull.user?.login === author && pull.user.type === 'Bot';
+}
+
+export async function getExistingPullRequest(
+	octokit: Octokit,
+	repoName: string,
+	author: string,
+) {
```
marsavar, line 74: Can we add a return type?

marsavar (approved):
🚀

packages/dependency-graph-integrator/src/repo-functions.ts
tjsilver: Yeah, I know what you mean. Will take another look at how this could be structured.

packages/dependency-graph-integrator/src/repo-functions.ts
tjsilver: This function returns undefined if found[0] is empty IIUC - have clarified that by adding a type to the function.

NovemberTang (approved):

PR #1137, opened 2024-06-27
Use same name for aws_account_id contact in Obligatron

AshCorr:
## What does this change?

Fixes an inconsistency in the contacts column of our AMI Tagging results of so that it matches our other Tagging results.

## Why?

Our Security Results use `aws_account_id` whilst our AMI results use `aws_account`


akash1810 (approved):
Ah thanks! Worth adding an `interface` for contacts?

PR #1105, opened 2024-06-19
feat(obligatron): Augment tagging results with AMI info

akash1810:
## What does this change, and why?
The adoption of [AWS Security Hub Resource Tagging Standard](https://docs.aws.amazon.com/securityhub/latest/userguide/standards-tagging.html) means a subset of all AWS resources are having their tags evaluated.

Amazon Machine Images (AMIs) are one of the resources no covered. We produce a large number of AMIs. This change updates obligatron to query the `aws_ec2_images` table, to evaluate the tagging of our AMIs.

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/74848b28-ae9b-4f5c-bfd3-ef6fcb5c5b7a), and successfully ran the lambda with the results viewable [here](https://metrics.code.dev-gutools.co.uk/goto/cXu5XjUSg?orgId=1).

PR #1146, opened 2024-07-01
Create an obligation lambda for production repositories with out-of-SLA dependency vulnerabilities

NovemberTang:
## What does this change?

Creates a new running mode for the obligation lambda, called `PRODUCTION_DEPENDENCIES`. This summarises whether or not a repository contains vulnerable dependencies that exceed our SLAs.

Eventually, it should just link directly to the security or insights tab on the given repo, but while we still have vulnerabilities being sourced from Snyk, let's link to grafana instead to provide visibility of the vulnerabilities regardless of source.

Runs obligation lambdas on a staggered schedule to avoid overloading the DB (to avoid problems when we have like 10 of these things executing at once)

## Why?

So we are able to track our SLA progress over time

## How has it been verified?

Added unit tests, local execution works.

## Next steps

Change the way the CDK works to guarantee that repocop (which generates the `repocop_vulnerabilities` table) executes just before the lambda.

Reuse some of this logic to create a lambda that evaluates AWS FSBP vulnerabilities

### DIFF
--- a/packages/cdk/lib/obligatron.ts
+++ b/packages/cdk/lib/obligatron.ts
@@ -53,13 +53,19 @@ export class Obligatron {
 			logFormat: LoggingFormat.TEXT,
 		});
 
-		for (const obligation of Obligations) {
-			new Rule(stack, `obligatron-${obligation}`, {
-				description: `Daily execution of Obligatron lambda for '${obligation}' obligation`,
-				schedule: Schedule.cron({ minute: '0', hour: '7' }),
+		const ObligationsWithIndex = Obligations.map((obligation) => ({
+			obligation,
+			index: Obligations.indexOf(obligation),
+		}));
+
+		for (const obligation of ObligationsWithIndex) {
+			const startTime = (9 + obligation.index).toString();
+			new Rule(stack, `obligatron-${obligation.obligation}`, {
+				description: `Daily execution of Obligatron lambda for '${obligation.obligation}' obligation`,
+				schedule: Schedule.cron({ minute: '0', hour: startTime }),
 				targets: [
 					new LambdaFunction(lambda, {
-						event: RuleTargetInput.fromText(obligation),
+						event: RuleTargetInput.fromText(obligation.obligation),
 					}),
 				],
 			});
--- a/packages/cdk/lib/obligatron.ts
+++ b/packages/cdk/lib/obligatron.ts
@@ -53,23 +53,18 @@ export class Obligatron {
 			logFormat: LoggingFormat.TEXT,
 		});
 
-		const ObligationsWithIndex = Obligations.map((obligation) => ({
-			obligation,
-			index: Obligations.indexOf(obligation),
-		}));
-
-		for (const obligation of ObligationsWithIndex) {
-			const startTime = (9 + obligation.index).toString();
-			new Rule(stack, `obligatron-${obligation.obligation}`, {
-				description: `Daily execution of Obligatron lambda for '${obligation.obligation}' obligation`,
+		Obligations.forEach((obligation, index) => {
+			const startTime = (9 + index).toString();
+			new Rule(stack, `obligatron-${obligation}`, {
+				description: `Daily execution of Obligatron lambda for '${obligation}' obligation`,
 				schedule: Schedule.cron({ minute: '0', hour: startTime }),
 				targets: [
 					new LambdaFunction(lambda, {
-						event: RuleTargetInput.fromText(obligation.obligation),
+						event: RuleTargetInput.fromText(obligation),
 					}),
 				],
 			});
-		}
+		});
 
 		db.grantConnect(lambda, 'obligatron');
 	}
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -223,7 +223,7 @@ export class ServiceCatalogue extends GuStack {
 
 		const prodSchedule = Schedule.cron({
 			weekDay: 'MON-FRI',
-			hour: '10',
+			hour: '9',
 			minute: '30',
 		});
 
--- a/packages/common/src/functions.ts
+++ b/packages/common/src/functions.ts
@@ -3,12 +3,14 @@ import type { Action } from '@guardian/anghammarad';
 import { createAppAuth } from '@octokit/auth-app';
 import type { SNSEvent } from 'aws-lambda';
 import { Octokit } from 'octokit';
-import type {
-	GitHubAppConfig,
-	GithubAppSecret,
-	NonEmptyArray,
-	Severity,
-} from 'common/types';
+import {
+	type GitHubAppConfig,
+	type GithubAppSecret,
+	type NonEmptyArray,
+	type RepocopVulnerability,
+	type Severity,
+	SLAs,
+} from 'common/src/types';
 
 export async function getGithubClient(githubAppConfig: GitHubAppConfig) {
 	const auth = createAppAuth(githubAppConfig.strategyOptions);
@@ -171,6 +173,21 @@ export function stringToSeverity(severity: string): Severity {
 	} else {
 		return 'unknown';
 	}
+}
+
+export function daysLeftToFix(vuln: RepocopVulnerability): number | undefined {
+	const daysToFix = SLAs[vuln.severity];
+	if (!daysToFix) {
+		return undefined;
+	}
+	const fixDate = new Date(vuln.alert_issue_date);
+	fixDate.setDate(fixDate.getDate() + daysToFix);
+	const millisecondsInADay = 1000 * 60 * 60 * 24;
+	const daysLeftToFix = Math.ceil(
+		(fixDate.getTime() - new Date().getTime()) / millisecondsInADay,
+	);
+
+	return daysLeftToFix < 0 ? 0 : daysLeftToFix;
 }
 
 export function toNonEmptyArray<T>(value: T[]): NonEmptyArray<T> {
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -1,5 +1,8 @@
 import { type StrategyOptions } from '@octokit/auth-app';
-import type { repocop_vulnerabilities } from '@prisma/client';
+import type {
+	github_repositories,
+	repocop_vulnerabilities,
+} from '@prisma/client';
 
 export type GithubAppSecret = {
 	appId: string;
@@ -61,6 +64,26 @@ export type RepocopVulnerability = Omit<
 > & {
 	severity: Severity;
 };
+
+type RepositoryFields = Pick<
+	github_repositories,
+	| 'archived'
+	| 'name'
+	| 'full_name'
+	| 'topics'
+	| 'updated_at'
+	| 'pushed_at'
+	| 'created_at'
+	| 'id'
+	| 'default_branch'
+>;
+
+export interface Repository extends RepositoryFields {
+	archived: NonNullable<RepositoryFields['archived']>;
+	name: NonNullable<RepositoryFields['name']>;
+	full_name: NonNullable<RepositoryFields['full_name']>;
+	id: NonNullable<RepositoryFields['id']>;
+}
 
 // The number of days teams have to fix vulnerabilities of a given severity
 export const SLAs: Record<Severity, number | undefined> = {
--- a/packages/obligatron/src/obligations/dependency-vulnerabilities.ts
+++ b/packages/obligatron/src/obligations/dependency-vulnerabilities.ts
@@ -10,9 +10,13 @@ import type { ObligationResult } from '.';
 
 // type UserEvent = Event & {UserId: string}
 
+type ObligatronRepocopVulnerability = RepocopVulnerability & {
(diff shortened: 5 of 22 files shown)


packages/common/src/functions.ts
```diff
@@ -173,6 +175,21 @@ export function stringToSeverity(severity: string): Severity {
 	}
 }
 
+export function daysLeftToFix(vuln: RepocopVulnerability): number | undefined {
```
NovemberTang, line 178: this has been moved from vuln-digest. No new code.

packages/common/src/types.ts
```diff
@@ -62,6 +65,26 @@ export type RepocopVulnerability = Omit<
 	severity: Severity;
 };
 
+type RepositoryFields = Pick<
```
NovemberTang, line 68: this has moved from vuln-digest/types

packages/repocop/src/remediation/vuln-digest/vuln-digest.ts
```diff
@@ -24,27 +25,13 @@ function getOwningRepos(
...
 
 export function getTopVulns(vulnerabilities: RepocopVulnerability[]) {
+	//TODO delete this. It's no longer being used
 	return vulnerabilities
 		.sort(vulnSortPredicate)
 		.slice(0, 10)
 		.sort((v1, v2) => v1.full_name.localeCompare(v2.full_name));
 }
 
-export function daysLeftToFix(vuln: RepocopVulnerability): number | undefined {
```
NovemberTang, line 33: moved to `common`

packages/cdk/lib/obligatron.ts
```diff
@@ -53,13 +53,19 @@ export class Obligatron {
...
-		for (const obligation of Obligations) {
-			new Rule(stack, `obligatron-${obligation}`, {
-				description: `Daily execution of Obligatron lambda for '${obligation}' obligation`,
-				schedule: Schedule.cron({ minute: '0', hour: '7' }),
+		const ObligationsWithIndex = Obligations.map((obligation) => ({
+			obligation,
+			index: Obligations.indexOf(obligation),
+		}));
+
+		for (const obligation of ObligationsWithIndex) {
```
marsavar, line 61: I think this can simply be 
```typescript
Obligations.forEach((obligation, index) => { ... })
```
This  avoids allocating an intermediate ObligationsWithIndex array.

packages/cdk/lib/obligatron.ts
```diff
@@ -53,13 +53,19 @@ export class Obligatron {
...
+			const startTime = (9 + obligation.index).toString();
```
NovemberTang, line 62: Adding an offset to stagger the start times of the lambdas

marsavar, line 62: What's the behaviour once we have a sufficient number of obligations such as this addition operation results in > 24? Does it wrap around or does it error?

Taking a step back here, what's the concern with running all the lambdas at the same time?

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
```diff
@@ -0,0 +1,97 @@
...
+			topics: {
+				has: 'production',
+			},
+		},
+	});
+
+	return toNonEmptyArray(repositories.map((r) => r as Repository));
+}
+
+//TODO test me
```
tjsilver, line 50: Does this comment need to be removed now?

tjsilver (approved):
Looks good! 👍

packages/cdk/lib/obligatron.ts
AshCorr: From the PR description

> Runs obligation lambdas on a staggered schedule to avoid overloading the DB

If we think this is a concern maybe we could use setup lambda concurrency for this lambda instead of dealing with a bunch of different schedules? https://docs.aws.amazon.com/lambda/latest/dg/lambda-concurrency.html

Worth noting that Grafana & Cloudquery probably open far more connections and query more tables than Obligatron!

packages/cdk/lib/service-catalogue.ts
```diff
@@ -223,7 +223,7 @@ export class ServiceCatalogue extends GuStack {
 
 		const prodSchedule = Schedule.cron({
 			weekDay: 'MON-FRI',
-			hour: '10',
```
AshCorr, line 226: FYI we picked 7AM pretty arbitrarily for Obligatron. Maybe it might make more sense to run it EOD if its job is to take a snapshot of the day?

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
```diff
@@ -0,0 +1,97 @@
...
+	client: PrismaClient,
+): Promise<NonEmptyArray<ObligatronRepocopVulnerability>> {
+	const rawResponse = await client.repocop_vulnerabilities.findMany({});
+
+	return toNonEmptyArray(rawResponse.map(prismaToCustomType));
+}
+
+async function getProductionRepos(client: PrismaClient): Promise<Repository[]> {
+	console.debug('Discovering repositories');
+	const repositories = await client.github_repositories.findMany({
```
akash1810, line 38: In https://github.com/guardian/service-catalogue/pull/1106 we version controlled the obligatron database user, including the tables it has access to. I think we'd want a new migration, adding permissions to the `repocop_vulnerabilities` and `github_repositories` tables.

An example can be seen in https://github.com/guardian/service-catalogue/pull/1105.

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
```diff
@@ -0,0 +1,97 @@
...
+	repo: Repository,
+): ObligationResult | undefined {
+	const repoVulns = vulns.filter((v) => v.full_name === repo.full_name);
+	const repoOwners = [...new Set(repoVulns.flatMap((v) => v.repo_owner))];
+
+	const urlParams = repoOwners
+		.map((owner) => `var-repo_owner=${owner}`)
+		.join('&');
+
+	if (repoVulns.length > 0) {
```
akash1810, line 61: minor: this could be simplified a little by inverting:

```ts
if (repoVulns.length === 0) {
  logger.log(`Repo ${repo} has no vulns, yay!`);
  return undefined;
}

const vulnNames = ...
return {
  resource: repo.full_name,
  ...etc
}
```

packages/cdk/lib/service-catalogue.ts
NovemberTang: I think there's probably a bigger issue here, which is that the service catalogue contains a lot of jobs that run at seemingly random times of the day. I wonder if it might make sense at some point to create a big step function, where we run all the CQ jobs first, then repocop, then finally all the obligatron lambdas?

AshCorr: Might be a better UX experience aswell as:
 - The dashboard won't update throughout the day when people might be looking at it
 - It makes it clear that the results are not real time as you will always see yesterdays results when you look at the dashboard

akash1810: Might we want to move to an event driven trigger now? An example of this can be seen in https://github.com/guardian/service-catalogue/blob/f3c5244757b14c8145857bf8795e584f38df3b20/packages/cdk/lib/github-actions-usage.ts#L41-L51.

AshCorr: > I think there's probably a bigger issue here, which is that the service catalogue contains a lot of jobs that run at seemingly random times of the day. I wonder if it might make sense at some point to create a big step function, where we run all the CQ jobs first, then repocop, then finally all the obligatron lambdas? Otherwise you could end up with obligatron being nearly 48 hours behind the live picture

Definitely agree, we did want to make Obligatron event driven but compromised with a schedule for the time being!

Maybe a short term solution might be to dedicate parts of the day to each step in the process, so for example all Cloudquery jobs should run in the morning, Repocop should run midday, and Obligatron should run end of day!

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
NovemberTang: 1. is there a tangible risk to granting read on all tables to obligatron? Is that risk worth needing a new migration every time we add an obligatron lambda, given those migrations are a bit of a pain to create/apply?
2. Do we need to create a migration like this for repocop?
3. Question 1 but replace obligatron with repocop.

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
```diff
@@ -0,0 +1,96 @@
...
+	const repos = await getProductionRepos(client);
+	const vulns = (await getRepocopVulnerabilities(client)).filter(
+		(v) => daysLeftToFix(v) === 0,
+	);
+
+	const resultsOrUndefined: Array<ObligationResult | undefined> = repos.map(
+		(repo) => {
+			return evaluateObligationForOneRepo(vulns, repo);
+		},
+	);
```
akash1810, line 91: minor: IIUC this logic exists to join two datasets within the application code? Did we consider performing this join at the query level? Prisma likely doesn't know about the relation, however we can execute raw SQL ([example](https://github.com/guardian/service-catalogue/blob/f3c5244757b14c8145857bf8795e584f38df3b20/packages/github-actions-usage/src/db-read.ts#L35-L42)). I wonder if this reduces the complexity of the application code too?

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
akash1810: > is there a tangible risk to granting read on all tables to obligatron?

Possibly not, but as this will happen via a migration, I'd suggest following POLP here has the same complexity.

> given those migrations are a bit of a pain to create/apply?

Could you expand on this? I've found the automation works pretty reliably now. Very occasionally there's a bit of a race to use CODE.

> Do we need to create a migration like this for repocop?

This has been done in https://github.com/guardian/service-catalogue/pull/1085. Indeed, all IAM authenticated database users that we've previously created by hand now have a Prisma migration. IIUC the Grafana user is the only one that is not managed by Prisma, as it has a text password, which cannot go into source control.

NovemberTang: **2 continued** (sorry in advance for the wall of text)

I wouldn't have been able to guess that setting up a new obligatron lambda would require creating a new migration until after it started failing, if it hadn't been pointed out to me ahead of time (luckily) before merging, especially as repocop follows a different pattern. It's a hidden, but required task when creating a new obligation. If this had been merged a bit earlier, it would have caused a failure that would require time to debug. In order to create the migration, I'd need to:

- remember/be reminded that a migration needs to be applied
- context-switch into SQL
- remind myself of how prisma migrations work
- dig back through the typescript to work out what tables I need (maybe double checking that obligatron doesn't already have access to one of them via another migration?)
- leaf through previous migrations to copy over an example to modify (potentially introducing copy/paste errors)
- run a migration in DEV to make sure it worked.

I could easily see that adding a couple of hours development time, especially if the user makes a mistake, for not very much benefit, especially as the risk of tampering is basically zero

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
NovemberTang: Generally I'm quite averse to doing anything more complicated than a `SELECT ... WHERE ...` using the prisma client, as it's easy to introduce untestable mistakes that are easy to miss. For me, the main advantage of Prisma is being able to perform these joins/transformations in a unit testable way. Having `evaluateObligationForOneRepo` does slightly increase the lines of code over raw SQL, but imo, is more maintainable long term because we can precisely verify its behaviour easily via unit tests rather than eyeballing the results.

packages/cdk/lib/service-catalogue.ts
NovemberTang: Makes sense, I think we could leave this for another PR though!

packages/obligatron/src/obligations/dependency-vulnerabilities.ts
akash1810: Are there parallels to building IAM policies in the process you describe? In that it's easy to forget the permissions used in DEV is different from those in CODE/PROD, understanding what the policy needs requires digging through the application code, understanding the IAM policy statement requires reading through the [AWS docs](https://awspolicygen.s3.amazonaws.com/policygen.html).

That's not to say the current DX is great! However, I'd suggest we invest a bit of time to improve the DX before applying `*` permissions.

Possible options might include:
- Structuring the project to make database interactions explicit. The [`github-actions-merge` package](https://github.com/guardian/service-catalogue/tree/main/packages/github-actions-usage/src) attempts this, with database interactions in files prefixed `db-`.
- Running integration tests for obligatron, repocop, etc. in CI.
- Add some [event based logging](https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/logging#event-based-logging), logging out the tables being queried.

NovemberTang: As an update: PR to allow obligatron and repocop to read all now and future tables is here: #1154

PR #1152, opened 2024-07-02
only create refresh-materialized-view user if it doesn't already exist

NovemberTang:
## What does this change?

Adds the equivalent of an IF NOT EXISTS to the `refresh-materialized-view` user.

## Why?

To stop DEV migrations from failing.

## How has it been verified?

Deployed to CODE, which detected no new migrations. If we want to be good citizens, after merging, we can go in and manually adjust the checksum for this migration, but this is not required.


akash1810 (approved):

akash1810:
> we can go in and manually adjust the checksum for this migration

See also https://github.com/prisma/prisma/issues/12666. I think it's odd for Prisma to not perform checksum comparisons each time TBH!

PR #1154, opened 2024-07-03
Allow repocop and obligatron to read all tables

NovemberTang:
## What does this change?

Rather than creating a new migration every time obligatron or repocop needs access to a table, let's grant read only access to all tables in the DB

## Why?

Creating a migration is a manual, infrequent task, which means it takes a while, and is prone to mistakes. Allowing repocop and obligatron to read all tables in the DB up front eliminates this task, in exchange for minimal additional risk. No additional write permissions were granted

## How has it been verified?

Verified locally by creating the DB, running the migration, and running the SQL "unit tests" in `sql/ci.sql`. CI should also verify this.

Once approved, I will manually deploy to CODE to apply the migration, then merge.


packages/common/prisma/migrations/20240703133028_obligatron_full_read/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Allow repocop and obligatron read access to all current and future table.
+ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO repocop;
+
+GRANT SELECT ON ALL TABLES IN SCHEMA public TO obligatron;
```
akash1810, line 4: Why is this `GRANT SELECT` needed for the `obligatron` user, but not the `repocop` user?

akash1810:
> Once approved, I will manually deploy to CODE to apply the migration, then merge.

Why wouldn't we use the automation here?

NovemberTang:
> > Once approved, I will manually deploy to CODE to apply the migration, then merge.
> 
> Why wouldn't we use the automation here?

Sorry I meant manually deploy, as we don't have CD on CODE!

packages/common/prisma/migrations/20240703133028_obligatron_full_read/migration.sql
NovemberTang: Repocop already had GRANT SELECT applied during the `repocop_user` migration. And the obligatron change wouldn't work without it, so I guess it's necessary!

packages/common/prisma/migrations/20240703133028_obligatron_full_read/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Allow repocop and obligatron read access to all current and future table.
```
tjsilver, line 1: nit: "current and future tables."

akash1810 (approved):

tjsilver (approved):
Tested locally too and behaved as described.

PR #1156, opened 2024-07-03
parallelise ci

NovemberTang:
## What does this change?

Creates a composite action for running an arbitrary npm script. Then, uses that action to run `lint`, `typecheck`, etc in parallel, before moving on to the final build step.

So that we can still run these steps locally, I've created a `local-ci.sh` to run those steps.

## Why?

The CI script runs pretty fast on our machines with a bajillion cores. GH runners only have two, meaning CI times on GHA have crept up over the last 2 years from about 1:30 to over 5 minutes. This change brings that time back down to 2:49-ish.

## How has it been verified?

CI passes. `local-ci.sh` works.

### DIFF
--- a/.github/actions/npm-run/action.yml
+++ b/.github/actions/npm-run/action.yml
@@ -1,0 +1,32 @@
+name: Run an npm script
+description: 'Run an npm script'
+inputs:
+  script:
+    description: 'The npm script to run'
+    required: true
+  deployment_key:
+    description: 'The private key to use for deployment'
+    required: true
+
+
+runs:
+  using: 'composite'
+  steps:
+
+    - name: Setup Node
+      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+      with:
+        node-version-file: '.nvmrc'
+        cache: 'npm'
+
+    - uses: guardian/actions-read-private-repos@8792b5279dc2e6dfb6b9aa6ba2f26b6226be444c # v0.1.1
+      with:
+        private-ssh-keys: ${{ inputs.deployment_key }}
+
+    - name: install dependencies
+      shell: bash
+      run: npm ci
+
+    - name: Run script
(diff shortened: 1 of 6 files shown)


.github/actions/npm-run/action.yml
```diff
@@ -0,0 +1,32 @@
+name: Run an npm script
+description: 'Run an npm script'
+inputs:
+  script:
+    description: 'The npm script to run'
+    required: true
+  deployment_key:
```
NovemberTang, line 7: Using the SHA for our action rather than the version number means passing the key in is safer, as the implementation of the action can't change under our feet, and suddenly start sending its value somewhere. We're also using only trusted actions here. It's not possible to pass values into composite actions as secrets, and it's also not possible to run a shared workflow as a step, rather than a job, so a composite action is needed here. I'm open to suggestions to make this a bit safer if people have any!

scripts/local-ci.sh
NovemberTang, line 1: I'm not married to the name of this file tbh

tjsilver (approved):
Great work!

.github/workflows/ci.yml
```diff
@@ -20,9 +20,58 @@ concurrency:
...
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/npm-run
+        with:
+          script: typecheck
+          deployment_key: ${{ secrets.PRIVATE_INFRASTRUCTURE_CONFIG_DEPLOY_KEY }}
+
+  synth:
```
akash1810, line 59: Do we need this job? You've added a `build` script to `cdk/package.json`, which is an alias to `synth`, I think for consistency? IIUC we're running `cdk synth` twice now: once as a job here, and once via the `build.sh` script.

PR #1157, opened 2024-07-03
[1/x] Report FSBP findings

marsavar:
## What does this change?
First stab at Cloudbuster.

At the moment, the code queries the database and gets the Security Hub findings.
It then groups up findings for each account and creates a digest per team.

The digest isn't currently emailed to teams, but this will be done with Anghammarad in a subsequent PR.
For now, I just wanted to sense check that this structure makes sense.

## How has it been verified?

You will need to add `aws_securityhub_findings` to `docker-compose.yaml` to get the data in your local db.
Then:

1. Grab Janus credentials.
2. Run `npm run cloudbuster:start` from the root

And observe the digests being printed to stdout.


package.json
```diff
@@ -9,7 +9,9 @@
     "pretypecheck": "npm -w common run typecheck",
     "typecheck": "npm run typecheck --workspaces",
     "build": "npm run build --workspaces --if-present",
-    "lint": "eslint packages/** --ext .ts,.tsx --no-error-on-unmatched-pattern"
+    "lint": "eslint packages/** --ext .ts,.tsx --no-error-on-unmatched-pattern",
+    "cloudbuster::start": "npm run start --workspace cloudbuster",
```
NovemberTang, line 13: why are we introducing this kind of notation?

packages/cloudbuster/src/findings.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	it('Returns false if a critical finding is outside the window', () => {
+		const firstObservedAt = new Date(MOCK_ONE_WEEK_AGO);
+		const severity = 'CRITICAL';
+
+		const isWithinSla = isWithinSlaTime(firstObservedAt, severity);
+		expect(isWithinSla).toBe(false);
+	});
+
+	it('Returns false if a low finding was first observed one day ago', () => {
```
NovemberTang, line 49: Would it make sense for this to always return true for a low severity finding? As we don't have an expectation for when these should be fixed.

packages/cloudbuster/src/findings.ts
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return {
+		accountId: awsAccountId,
+		subject: `Security Hub vulnerabilities detected in AWS account ${teamFindings[0]?.awsAccountName}`,
+		message: createEmailBody(teamFindings),
+	};
+}
+
+export function groupFindingsByTeam(
```
NovemberTang, line 118: should we be testing this function? especially as it's doing some mutating, and isn't simple to read

package.json
marsavar: It's fairly standard to add scripts to the root `package.json` to be able to run sub-projects without having to `cd` into the projects, or without having to manually pass the `--workspace` flag every time.

Though after doing some googling, a single colon is more conventional as a separator (see the naming convention section in the [eslint docs](https://eslint.org/docs/latest/contribute/package-json-conventions#names)), so I'll change it to that!

packages/cloudbuster/src/findings.test.ts
marsavar: I ported the logic from the SQL that powers the [FSBP compliance dashboard](https://metrics.gutools.co.uk/d/ddi3x35x70jy8d/fsbp-compliance):

```sql
    when severity='CRITICAL' AND first_observed_at >= (CURRENT_DATE - 2) then 'true'
    when severity='HIGH' AND first_observed_at >= (CURRENT_DATE - 30) then 'true'
    else 'false'
```
So if we change it here it should also be adjusted in the dashboard. One for discussion!

packages/cloudbuster/src/findings.ts
marsavar: This function doesn't mutate the input, `reduce` creates a new object (and mutates the accumulator by adding new key/values). 

Though I agree that `reduce` functions are a bit hard to read. I'll refactor to make it more readable, and add tests.

packages/cloudbuster/src/findings.test.ts
NovemberTang: I think a slight complication here is that the dashboard only contains critical and high severity issues, so there isn't really any handling of medium/low/informational vulnerabilities. Introducing handling for severities lower than `high`, is [new logic](https://github.com/guardian/service-catalogue/blob/cc1b6c1a103018dcacdf2f7219af1e346da32f5f/packages/common/src/types.ts#L89), although it's not immediately obvious

NovemberTang: Discussed IRL, we'll standardise SLAs in a follow up PR

NovemberTang (approved):

PR #1164, opened 2024-07-04
Enable Dependency Graph Integrator to raise PRs to add workflow

tjsilver:
## What does this change?

Enables repocop to identify production Scala repos that don't have the SBT dependency submission workflow already, and pick one repo at random to call Dependency Graph Integrator which will raise a PR on the repo to include the workflow, and enable Dependabot alerts/Dependency Graph if not enabled.

## Why?

This is part of our migration away from Snyk over the next year. 

## How has it been verified?

Tested locally with CODE data. 

## Next steps
- Prioritise this for repos that don't have the Snyk workflow - this could be handled manually as it's only a handful.
- Enable the integrator for Gradle/Kotlin repos, and any other of our languages for which there is a dependency submission workflow available.
- Increase the number of PRs raised per run...?
- Update our recommendations

### DIFF
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -29,7 +29,7 @@ import {
 	getTeams,
 } from './query';
 import { protectBranches } from './remediation/branch-protector/branch-protection';
-import { sendOneRepo } from './remediation/dependency_graph-integrator/send-to-sns';
+import { sendOneRepoToDepGraphIntegrator } from './remediation/dependency_graph-integrator/send-to-sns';
 import { sendUnprotectedRepo } from './remediation/snyk-integrator/send-to-sns';
 import { sendPotentialInteractives } from './remediation/topics/topic-monitor-interactive';
 import { applyProductionTopicAndMessageTeams } from './remediation/topics/topic-monitor-production';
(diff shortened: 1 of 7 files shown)


packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -0,0 +1,99 @@
...
+	const dependencySubmissionWorkflow = actionsForRepo.find((action) =>
+		action.includes('scalacenter/sbt-dependency-submission'),
+	);
+	if (dependencySubmissionWorkflow) {
+		return true;
+	}
+	return false;
+}
+
+export function getSuitableRepoEvents(
```
NovemberTang, line 43: I think the name of this function is a bit confusing. I'm not clear on what an `Event` is, or what makes one `Suitable`.

To me `get`, implies a side-effect like a HTTP request or a DB call, but here it looks like we're just transforming data we already have, so maybe `find` or `create` would be more accurate?

Maybe something like `createSnsEventsForDependabotIntegration`

packages/repocop/src/index.ts
```diff
@@ -189,5 +192,15 @@ export async function main() {
 		);
 	}
 
+	// Dependency Graph Integrator
+	const productionWorkflowUsages: guardian_github_actions_usage[] =
```
NovemberTang, line 196: All the other DB calls are at the top of the function so that we can fail quickly if some data isn't available, to avoid ending up in a state where some parts of repocop have run and other parts haven't. This is handy because if we restore the data, we can run repocop again without worrying about double notifications. Could we move this call up with the others?

NovemberTang (approved):
Could we test this on CODE to make sure it works end-to-end (except for the actual raise the PR bit?)

tjsilver:
> Could we test this on CODE to make sure it works end-to-end (except for the actual raise the PR bit?)

Yes, will try again. Having trouble getting repocop to run on CODE due to a permissions error.

tjsilver:
Tested on CODE and worked as expected.

PR #1165, opened 2024-07-04
Open janus before exiting setup script

NovemberTang:
## What does this change?

Almost every time this script fails due to invalid AWS credentials, the user needs to manually navigate to Janus. Let's automate this step

## Why?

Shaving a few seconds off the startup time for users who don't already have credentials.

## How has it been verified?

Tested locally

### DIFF
--- a/packages/dev-environment/script/start
+++ b/packages/dev-environment/script/start
@@ -29,15 +29,22 @@ create_sql_migrations() {
   done
 }
 
+open_janus() {
+  echo "$1"
+  echo "Opening Janus"
+  sleep 3
+  open "https://janus.gutools.co.uk"
+}
+
 check_aws_credentials() {
   step "Checking AWS credentials"
 
   STATUS=$(aws sts get-caller-identity --profile deployTools 2>&1 || true)
   if [[ ${STATUS} =~ (ExpiredToken) ]]; then
-    echo "Credentials for the deployTools profile have expired. Please fetch new credentials, and run this script again."
+    open_janus "Credentials for the deployTools profile have expired. Please fetch new credentials, and run this script again."
     exit 1
   elif [[ ${STATUS} =~ ("could not be found") ]]; then
-    echo "Credentials for the deployTools profile are missing. Please fetch some, and run this script again."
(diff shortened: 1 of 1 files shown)


akash1810:
I'm not a huge fan of this, mainly because I don't need another browser tab open 😅.

> Let's automate this step

Not sure this change achieves this? As one still needs to switch to the browser tab, navigate to the deployTools credentials page, copy the credentials, and paste into the terminal. https://github.com/guardian/feria could be used (albeit it is archived), However I'm uncertain this is a problem that needs solving...

NovemberTang:
> I'm not a huge fan of this, mainly because I don't need another browser tab open 😅.
> 
> > Let's automate this step
> 
> Not sure this change achieves this? As one still needs to switch to the browser tab, navigate to the deployTools credentials page, copy the credentials, and paste into the terminal. https://github.com/guardian/feria could be used (albeit it is archived), However I'm uncertain this is a problem that needs solving...

For me, the tab is foregrounded immediately, so `switch to the browser tab` is handled. We could also change the url to point to the relevant permission page, eliminating `navigate to the deployTools credentials page`. I'm afraid I don't have an answer to the others! Maybe a hyperlink to Janus would be more useful?

tjsilver:
> > I'm not a huge fan of this, mainly because I don't need another browser tab open 😅.
> > > Let's automate this step
> > 
> > 
> > Not sure this change achieves this? As one still needs to switch to the browser tab, navigate to the deployTools credentials page, copy the credentials, and paste into the terminal. https://github.com/guardian/feria could be used (albeit it is archived), However I'm uncertain this is a problem that needs solving...
> 
> For me, the tab is foregrounded immediately, so `switch to the browser tab` is handled. We could also change the url to point to the relevant permission page, eliminating `navigate to the deployTools credentials page`, though that would expose our permissions names without some parameter-y/secret-y shenanigans. I'm afraid I don't have an answer to the others! Maybe a hyperlink to Janus would be more useful?

I'm not the original commenter, but I agree that a link would be most useful and least intrusive, here.

PR #1166, opened 2024-07-04
Vulnerability digest should show CVEs, to make two vulns from the same library look distinct

NovemberTang:
## What does this change?

When a single library was causing multiple CVEs to show up against a repo, the vulnerability digests can end up looking identical, especially if dependabot is being enabled for the first time. Let's include the CVE in the vulnerability digest to make it clearer that these are unique issues, as they may have slightly different remediations (ie one requires bumping to `x.y.z+1`, and another requires patching to `x.y.z+4`).

Also, removed boilerplate from the tests by setting up a vuln object at the beginning that is reused in subsequent tests.

## How has it been verified?

Added unit tests to verify expected behaviour

### DIFF
--- a/packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
+++ b/packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
@@ -70,6 +70,19 @@ const anotherResult: EvaluationResult = {
 	vulnerabilities: [],
 };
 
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
Has the digest view been tested in chat?

NovemberTang:
> Has the digest view been tested in chat?

No, but I've rendered the markdown and it looks normal!

PR #1174, opened 2024-07-09
feat(prisma): Create `aws_organizations_tree` view 

akash1810:
## What does this change?
AWS Organisations allows one to structure accounts into business units, with seemingly limitless levels.

We currently have something like this:

```
root
├── Editorial
│   └── multimedia
├── Product & Engineering
│   ├── Editorial Tools
│   │   ├── cmsFronts
│   │   └── composer
│   └── deployTools
└── glabs
```

The `aws_organizations_tree` view allows one to query accounts via a tree structure, allowing for ancestry queries. This differs from the `aws_organizations_account_parents` table, which only shows the immediate parent. The `aws_organizations_tree` view also includes each node in the tree: the root, the organisational units, and the accounts.

With `aws_organizations_tree` we can identify the `cmsFronts` account sits within `Product & Engineering`.

The format was inspired by https://leonardqmarcq.com/posts/modeling-hierarchical-tree-data. I opted against using the [`ltree` data type](https://www.postgresql.org/docs/current/ltree.html) as we're not dealing with large numbers of records (fewer than 100).

This change is split across two Prisma migrations (and [two commits](https://github.com/guardian/service-catalogue/pull/1174/commits)):
1. Add the necessary base tables to Prisma. This should be a no-op on CODE and PROD.
2. Create the view.

## Why?
A number of apps within Service Catalogue currently run against Product & Engineering (P&E) owned accounts, for example Obligatron. We're currently [evaluating AMI tag coverage for all accounts](https://github.com/guardian/service-catalogue/blob/9b05209f7f78cf27b8943121898d7b5d878648b2/packages/obligatron/src/obligations/tagging.ts#L49-L55), which is producing some noise. The next PR (https://github.com/guardian/service-catalogue/pull/1175) will utilise `aws_organizations_tree` to filter for P&E accounts.

## How has it been verified?
- Either run the project locally, run `SELECT * FROM aws_organizations_tree;`, and observe the results.
- OR [execute the query body on PROD](https://metrics.gutools.co.uk/goto/NXja3t_IR?orgId=1), and observe the results.

### DIFF
--- a/packages/common/prisma/migrations/20240709072758_additional_aws_organizations_tables/migration.sql
+++ b/packages/common/prisma/migrations/20240709072758_additional_aws_organizations_tables/migration.sql
@@ -1,0 +1,27 @@
+CREATE TABLE IF NOT EXISTS "aws_organizations_organizational_unit_parents" (
+    "_cq_sync_time" TIMESTAMP(6),
+    "_cq_source_name" TEXT,
+    "_cq_id" UUID NOT NULL,
+    "_cq_parent_id" UUID,
+    "request_account_id" TEXT NOT NULL,
+    "id" TEXT NOT NULL,
+    "parent_id" TEXT NOT NULL,
+    "type" TEXT NOT NULL,
+
+    CONSTRAINT "aws_organizations_organizational_unit_parents_cqpk" PRIMARY KEY ("_cq_id")
+);
+
+CREATE TABLE IF NOT EXISTS "aws_organizations_roots" (
+    "_cq_sync_time" TIMESTAMP(6),
+    "_cq_source_name" TEXT,
+    "_cq_id" UUID NOT NULL,
+    "_cq_parent_id" UUID,
+    "request_account_id" TEXT NOT NULL,
+    "tags" JSONB,
+    "arn" TEXT NOT NULL,
+    "id" TEXT NOT NULL,
+    "name" TEXT NOT NULL,
+    "policy_types" JSONB,
+
+    CONSTRAINT "aws_organizations_roots_cqpk" PRIMARY KEY ("_cq_id")
+);
--- a/packages/common/prisma/migrations/20240709073818_aws_organizations_tree/migration.sql
+++ b/packages/common/prisma/migrations/20240709073818_aws_organizations_tree/migration.sql
@@ -1,0 +1,72 @@
+-- Uses recursive common table expressions to build an ancestry of accounts and organisational units.
+CREATE OR REPLACE FUNCTION fn_aws_organizations_tree() RETURNS TABLE (
+    id                              TEXT
+    , type                          TEXT
+    , name                          TEXT
+    , ancestors                     TEXT[]
+    , is_product_and_engineering    BOOLEAN
+) AS $$
+    WITH RECURSIVE
+        nodes AS (
+            -- Root node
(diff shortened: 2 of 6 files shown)


packages/common/prisma/migrations/20240709072758_additional_aws_organizations_tables/migration.sql
akash1810, line 1: These tables already exist in CODE and PROD. The `IF NOT EXISTS` should mean this is a no-op.

packages/common/prisma/migrations/20240709073818_aws_organizations_tree/migration.sql
```diff
@@ -0,0 +1,72 @@
...
+        )
+        , productAndEngineering AS (
+            SELECT  id
+                    , name
+            FROM    aws_organizations_organizational_units
+            WHERE   name = 'Product & Engineering' -- assumes this string will never change
+    )
+
+    SELECT  tree.*
+            , productAndEngineering.id = ANY (tree.ancestors) AS is_product_and_engineering
```
akash1810, line 64: Whilst this boolean isn't strictly necessary for a tree view of the accounts, we perform a "P&E filter" pretty often. This boolean column makes this easier:

```sql
SELECT * FROM fn_aws_organizations_tree() WHERE is_product_and_engineering IS TRUE;
```

packages/common/prisma/views/public/aws_organizations_tree.sql
akash1810, line 1: This file is auto-generated by Prisma.

packages/dev-environment/config/cloudquery.yaml
```diff
@@ -9,6 +9,11 @@ spec:
     - aws_securityhub_findings
     - aws_s3_buckets
     - aws_ec2_images
+    - aws_organizations_accounts
+    - aws_organizations_account_parents
+    - aws_organizations_organizational_units
+    - aws_organizations_organizational_unit_parents
+    - aws_organizations_roots
```
akash1810, line 16: Collect these tables directly via CloudQuery (vs. copying from CODE) as there aren't many rows (fewer than 100).

AshCorr (approved):
Won't pretend to fully understand the recursive tree query but if it works it works! LGTM!

tjsilver (approved):
Looked at the query output on PROD, looks good.

packages/common/prisma/migrations/20240709073818_aws_organizations_tree/migration.sql
```diff
@@ -0,0 +1,72 @@
...
+    FROM    tree
+            , productAndEngineering;
+$$ LANGUAGE SQL;
```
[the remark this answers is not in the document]
AshCorr: Should be valid SQL! Postgres functions can be written in plain `SQL` or `PGSQL` (fancy SQL which looks more akin to a programming language) and you need to specifically tell Postgres what kind of function it is

akash1810:
> Won't pretend to fully understand the recursive tree query but if it works it works! LGTM!

I found this description from https://leonardqmarcq.com/posts/modeling-hierarchical-tree-data helpful:

> For each level in the tree, we append the ancestor we find into an array and go up one level until we hit the root node.

PR #1175, opened 2024-07-09
Add ancestry path to `aws_accounts` view

akash1810:
> [!NOTE]
> Builds on (and requires) #1174 and #1176.

## What does this change?
The `organizational_unit` column in the `aws_accounts` view shows the immediate parent. This change replaces this column with an `ancestors` column, detailing the full ancestry path.

A boolean `is_product_and_engineering` column is also added to make it easy to filter for P&E accounts:

```sql
SELECT * FROM aws_accounts WHERE is_product_and_engineering = TRUE;
```

> [!WARNING]
> This is a breaking change to the `aws_accounts` view.
> 
>  Columns removed:
>   - `organizational_unit` (`TEXT`)
>
> Columns added:
>   - `ancestors` (`TEXT[]`)
>   - `is_product_and_engineering` (`BOOLEAN`)

## Why?
A number of apps within Service Catalogue currently run against Product & Engineering (P&E) owned accounts, for example Obligatron. We're currently [evaluating AMI tag coverage for all accounts](https://github.com/guardian/service-catalogue/blob/9b05209f7f78cf27b8943121898d7b5d878648b2/packages/obligatron/src/obligations/tagging.ts#L49-L55), which is producing some noise.

## How has it been verified?
Run the project locally, run `SELECT * FROM aws_accounts;`, and observe the results.

I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/d6a40ce1-3065-4c3a-9274-2a7194ed7084), and [querying the view on CODE](https://metrics.code.dev-gutools.co.uk/goto/4q2Yc2lIg?orgId=1) shows the migration applied successfully.

### DIFF
--- a/packages/common/prisma/migrations/20240709110616_update_aws_accounts_view/migration.sql
+++ b/packages/common/prisma/migrations/20240709110616_update_aws_accounts_view/migration.sql
@@ -1,0 +1,43 @@
+/*
+ Redefine the view `aws_accounts`.
+
+ Removed:
+   - organizational_unit
+
+ Added:
+   - ancestors
+   - is_product_and_engineering
+ */
+
+BEGIN TRANSACTION;
+
+DROP VIEW IF EXISTS aws_accounts;
+DROP FUNCTION IF EXISTS fn_aws_accounts;
+
+CREATE FUNCTION fn_aws_accounts() RETURNS TABLE (
+    id                              TEXT
+    , name                          TEXT
+    , email                         TEXT
+    , status                        TEXT
+    , joined_timestamp              TIMESTAMP
+    , ancestors                     TEXT[]
+    , is_product_and_engineering    BOOLEAN
+) AS $$
+    SELECT acc.id
+           , acc.name
+           , acc.email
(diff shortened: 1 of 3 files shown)


packages/common/prisma/migrations/20240709110616_update_aws_accounts_view/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+
+ Removed:
+   - organizational_unit
+
+ Added:
+   - ancestors
+   - is_product_and_engineering
+ */
+
+BEGIN TRANSACTION;
```
AshCorr, line 12: Ooh good idea!

AshCorr (approved):

packages/common/prisma/migrations/20240709110616_update_aws_accounts_view/migration.sql
akash1810: This is leftover from when I was doing a cascading delete, but I think it's useful here anyway.

PR #1176, opened 2024-07-09
fix: Grant all database users blanket read access

akash1810:
## What does this change?
Extends #1154 to all database users.

## Why?
This'll make the changes in #1175 easier.

#1175 drops and recreates a view. To keep the current access status quo, we would need to re-grant permissions on each table the view queries, which is a bit longer than this option.

## How has it been verified?
Observe CI, the [checks](https://github.com/guardian/service-catalogue/blob/main/sql/ci.sql) should continue to pass.

I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/8950dbc0-e1ba-4dd0-8b40-3c2f7f2e1699), and the migration logs show the migration was successfully applied.


packages/common/prisma/migrations/20240709104714_user_permissions/migration.sql
akash1810, line 1: https://github.com/guardian/service-catalogue/pull/1154 already made these changes to the `repocop` and `obligatron` users. There's no harm doing it again though. Doing so makes this migration affect all our database users at once.

PR #1178, opened 2024-07-09
Update the dependency graph integrator PR body 

tjsilver:
## What does this change?

Adds a new step to the PR checklist for the dependency graph integrator to let users know that sbt needs to be v1.5+.

## Why?

We noticed that some repos have lower versions of sbt and the dependency submission action failed.

## How has it been verified?

Ran `dependency-graph-integrator` locally and the extra checklist item rendered correctly in the PR output.


NovemberTang (approved):
🥳

PR #1185, opened 2024-07-10
ci: are tests being run?

akash1810:
## What does this change?
I think the changes in https://github.com/guardian/service-catalogue/pull/1156 have broken CI such that the unit tests are not actually being run.

This PR should fail the build as 1 does not equal 2.


akash1810:
> are tests being run?

They are not! Closing.

PR #1196, opened 2024-07-10
Create cloudbuster lambda

marsavar:
## What does this change?
- Adds notification sending to cloudbuster
- Adds AWS infrastructure for cloudbuster to run on a Lambda on two separate schedules (one for critical findings, one for high findings)

## Why?
Currently Cloudbuster runs locally and digests aren't sent out.
We need these digests to be sent out to people.

> [!IMPORTANT]  
> The channel for sending notifications is currently set to Anghammarad Test Channel. Decide as a team whether to consolidate FSBP digests with vulnerability digest into a single email, then change the target.

## How has it been verified?
Triggered in AWS, witnessed the reports land in the dedicated Anghammarad Test Channel.


packages/cdk/lib/cloudbuster.ts
```diff
@@ -0,0 +1,82 @@
...
+				toleratedErrorPercentage: 0,
+				snsTopicName: 'devx-alerts',
+			},
+
+			/*
+			Override the default provided by GuCDK for improved compatability with https://github.com/guardian/cloudwatch-logs-management when producing log lines with markers.
+
+			Note: `logFormat` is a deprecated property, replaced with `loggingFormat`.
+			However, we can only specify one of `logFormat` or `loggingFormat`, and as GuCDK is setting `logFormat`, we have to do the same.
+			TODO - patch GuCDK.
```
akash1810, line 51: With https://github.com/guardian/service-catalogue/pull/1199, this TODO can be dropped, and we can use `loggingFormat`.

packages/cdk/lib/cloudbuster.ts
```diff
@@ -0,0 +1,72 @@
...
+			handler: 'index.main',
+			environment: {
+				ANGHAMMARAD_SNS_ARN: anghammaradTopic.topicArn,
+				DATABASE_HOSTNAME: db.dbInstanceEndpointAddress,
+				QUERY_LOGGING: 'false',
+			},
+			timeout: Duration.minutes(5),
+			// Unfortunately Prisma doesn't support streaming data from Postgres at the moment https://github.com/prisma/prisma/issues/5055
+			// This means that all rows need to be loaded into memory at the same time whenever a query is ran hence the high memory requirement.
+			memorySize: 4096,
```
NovemberTang, line 40: minor point, but I don't think 4gb is necessary. The function we're making to retrieve the data from the db roughly translates to

```
select * from aws_securityhub_findings
where generator_id like '%aws-foundational-security-best-practices/v/1.0.0%'
and (severity->>'Label' = 'HIGH' or  severity->>'Label'='CRITICAL')
```,

which only returns about 1000 rows. Repocop pulls in much more data than that, and has never exceeded 700mb

File: packages/cdk/lib/cloudbuster.ts [hunk 3]
```diff
@@ -0,0 +1,72 @@
...
+			targets: [
+				new LambdaFunction(lambda, {
+					event: RuleTargetInput.fromObject({ severities: ['CRITICAL'] }),
+				}),
+			],
+		});
+
+		new Rule(stack, `cloudbuster-high`, {
+			description: `Weekly execution of the Cloudbuster lambda for high findings`,
+			schedule: Schedule.cron({ weekDay: 'MON', hour: '9', minute: '0' }),
```
NovemberTang, line 62: minor: Generally I'd recommend avoiding Mondays because of all the bank holidays

packages/cdk/lib/cloudbuster.ts
marsavar: Thanks Akash - FWIW I bumped to the latest version (58.1.2 -> 58.1.4)  but couldn't synth anymore, as the `test-update` was failing with this message:

```
    Only define LogFormat or LoggingFormat, not both.
```
No biggie, I replaced all `logFormat` with `loggingFormat`, but perhaps this should've been a breaking change?

akash1810: > but perhaps this should've been a breaking change?

Hmm, not sure. The change in GuCDK was resolving a deprecation warning. There's no change to the underlying infrastructure. The changes only impact those lambdas that override the default provided by GuCDK. I'm not sure GuCDK needs to consider this class of usage with regard to it's release number. WDYT?

marsavar:
I'll squash and merged once approved.

packages/cloudbuster/src/digests.ts
```diff
@@ -15,18 +18,29 @@ export function createDigestsFromFindings(findings: Finding[]): Digest[] {
...
 	if (!teamFindings || teamFindings.length == 0) {
 		return undefined;
 	}
 
+	const accountName = teamFindings[0]?.awsAccountName as string;
+
+	const actions = [
+		{
+			cta: `View all ${teamFindings.length} findings on Grafana`,
+			url: `https://metrics.gutools.co.uk/d/ddi3x35x70jy8d/fsbp-compliance?var-account_name=${accountName}`,
```
NovemberTang, line 35:
```suggestion
			url: `https://metrics.gutools.co.uk/d/ddi3x35x70jy8d?var-account_name=${accountName}`,
```

The link still works without this string, and is slightly safer, as changing the name of the dashboard causes this string to change. It won't break anything, and isn't important enough to be a blocker, but might be slightly confusing for future travellers

NovemberTang (approved):
LGTM!

packages/cdk/lib/cloudbuster.ts
```diff
@@ -0,0 +1,74 @@
...
+			targets: [
+				new LambdaFunction(lambda, {
+					event: RuleTargetInput.fromObject({
+						severities: ['HIGH'],
+					}),
+				}),
+			],
+		});
+
+		db.grantConnect(lambda, 'repocop');
```
akash1810, line 72: This is a change from the current status quo where each lambda (/app) within Service Catalogue has it's own database user. Any particular reason to break this pattern?

packages/cloudbuster/src/digests.ts
```diff
@@ -36,13 +50,33 @@ function createEmailBody(findings: Finding[]): string {
...
+	digest: Digest,
+): Promise<void> {
+	console.log(`Sending digest to ${digest.accountId}...`);
+
+	await anghammaradClient.notify({
+		subject: digest.subject,
+		message: digest.message,
+		actions: digest.actions,
+		// target: { AwsAccount: digest.accountId },
+		target: { Stack: 'testing-alerts' },
```
akash1810, line 76: Worth pulling this out into a variable for use in the `console.log` above too?

```ts
const target = { Stack: 'testing-alerts' }; // TODO replace this with `{ AwsAccount: digest.accountId }` to send real alerts
console.log(`Sending digest to ${target}`);
```

packages/cloudbuster/src/config.ts
```diff
@@ -26,14 +30,14 @@ export async function getConfig(): Promise<Config> {
 		? await getDevDatabaseConfig()
 		: await getDatabaseConfig(stage, 'repocop'); //TODO create a new db user for cloudbuster before deploying.
 
-	const severities: SecurityHubSeverity[] = isDev
-		? ['CRITICAL', 'HIGH', 'MEDIUM', 'LOW', 'INFORMATION'] // Using all severities in DEV for more data.
-		: ['CRITICAL', 'HIGH'];
+	const anghammaradClient = isDev ? undefined : new Anghammarad();
```
akash1810, line 33: This feels strange to me, as it means we can only test the Anghammarad code paths in CODE and PROD, which is quite a long feedback loop. Might an alternative be to have [a flag](https://github.com/guardian/service-catalogue/blob/c98620a76584ff87ed90bc76f30af5e6f91623d1/.env#L51-L52) that's defaulted to false in DEV to control if a the `notify` method on the Anghammarad client is called?

```ts
const payload: AnghammaradMessagePayloadType = { };

if(messagingEnabled) {
  anghammaradClient.notify(payload);
} else {
  console.log(`Messaging disabled. Would have sent: ${JSON.stringify(payload}`);
}
```

Testing the Anghammarad code paths in DEV would then require a change of the boolean. It should also allow the removal of the typecast of `string | undefined` to `string` below:

```ts
topicArn: config.anghammaradSnsTopic as string,
```

packages/cloudbuster/src/digests.ts
```diff
@@ -15,18 +18,29 @@ export function createDigestsFromFindings(findings: Finding[]): Digest[] {
...
 	findings: GroupedFindings,
 ): Digest | undefined {
-	const teamFindings = findings[awsAccountId];
+	const teamFindings = findings[accountId];
 
 	if (!teamFindings || teamFindings.length == 0) {
 		return undefined;
 	}
 
+	const accountName = teamFindings[0]?.awsAccountName as string;
```
akash1810, line 30: I think destructuring the array, and checking for undefined on that should allow us to avoid the optional chaining an typecast:

```ts
const teamFindings = findings[accountId];

const [teamFinding, ] = teamFindings; // (attempt to) get the first finding

// check a finding has been found
if(!teamFinding) {
  return undefined;
}

const { awsAccountName } = teamFinding;
```

However I think we could make this more readable still by changing the input of this function, passing only the findings for the account, and not all the findings.

packages/cdk/lib/cloudbuster.ts
marsavar: I think all users have read access to the db now, so it felt a bit overkill to create a migration to add a new user for this lambda, though I agree that that should probably be the right thing to do for domain separation

PR #1200, opened 2024-07-11
Create obligatron mode for AWS infrastructure vulnerabilities

NovemberTang:
## What does this change

Creates a new obligatron mode that collects information about AWS infrastructure vulnerabilities. Currently, it only logs the output and does not write to the table. This is because we are currently not filtering out results that are within SLA. This will need to be done as a separate PR as further work is needed to unify:

1. the various `Severity` types
2. the calculations we perform on those types to calculate whether a resource is inside or outside SLA

So, that work will be done as a follow up PR. At that point, we will start writing to the DB

## Why?

So we can track departmental progress against our obligation to keep our AWS infrastructure compliant with AWS FSBP

## How has it been verified?

Verified locally. Added unit tests to confirm expected behaviour

### DIFF
--- a/packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
+++ b/packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
@@ -31,7 +31,6 @@ describe('The dependency vulnerabilities obligation', () => {
 
 	it('should return a result in the expected format', () => {
 		const actual = fsbpFindingsToObligatronResults([oneResourceFinding]);
-		console.log(actual);
 
 		const expected = {
 			contacts: {
@@ -50,8 +49,6 @@ describe('The dependency vulnerabilities obligation', () => {
 
 	it('should return multiple results if two resources are referenced in the same finding', () => {
 		const actual = fsbpFindingsToObligatronResults([twoResourceFinding]);
-		console.log(actual);
-
 		expect(actual.length).toEqual(2);
 	});
 
@@ -65,7 +62,6 @@ describe('The dependency vulnerabilities obligation', () => {
 			oneResourceFinding,
 			extraFinding,
 		])[0]?.reason;
-		console.log(actual);
 
 		expect(actual).toContain('S.1');
 		expect(actual).toContain('S.2');
--- a/packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
+++ b/packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
@@ -1,3 +1,4 @@
+import type { AwsContact } from '../obligations/index';
 import {
 	fsbpFindingsToObligatronResults,
 	type SecurityHubFinding,
@@ -65,5 +66,16 @@ describe('The dependency vulnerabilities obligation', () => {
 
 		expect(actual).toContain('S.1');
 		expect(actual).toContain('S.2');
+	});
+	it('should handle a resource with no tags', () => {
+		const finding = {
+			...oneResourceFinding,
+			resources: [{ ...resource1, Tags: {} }],
+		};
+
+		const actual = fsbpFindingsToObligatronResults([finding]);
+		const contacts = actual[0]?.contacts as AwsContact;
+
+		expect(contacts.Stack).toBeUndefined();
 	});
 });
--- a/packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
+++ b/packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
@@ -1,0 +1,73 @@
+import {
+	fsbpFindingsToObligatronResults,
+	type SecurityHubFinding,
+} from './aws-vulnerabilities';
+
+describe('The dependency vulnerabilities obligation', () => {
+	const resource1 = {
+		Id: 'arn:service:1',
+		Tags: { Stack: 'myStack' },
+		Region: 'some-region',
+		Type: 'some-type',
+	};
+
+	const resource2 = {
+		...resource1,
+		Id: 'arn:service:2',
+	};
+
+	const oneResourceFinding: SecurityHubFinding = {
+		resources: [resource1],
+		severity: { Label: 'HIGH', Normalized: 75 },
+		aws_account_id: '0123456',
+		first_observed_at: new Date('2020-01-01'),
+		product_fields: { ControlId: 'S.1', StandardsArn: 'arn:1' },
+	};
+
+	const twoResourceFinding: SecurityHubFinding = {
+		...oneResourceFinding,
+		resources: [resource1, resource2],
+	};
+
+	it('should return a result in the expected format', () => {
+		const actual = fsbpFindingsToObligatronResults([oneResourceFinding]);
+		console.log(actual);
+
+		const expected = {
+			contacts: {
+				App: undefined,
+				Stack: 'myStack',
+				Stage: undefined,
+				aws_account_id: '0123456',
+			},
+			reason: 'The following AWS FSBP controls are failing: S.1',
+			resource: 'arn:service:1',
+			url: 'https://docs.aws.amazon.com/securityhub/latest/userguide/fsbp-standard.html',
+		};
+
+		expect(actual).toEqual([expected]);
+	});
+
+	it('should return multiple results if two resources are referenced in the same finding', () => {
+		const actual = fsbpFindingsToObligatronResults([twoResourceFinding]);
+		console.log(actual);
+
+		expect(actual.length).toEqual(2);
+	});
+
+	it('should list multiple control IDs in one finding if the same resource has failed two controls', () => {
+		const extraFinding = {
+			...oneResourceFinding,
+			product_fields: { ControlId: 'S.2', StandardsArn: 'arn:1' },
+		};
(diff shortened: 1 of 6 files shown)


packages/obligatron/src/obligations/aws-vulnerabilities.ts
```diff
@@ -0,0 +1,108 @@
...
+function groupFailuresByResource(
+	failures: Failure[],
+): Record<string, Failure[]> {
+	const grouped: Record<string, Failure[]> = {};
+
+	for (const failure of failures) {
+		if (!grouped[failure.resource]) {
+			grouped[failure.resource] = [];
+		}
+		// @ts-expect-error - TS doesn't understand that we've just checked for the key
```
akash1810, line 51: [Using a `reduce`](https://www.typescriptlang.org/play/?ts=5.5.3#code/C4TwDgpgBAYghgSwDYFcBO0C8UDeAoASAwGcB7dAYwgC4pjg0EA7AcwG5CLSmHSkBJACa16jVhwJwKXFDyEiGzdoWBwWxWgCUIXNIIA8opQBo6i1gD4OAXw54uTelABmiVCVrxk6CAG0AulDYvnhQYbih4VEk5GhUtADkAEYoFADWEMAJxpFRYQ68AsJQCQCKAOoAotm5eVIycsUJAAytNXnhquq0ONa11jlR+B1hMZQ0JSnpme0jBWh88iWaACoAmrMd9eSNia3Nm3ldGrh9UQO1wyNjcRMJSHAAtkmCcIdR84tNAKr8APLvcLbWTAJYtNqDEbHHpncJ9fx4ezcJwsBYoSCCIIuNw+YgAOgwglSEH02l0BiMrFMXncfn8FgsAApGfVTAhgBBHgBKIIWCIfZHAXBQG5UKDWLHszl2KIIZws6S+UUQfw8q4dDDAdBMfkjKJ4g2s2p6pUQMjjfy0XwGvH1U3m27+Nkcx4IvWwsJ9Wqa7W6jo2o0m5WWqC+KWu-p4AanLmIhxkJAQPFIUgsRmo8gYrlAA) would resolve this issue:

```ts
function groupFailuresByResource(failures: Failure[]): Record<string, Failure[]> {
  return failures.reduce<Record<string, Failure[]>>((acc, item) => {
      const { resource } = item;
  
      if(acc[resource]) {
          return {
              ...acc,
              [resource]: [...acc[resource], item]
          }
      }
  
      return {
          ...acc,
          [resource]: [item]
      }
  }, {});
}
```

We could also convert to a ternary, but I think that is less accessible:

```ts
return failures.reduce<Record<string, Failure[]>>((acc, item) => {
    const { resource } = item;

    return {
        ...acc,
        [resource]: acc[resource] ? [...acc[resource], item] : [item]
    }
}, {})
```

packages/obligatron/src/obligations/aws-vulnerabilities.ts
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return grouped;
+}
+
+function failuresToObligationResult(
+	arn: string,
+	failures: Failure[],
+): ObligationResult {
+	const oneFailure = failures[0]; //this should always exist because groupFailuresByResource should always return at least one failure
```
akash1810, line 62: I don't think this is a strong assumption we can make, as there's no control over the caller of `failuresToObligationResult` having previously used `groupFailuresByResource`. Equally, the type of `failures` doesn't indicate prior usage of `groupFailuresByResource`.

packages/obligatron/src/obligations/aws-vulnerabilities.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export async function evaluateFsbpVulnerabilities(
+	client: PrismaClient,
+): Promise<ObligationResult[]> {
+	const findings = (await getFsbpFindings(client, ['CRITICAL', 'HIGH'])).map(
+		(v) => v as unknown as SecurityHubFinding,
+	);
+
+	const results = fsbpFindingsToObligatronResults(findings).slice(0, 5);
```
akash1810, line 104: Why aren't we processing the entire array?

packages/obligatron/src/obligations/aws-vulnerabilities.ts
```diff
@@ -0,0 +1,108 @@
...
+	console.log(results);
+
+	return []; //failuresToObligationResults(failuresByResource).slice(0, 5);
```
akash1810, line 107: Redundant comment?

packages/obligatron/src/obligations/aws-vulnerabilities.ts
```diff
@@ -0,0 +1,108 @@
...
+): ObligationResult[] {
+	const allFailures = findings.flatMap(findingToFailures);
+	const failuresByResource = groupFailuresByResource(allFailures);
+	return failuresToObligationResults(failuresByResource);
+}
+
+export async function evaluateFsbpVulnerabilities(
+	client: PrismaClient,
+): Promise<ObligationResult[]> {
+	const findings = (await getFsbpFindings(client, ['CRITICAL', 'HIGH'])).map(
```
akash1810, line 100: If these severities were passed via configuration, would it make local development easier?

packages/obligatron/src/obligations/aws-vulnerabilities.ts
akash1810: One alternative might be to assert on the distinctness of account id:

```ts
const accountIds = new Set(failures.map(_ => _.accountId))
if(accountIds.size !== 1) {
    throw new Error("you called this function incorrectly!")
}
const [accountId, ] = Array.from(accountIds)
```

packages/obligatron/src/obligations/aws-vulnerabilities.test.ts
akash1810, line 1: I think there are some redundant `console.log`s across this file?

packages/obligatron/src/obligations/aws-vulnerabilities.ts
NovemberTang: why would it make local development easier?

packages/obligatron/src/obligations/aws-vulnerabilities.ts
NovemberTang: I'm visually inspecting the array to sense check it while in development. I'm not sure there's much benefit to logging more than 5 or so. We'll process the whole array when we're ready to write to the table. We could move the `slice()` call to inside the `console.log()`, but we're going to get rid of it in a couple of days anyway

packages/obligatron/src/obligations/aws-vulnerabilities.ts
NovemberTang: So I tried reduce with @marsavar , but in addition to being quite difficult to read, we kept running up against a (we think, incorrect( typecheck error that we would have had to `ts-ignore` anyway. We've added a `?` and it seems to have solved the issue

packages/obligatron/src/obligations/aws-vulnerabilities.ts
```diff
@@ -0,0 +1,108 @@
...
+		(v) => v as unknown as SecurityHubFinding,
```
AshCorr, line 101: This feels like a bit of a spooky typecast! Are we sure that our schema will always match AWSs? Could lead to some annoying to debug errors if not!

We're quite explicit with type checking in the TAGGING job (maybe too much!), I wonder if we could have any shared behaviour for checking the schema matches? https://github.com/guardian/service-catalogue/blob/ba131950970d3e41d4925d0ec7b2bd387d8903f4/packages/obligatron/src/obligations/tagging.ts#L40-L44

AshCorr (approved):

packages/obligatron/src/obligations/aws-vulnerabilities.ts
NovemberTang: I've used `toNonEmptyArray` to guarantee the array has at least one element or throws an error

packages/obligatron/src/obligations/aws-vulnerabilities.ts
NovemberTang: `as unknown as <CUSTOM TYPE>` is a pretty common pattern across the service catalogue, especially in repocop. If we want to move away from it, could that be a separate piece of work?

packages/obligatron/src/obligations/aws-vulnerabilities.ts
akash1810: Since Node 21, this is available in the standard library 😍  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy.

packages/obligatron/src/obligations/aws-vulnerabilities.ts
akash1810: Was thinking of this within `cloudbuster` https://github.com/guardian/service-catalogue/blob/eb35b3e740c43cf637aff823e3a9ab4b42b043b9/packages/cloudbuster/src/config.ts#L29-L31.

PR #1208, opened 2024-07-15
ci: Silence console messages to make build log easier to read

akash1810:
## What does this change?
Silence the console during the CI build to make the build log easier to read, and see build errors.

## How has it been verified?
Observe CI log on this branch, and compare it to the build log of main. Has it improved?

PR #1209, opened 2024-07-15
Aa/quieter ci test

akash1810:
## What does this change?
Testing https://github.com/guardian/service-catalogue/pull/1208, specifically https://github.com/guardian/service-catalogue/pull/1208#pullrequestreview-2177310238.


akash1810:
Test errors are shown in the build log:

image

And as an annotation:

image

AshCorr:
image

Neato! My concern was that `expect` used console.log to print the difference but it looks like it throws an error!

PR #1210, opened 2024-07-15
Filter out FSBP results within SLA time and write to obligatron DB

NovemberTang:
## What does this change?

- Moves `isWithinSlaTime` and associated tests to common
- Filter FSBP obligatron results so that anything within SLA is excluded

## Why?

So we can start collecting accurate data about which resources have out of SLA FSBP vulnerabilities.

## How has it been verified?

Tests have been moved, CI passes, local execution produces the expected results

### DIFF
--- a/packages/cloudbuster/src/findings.test.ts
+++ b/packages/cloudbuster/src/findings.test.ts
@@ -1,10 +1,5 @@
-import { groupFindingsByAccount, isWithinSlaTime } from './findings';
+import { groupFindingsByAccount } from './findings';
 import type { Finding, GroupedFindings } from './types';
-
-const MOCK_TODAY = new Date('2024-01-10');
-const MOCK_ONE_DAY_AGO = new Date('2024-01-09');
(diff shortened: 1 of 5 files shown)


tjsilver:
I noticed that all the FSBP findings for EC2.8 were within SLA time, even though they've probably had the issues for some time - but we didn't track it until recently, so it's coming up within the 30 days for Highs. Are we confident that in the context of this change, that doesn't matter?

tjsilver (approved):
Looks good! (See other comment)

NovemberTang:
> I noticed that all the FSBP findings for EC2.8 were within SLA time, even though they've probably had the issues for some time - but we didn't track it until recently, so it's coming up within the 30 days for Highs. Are we confident that in the context of this change, that doesn't matter?

Yeah, this is a recurring problem. I think we'll need a dedicated solution for what to do about EC2 issues, as it's also creating misleading dashboards, but it shouldn't block this ticket.

PR #1227, opened 2024-07-31
feat: Add Kotlin to Dependency Graph Integrator

tjsilver:
## What does this change?

- The Dependency Graph Integrator takes a language parameter, and generates the workflow yaml and PR for both Scala and Kotlin.
- We now refer to the language rather than the package manager in the PR, as this makes things simpler. 

## Why?

We want to support Kotlin repos, and this now gives the flexibility to add further languages/workflows to the integrator.

## How has it been verified?
- [x] Tested locally
- [x] Tested on CODE

### DIFF
--- a/packages/dependency-graph-integrator/src/file-generator.test.ts
+++ b/packages/dependency-graph-integrator/src/file-generator.test.ts
@@ -2,7 +2,7 @@ import { createYaml } from './file-generator';
 
 describe('createYaml for Scala', () => {
 	it('should generate the following yaml file', () => {
-		const yaml = createYaml('branch', 'Scala');
+		const yaml = createYaml('branch', 'Scala', 'repo1');
 		const result =
 			String.raw`name: Update Dependency Graph for Scala
 on:
@@ -33,9 +33,9 @@ jobs:
 	});
 });
 
-describe('createYaml for Scala', () => {
+describe('createYaml for Kotlin', () => {
 	it('should generate the following yaml file', () => {
-		const yaml = createYaml('branch', 'Kotlin');
+		const yaml = createYaml('branch', 'Kotlin', 'repo2');
 		const result =
 			String.raw`name: Update Dependency Graph for Kotlin
 on:
@@ -51,13 +51,19 @@ jobs:
       - name: Checkout branch
         id: checkout
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Set up Java
+        id: setup
+        uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 # v4.2.1
+        with:
+          distribution: temurin
+          java-version: 17
       - name: Submit dependencies
         id: submit
         uses: gradle/actions/dependency-submission@d9c87d481d55275bb5441eef3fe0e46805f9ef70 # v3.5.0
       - name: Log snapshot for user validation
         id: validate
-        run: cat` +
-			' ${{ steps.submit.outputs.snapshot-json-path }} | jq' + // Need to split this line to avoid syntax errors due to the template string
+        run: cat ` + // Need to split this line to avoid errors due to new line produced in yaml
+			'/home/runner/work/repo2/repo2/dependency-graph-reports/update_dependency_graph_for_kotlin-dependency-graph.json\n          | jq' +
 			String.raw`
     permissions:
       contents: write
--- a/packages/dependency-graph-integrator/src/file-generator.test.ts
+++ b/packages/dependency-graph-integrator/src/file-generator.test.ts
@@ -1,10 +1,10 @@
 import { createYaml } from './file-generator';
 
-describe('createYaml', () => {
+describe('createYaml for Scala', () => {
 	it('should generate the following yaml file', () => {
-		const yaml = createYaml('branch');
+		const yaml = createYaml('branch', 'Scala');
 		const result =
-			String.raw`name: Update Dependency Graph for SBT
+			String.raw`name: Update Dependency Graph for Scala
 on:
   push:
     branches:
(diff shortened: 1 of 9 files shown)


packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -1,10 +1,65 @@
...
+				run: `cat /home/runner/work/${repo}/${repo}/dependency-graph-reports/update_dependency_graph_for_kotlin-dependency-graph.json | jq`,
+			},
+		],
+	};
+}
+
+export function createYaml(
+	prBranch: string,
+	language: DepGraphLanguage,
+	repo: string,
```
tjsilver, line 59: We need the repo as a parameter here as it's used to log out the snapshot for Kotlin deps

packages/dependency-graph-integrator/src/index.ts
```diff
@@ -12,19 +12,20 @@ import {
...
 	const boardNumber = 110;
-	const author = 'gu-dependency-graph-integrator'; // TODO: create new 'gu-dependency-graph-integrator' app
-	const title =
-		'Submit sbt dependencies to GitHub for vulnerability monitoring';
-	const fileName = '.github/workflows/sbt-dependency-graph.yaml';
-	const commitMessage = 'Add sbt-dependency-graph.yaml';
-	const yamlContents = createYaml(branch);
+	const author = 'gu-dependency-graph-integrator';
+	const title = `Submit ${event.language} dependencies to GitHub for vulnerability monitoring`;
+	const fileName = `.github/workflows/${languageLowerCase}-dependency-graph.yaml`;
```
NovemberTang, line 24: I'm thinking about a downstream piece of work, which is updating repocop to detect these. As the naming is now inconsistent (some will be `scala-`, and some will be `sbt-`, i guess we can just wildcard search for `*-dependency-graph.yaml` and make sure the correct action is in the file?

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -1,10 +1,65 @@
 import { markdownChecklist } from 'common/src/string';
+import type { DepGraphLanguage } from 'common/types';
 import { h2, p, tsMarkdown } from 'ts-markdown';
 import { stringify } from 'yaml';
 
-export function createYaml(prBranch: string): string {
+function createYamlLanguageSteps(
```
NovemberTang, line 6: could we rename this to something more specific, maybe `createLanguageSpecificWorkflowSteps`

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -14,62 +69,68 @@ export function createYaml(prBranch: string): string {
...
+	const finalSteps = [
+		'A run of this action should have been triggered when the branch was ' +
+			'created. Sense check the output of "Log snapshot for user validation", ' +
+			'and make sure that your dependencies look okay.',
+		`When you are happy the action works, remove the branch name \`${branchName}\` ` +
+			'trigger from the the yaml file (aka delete line 6), approve, and merge. ',
+	];
+	stepsForLanguage.forEach((step) => allSteps.push(step));
+	finalSteps.forEach((step) => allSteps.push(step));
+	return allSteps;
```
NovemberTang, line 114: would something like `return [...stepsForLanguage, ...finalSteps]` work? and then we wouldn't need to create and mutate a new array?

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -14,62 +69,68 @@ export function createYaml(prBranch: string): string {
...
 	return stringify(dependencyGraphWorkflowJson, { lineWidth: 120 })
-		.replace('{}', '')
+		.replaceAll('{}', '')
 		.replaceAll('"', '');
 }
 
-function createPRChecklist(branchName: string): string[] {
-	const step1 =
-		'Ensure that the [version of sbt in the project is v1.5 or above](https://github.com/scalacenter/sbt-dependency-submission?tab=readme-ov-file#support) in order for the dependency submission action to run.';
+const stepsForLanguages: Record<DepGraphLanguage, string[]> = {
```
NovemberTang, line 83: Why are we using a `string[]` as the value, when there's only one element in each list?

tjsilver: Ah, originally there was more than one string in the arrays, will change the type for now.

tjsilver: There was originally more than one step for a language.

packages/dependency-graph-integrator/src/file-generator.ts
```diff
@@ -14,62 +69,68 @@ export function createYaml(prBranch: string): string {
...
-export function generatePrBody(branchName: string, repoName: string): string {
+export function generatePrBody(
+	branchName: string,
+	repoName: string,
+	language: DepGraphLanguage,
+): string {
 	const body = [
 		h2('What does this change?'),
 		p(
-			'This PR sends your sbt dependencies to GitHub for vulnerability monitoring via Dependabot. ' +
```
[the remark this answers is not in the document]
tjsilver: Yeah I've reverted it back to using the package manager rather than the language, so this should be clearer now.

packages/dependency-graph-integrator/src/file-generator.test.ts
```diff
@@ -32,3 +32,42 @@ jobs:
 		expect(yaml).toEqual(result);
 	});
 });
+
+describe('createYaml for Kotlin', () => {
+	it('should generate the following yaml file', () => {
```
akash1810, line 37: minor: Did we consider using [Jest's snapshot testing](https://jestjs.io/docs/snapshot-testing) here instead? I think it'll provide a simpler DX and removes the usage of `String.raw` and string concatenation.

tjsilver: Not sure, as I didn't write the original test in this file, but this looks interesting. Perhaps something for a future PR?

NovemberTang (approved):

PR #977, opened 2024-05-03
feat(dashboard): Initial Grafana Scene plugin

akash1810:
> [!NOTE]
> It might be easiest to review this PR [commit by commit](https://github.com/guardian/service-catalogue/pull/977/commits).

Currently, we're creating Grafana dashboards by hand. Whilst this works, there are a few notable limitations:
- Code sharing is difficult
- Gaining an audit trail is difficult
- Getting a review of a change is not possible (unless pairing)

## What does this change, and why?
This change adds a new package to create a [Grafana Scene](https://grafana.com/developers/scenes). Scenes allow us to create dashboards (technically [apps](https://grafana.com/developers/plugin-tools/introduction/plugin-types-usage)) as a React app, and improves on the aforementioned limitations.

This change adds a large number of new files; most of them being created with the project generator: `npx @grafana/create-plugin@latest`.

There are some additional changes to fit the project into this repository, including:
- Applying our departmental configuration for eslint, Prettier and TypeScript.
- Merging the developer environment into the existing one, so we can continue to run the project via `npm -w dev-environment start`.
- Adding a simple query to the Service Catalogue database, demonstrating end to end functionality.

These changes have been [commented on](https://github.com/guardian/service-catalogue/pull/977/files).

The generator has added some potentially superfluous files, such as [Playwright](https://playwright.dev/) configuration for [end to end testing](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/introduction). There aren't actually any Playwright tests yet, so they're not being run in CI. I'd suggest we rationalise the files in a follow-up PRs; keeping this one as close to "standard" as possible reduces the chance of misconfiguring something.

### Distribution
[Grafana's documentation](https://grafana.com/developers/plugin-tools/publish-a-plugin/sign-a-plugin) for Scene distribution includes signing to improve verification. This requires creating an Enterprise account.

This change does not including signing. Instead, the Scene is distributed in the same way as most of our applications - uploading to a private S3 bucket, via Riff-Raff deployment. This means we'll [install the Scene from a local file](https://grafana.com/docs/grafana/latest/administration/plugin-management/#install-plugin-on-local-grafana). See https://github.com/guardian/grafana/pull/359 as an example. It also means two steps are required to deploy changes:
1. Deploy Service Catalogue
2. Deploy Grafana

I suspect we'll want to sign the plugin in future; I'd suggest we investigate this once Scenes have proven to be a good fit for the project.

### Further improvements
[...]

PR #1257, opened 2024-09-04
feat: remove dashboard

tjsilver:
## What does this change?

This PR removes the dashboard package, a Grafana Scenes project. This essential reverses https://github.com/guardian/service-catalogue/pull/977.

## Why?

The package has not been used due to the fact that it is difficult and fiddly to implement. Removing it reduces the maintenance burden and makes upgrading deps in the repo easier. Merging this PR will hopefully allow [Octokit to be upgraded which is now ESM only](https://github.com/guardian/service-catalogue/pull/1256).

## How has it been verified?

Grafana runs locally successfully.


NovemberTang:
> ## How has it been verified?

Could we say some more about how this has been tested? For example, are we still able to use grafana locally?

package.json
NovemberTang, line 1: What are these changes about?

PR #1261, opened 2024-09-06
Repocop should skip resolved snyk issues

NovemberTang:
## What does this change?

- make `snyk_issues.attributes.status` a union type, that can either be `open` or `resolved`
- filter out `resolved` issues
- test the new filter

## Why?

Resolved issues have been remediated, so repocop doesn't care about them.

## How has it been verified?

A new unit test has been added to verify this behaviour.
After running on CODE, we see 80% fewer snyk issues loaded into the lambda's memory, and a reduction in runtime from 30s to 22s.

### DIFF
--- a/packages/repocop/src/evaluation/repository.ts
+++ b/packages/repocop/src/evaluation/repository.ts
@@ -279,9 +279,9 @@ export function collectAndFormatUrgentSnykAlerts(
 		})
 		.map((project) => project.id);
 
-	const snykIssuesForRepo: SnykIssue[] = snykProjectIdsForRepo
-		.flatMap((projectId) => getIssuesForProject(projectId, snykIssues))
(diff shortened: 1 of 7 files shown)


packages/repocop/src/evaluation/repository.ts
```diff
@@ -279,9 +279,9 @@ export function collectAndFormatUrgentSnykAlerts(
 		})
 		.map((project) => project.id);
 
-	const snykIssuesForRepo: SnykIssue[] = snykProjectIdsForRepo
-		.flatMap((projectId) => getIssuesForProject(projectId, snykIssues))
-		.filter((i) => !i.attributes.ignored);
```
NovemberTang, line 284: this logic has been moved to a function in `utils.ts`, and is utilised in the index, when we first load snyk issues

tjsilver (approved):
Great

PR #1262, opened 2024-09-10
temporarily switch off cloudbuster

NovemberTang:
## What does this change?

Stops cloudbuster from sending notifications every day

## Why?

We don't need to test this lambda every day. And it generates a bunch of anghammarad failures that we are not going to fix any time soon, as the product is not being rolled out to teams yet


tjsilver (approved):
👍 makes sense

PR #1265, opened 2024-09-11
Change vulnerability digest cutoff from a fixed date to a 60 day window

NovemberTang:
## What does this change?

Only show vulnerabilities from the last 60 days in vulnerability digests

## Why?

Previously, we were sending teams notifications, potentially daily, about vulnerabilities that were nearly six months old. I think vulnerabilities that old can reasonably be considered part of a 'historical backlog', and continuing to pester teams about it is unlikely to result in meaningful change. It may be actively detrimental, as it causes alert fatigue, resulting in teams ignoring messages.

#### Why 60 days?

The choice of 60 days is slightly arbitrary, and if it doesn't work, we can always change it later. I've made this value configurable, so we could, for example, experiment with different windows for different severities; but doubling the window given to resolve highs seemed like a reasonable starting point.

## How has it been verified?

Unit tests have been modified to test the desired behaviour

### DIFF
--- a/packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
+++ b/packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
@@ -163,21 +163,26 @@ describe('createDigest', () => {
 		expect(anotherDigest?.message).toContain('rightpad');
 	});
 
-	it('only returns vulnerabilities created after 30th April 2024', () => {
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
This seems like a sensible amount of time and as you said, we can monitor how it goes and change accordingly. 👍

PR #1268, opened 2024-09-11
Remove unnecessary newlines from mermaids diagram

NovemberTang:
## What does this change?

- Markdown formatted numbered lists don't show properly in the readme, so slightly tweak them
- \n is not an escape character, so remove it from the chart

## How has it been verified?

Mermaids diagram now renders properly. Check the rich diff


tjsilver (approved):
Looks lovely!

PR #1269, opened 2024-09-12
Update and expand the Dependency Graph Integrator docs

NovemberTang:
## What does this change?

- Corrects the lambda input
- Gives an example of the lambda output
- Some diagrams showing how the lambda works, and how the dependency submisison workflow works

View the rich diff for a rendered README

## Why?

Better documentation means less time spent trying to figure out how stuff works

## How has it been verified?

README renders correctly


tjsilver (approved):
Looks great!

PR #1271, opened 2024-09-17
parameterise github org

NovemberTang:
## What does this change?

the `guardian` github org shows up all the time in our codebase, let's make it a constant that we set via config, and pass it in via CDK. This PR does this for repocop and cloudquery. Follow-up PRs will be raised for other packages. 

## Why?

Magic values are bad! And this will make it easier for people outside the guardian to use and modify the service catalogue if they wish

## How has it been verified?

- [x] CI passes.
- [x] Local execution
- [x] CODE execution

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -62,6 +62,11 @@ interface ServiceCatalogueProps extends GuStackProps {
 	 */
 	rdsDeletionProtection?: boolean;
 	multiAz?: boolean;
+
+	/**
+	 * The GitHub org to search for repositories in.
+	 */
+	gitHubOrg?: string;
 }
 
 export class ServiceCatalogue extends GuStack {
@@ -71,7 +76,11 @@ export class ServiceCatalogue extends GuStack {
 		const { stage, stack } = this;
 		const app = props.app ?? 'service-catalogue';
 
-		const { rdsDeletionProtection = true, multiAz = false } = props;
+		const {
+			rdsDeletionProtection = true,
+			multiAz = false,
+			gitHubOrg = 'guardian',
+		} = props;
 
 		const nonProdSchedule = props.schedule;
 
@@ -172,8 +181,6 @@ export class ServiceCatalogue extends GuStack {
 			effect: Effect.ALLOW,
 			resources: [loggingStreamArn],
 		});
-
-		const gitHubOrg = 'guardian';
 
 		const cloudqueryCluster = addCloudqueryEcsCluster(this, {
 			nonProdSchedule,
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -173,6 +173,8 @@ export class ServiceCatalogue extends GuStack {
 			resources: [loggingStreamArn],
 		});
 
+		const gitHubOrg = 'guardian';
+
 		const cloudqueryCluster = addCloudqueryEcsCluster(this, {
 			nonProdSchedule,
 			db,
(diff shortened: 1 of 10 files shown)


tjsilver (approved):
Good idea to do this!

akash1810 (commented):
Worth adding a comment around [the GitHub secret](https://github.com/guardian/service-catalogue/blob/53cc41ed15ac036e58d7c82bf076923a21787106/packages/cdk/lib/cloudquery/index.ts#L346-L367) mentioning it should be scoped to the correct organisation?

packages/cdk/lib/service-catalogue.ts
```diff
@@ -173,6 +173,8 @@ export class ServiceCatalogue extends GuStack {
 			resources: [loggingStreamArn],
 		});
 
+		const gitHubOrg = 'guardian';
```
akash1810, line 176: Should we push this property up into `ServiceCatalogueProps`, to reinforce the idea that variance is added in the entry point [bin/cdk.ts](https://github.com/guardian/service-catalogue/blob/main/packages/cdk/bin/cdk.ts)?

packages/repocop/src/query.ts
```diff
@@ -128,17 +128,19 @@ export async function getRepositoryLanguages(
...
+		repoName = repoName.replace(prefix, '');
 	}
 
 	try {
 		const alert: DependabotVulnResponse =
 			await octokit.rest.dependabot.listAlertsForRepo({
-				owner: 'guardian',
-				repo: name,
+				owner: orgName,
+				repo: repoName,
```
akash1810, line 143: Is this overly complex? IIUC the `name` parameter looks like `guardian/service-catalogue`, or more abstractly `org/repo`. Rather than taking in the org, then stripping it from `repoName` could we instead do:

```ts
const [ org, repo ] = repoName.split('/');
```

Maybe also rename `repoName` to `fullRepoName` for added clarity too?

packages/cdk/lib/service-catalogue.ts
akash1810: IMO this would be a refactor; I don't think the snapshot would require updating. That is, the change would be a no-op and wouldn't require explicit testing on CODE. Either way works.

packages/repocop/src/query.ts
NovemberTang: There's actually some defensive programming going on here. `repoName` is supposed to be the short name, i.e. `service-catalogue`. However, this logic exists so that we don't error out if someone passes in the long repo name by accident (for example, while testing)

akash1810: Ah. Could we use the [type system](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) here instead ([example](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAYgrgGwQJQmA9gOQIYFtoC8UABgOZzYBOAJgJbYB2A9ACQDeAzsJbQ6QL7EA3ACgRAMzgMAxsFroGUaugDKwOOPEAKSUlQYAXLEQo0WPBACUUNiKj2o0hVygBtdJVIAaKJTMBdKCJdUwwAOg4wBFpgLQByJjjLUQcoOwcnBg50BAgwhHRSLWIAdQ8Aa15SKAVfMwN2Pwx+KF4azwa2DwFiZJF+MWU1DW0AInIqOkYmDghKADdaaQgAWmlsYGwC8ghRvpEmJlhsWgQOKGAAC2hQSGlr6XKDo6H1TS1R66R0PaEgA))?

NovemberTang: Is there a way to get that to do the opposite of what it's doing now? ie only fail if the `/` is present?

PR #1272, opened 2024-09-17
feat: update hasDependencyTracking to include repos with dep graph wo…

tjsilver:
## What does this change?

Changes the logic of the `hasDependencyTracking` function to include repos that have dependency submission workflows. The logic is a bit convoluted. The check now returns true for repos (not on Snyk) that have all their languages covered by dependabot and dependency graph integrator (Scala and Kotlin), as long as they also have the associated dependency submission workflow for each dependency graph integrator-supported language.

Tests have been added to cover these cases.

## Why?

This will give us a more accurate representation of repos with dependency vulnerability tracking.

## How has it been verified?

- [x] Tests pass locally
- [x] Vulns table reports 'true' for 'vulnerability_tracking' when repo only has dep submission workflow (no Snyk) and reports 'false' on 'main' (guardian/apps-anomaly-detection)
- [ ] Vulns table correctly reports for repos without dep submission workflows.

### DIFF
--- a/packages/repocop/src/evaluation/repository.test.ts
+++ b/packages/repocop/src/evaluation/repository.test.ts
@@ -539,7 +539,6 @@ describe('REPOSITORY_09 - Dependency tracking', () => {
 		);
 		expect(actual).toEqual(true);
 	});
-	//TODO: update this test to include dep graph integrator
 	test('is not valid if a project is not on snyk, and uses a language dependabot/dependency graph integrator does not support', () => {
 		const repo: Repository = {
 			...nullRepo,
--- a/packages/repocop/src/evaluation/repository.test.ts
+++ b/packages/repocop/src/evaluation/repository.test.ts
@@ -1,6 +1,7 @@
 import type {
 	github_languages,
 	github_repository_branches,
+	guardian_github_actions_usage,
 	view_repo_ownership,
 } from '@prisma/client';
 import type { RepocopVulnerability, Repository } from 'common/src/types';
@@ -32,6 +33,7 @@ function evaluateRepoTestHelper(
 	latestSnykIssues: SnykIssue[] = [],
 	snykProjects: SnykProject[] = [],
 	reposOnSnyk: string[] = [],
+	workflowsForRepo: guardian_github_actions_usage[] = [],
 ) {
 	return evaluateOneRepo(
 		dependabotAlerts,
@@ -42,6 +44,7 @@ function evaluateRepoTestHelper(
 		latestSnykIssues,
 		snykProjects,
 		reposOnSnyk,
+		workflowsForRepo,
 	).repocopRules;
 }
 
@@ -494,6 +497,7 @@ describe('REPOSITORY_09 - Dependency tracking', () => {
 			repo,
 			[snykSupportedLanguages],
 			['guardian/some-repo'],
+			[],
 		);
 		expect(actual).toEqual(true);
 	});
@@ -503,16 +507,27 @@ describe('REPOSITORY_09 - Dependency tracking', () => {
 			topics: ['production'],
 			full_name: 'guardian/some-repo',
 		};
-		const actual = hasDependencyTracking(repo, [fullySupportedLanguages], []);
+		const actual = hasDependencyTracking(
+			repo,
+			[fullySupportedLanguages],
+			[],
+			[],
+		);
 		expect(actual).toEqual(true);
 	});
+	//TODO: update this test to include dep graph integrator
 	test('is not valid if a project is not on snyk, and uses a language dependabot does not support', () => {
 		const repo: Repository = {
 			...nullRepo,
 			topics: ['production'],
 			full_name: 'guardian/some-repo',
 		};
-		const actual = hasDependencyTracking(repo, [snykSupportedLanguages], []);
+		const actual = hasDependencyTracking(
+			repo,
+			[snykSupportedLanguages],
+			[],
+			[],
+		);
 		expect(actual).toEqual(false);
 	});
 	test('is not valids not valid if a project is on snyk, and uses a language not supported by snyk', () => {
@@ -521,7 +536,7 @@ describe('REPOSITORY_09 - Dependency tracking', () => {
 			topics: ['production'],
 			full_name: 'guardian/some-repo',
 		};
-		const actual = hasDependencyTracking(repo, [unsupportedLanguages], []);
+		const actual = hasDependencyTracking(repo, [unsupportedLanguages], [], []);
 		expect(actual).toEqual(false);
 	});
 	test('is valid if a repository has been archived', () => {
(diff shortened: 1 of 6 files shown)


packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -47,6 +47,29 @@ export function doesRepoHaveWorkflow(
...
+	workflow_usages: guardian_github_actions_usage[],
+	language: DepGraphLanguage,
+): Repository[] {
+	let reposWithDepGraphLanguages: Repository[] = [];
+	const repos = productionRepos.filter((repo) =>
+		checkRepoForLanguage(repo, languages, language),
+	);
+	console.log(`Found ${repos.length} ${language} repos in production`);
+
+	reposWithDepGraphLanguages = reposWithDepGraphLanguages.concat(repos);
```
NovemberTang, line 62: if this variable is empty until this point, and isn't modified after this point, could we just instantiate it as an immutable const here instead?

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -47,6 +47,29 @@ export function doesRepoHaveWorkflow(
 	return false;
 }
 
+export function getDepGraphLanguageReposWithoutWorkflows(
```
NovemberTang, line 50: It's quite hard to tell what this function does at a glance. Could we rename it, or add a comment clarifying its behaviour?

packages/repocop/src/evaluation/repository.ts
```diff
@@ -143,34 +148,67 @@ export function hasDependencyTracking(
...
+		const allUnsupportedLanguagesSupportedByDepGraphIntegrator =
+			languagesNotSupportedByDependabot.every((language) =>
+				depGraphIntegratorSupportedLanguages.includes(language),
+			);
+
+		if (allUnsupportedLanguagesSupportedByDepGraphIntegrator) {
+			// Do all these languages have dependency submission workflows?
+			const allDepGraphSupportedLanguagesHaveWorkflows =
+				languagesNotSupportedByDependabot.every((language) =>
+					doesRepoHaveWorkflow(
```
NovemberTang, line 185: This function name is quite vague. Could we rename it to something more specific? Currently, it implies that it's checking a given repo to make sure it has any, arbitrary workflow file

packages/repocop/src/evaluation/repository.ts
```diff
@@ -143,34 +148,67 @@ export function hasDependencyTracking(
...
 				),
 			);
 		}
 		return containsOnlySnykSupportedLanguages;
+	} else if (containsOnlyDependabotSupportedLanguages) {
+		return true;
 	} else {
-		const containsOnlyDependabotSupportedLanguages = languages.every(
-			(language) => supportedDependabotLanguages.includes(language),
+		//not covered by Snyk or native Dependabot
```
NovemberTang, line 171: There's a lot of nesting here. I count at least 8 possible branches. I think this makes the code complicated to follow at times. I wonder if this could be reduced to a [switch statement](https://www.tutorialsteacher.com/typescript/typescript-switch), with one or two lines inside each case, abstracting the complexity away as much as possible

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
tjsilver: You're right, not sure why I pulled this function out - probably thought I would re-use it but didn't! I've reverted this now.

packages/repocop/src/evaluation/repository.test.ts
```diff
@@ -503,25 +531,64 @@ describe('REPOSITORY_09 - Dependency tracking', () => {
...
+		const actual = hasDependencyTracking(
+			repo,
+			[fullySupportedLanguages],
+			[],
+			[],
+		);
 		expect(actual).toEqual(true);
 	});
-	test('is not valid if a project is not on snyk, and uses a language dependabot does not support', () => {
+	//TODO: update this test to include dep graph integrator
```
NovemberTang, line 542: Does this TODO still need to be here?

packages/repocop/src/evaluation/repository.ts
```diff
@@ -123,55 +127,108 @@ function isMaintained(repo: Repository): boolean {
...
-				`${repo.name} contains the following languages not supported by Snyk: `,
+				`${repo.name} is on Snyk, but contains the following languages not supported by Snyk: `,
 				languages.filter(
 					(language) => !supportedSnykLanguages.includes(language),
 				),
 			);
+			return false;
 		}
-		return containsOnlySnykSupportedLanguages;
+		return true;
```
NovemberTang, line 149: I think we could eliminiate an additional layer of nesting here by flipping `if (!containsOnlySnykSupportedLanguages)`.

I can't do multiline suggestions on sections of the code where deletions have happend but it would look something like this


```
		if (containsOnlySnykSupportedLanguages) {
			return true;
		} else {
			console.log(
				`${repo.name} is on Snyk, but contains the following languages not supported by Snyk: `,
				languages.filter(
					(language) => !supportedSnykLanguages.includes(language),
				),
			);
		}
```

which eliminates the need for the first `return false`. To remove an extra layer of nesting, we could also calculate `containsOnlySnykSupportedLanguages` up front, with the caveat that we'd be trading performance for readability, so I don't have a strong opinion about it

packages/repocop/src/evaluation/repository.test.ts
tjsilver: No! Good spot.

packages/repocop/src/evaluation/repository.ts
```diff
@@ -123,6 +127,82 @@ function isMaintained(repo: Repository): boolean {
...
+	);
+	if (repoIsOnSnyk && !containsOnlySnykSupportedLanguages) {
+		console.log(
+			`${repo.name} is on Snyk, but contains the following languages not supported by Snyk: `,
+			languages.filter(
+				(language) => !supportedSnykLanguages.includes(language),
+			),
+		);
+	}
+	return repoIsOnSnyk && containsOnlySnykSupportedLanguages;
```
NovemberTang, line 147: This is loads better!!

packages/repocop/src/evaluation/repository.ts
```diff
@@ -123,6 +127,82 @@ function isMaintained(repo: Repository): boolean {
...
+				);
+
+			if (!allDepGraphSupportedLanguagesHaveWorkflows) {
+				console.log(
+					`${repo.name} contains the following languages supported by Dependency Graph Integrator for Dependabot, but that don't currently have a dependency submission workflow on the repo `,
+					languagesNotSupportedByDependabot.filter((language) =>
+						depGraphIntegratorSupportedLanguages.includes(language),
+					),
+				);
+				return false;
```
NovemberTang, line 189: could we simplify/encapsulate this big else block? Or break it into multiple `else if` blocks?

NovemberTang (approved):
Looks great!

PR #1273, opened 2024-09-23
Finish paramaterisation of org name

NovemberTang:
## What does this change?

A follow up to #1271 , this finishes the work to paramaterise all relevant instances of the `guardian` github org across the service catalogue.

## Why?

For the same reasons as the previous PR. Magic values are bad! And this will make it easier for people outside the guardian to use and modify the service catalogue if they wish

## How has it been verified?

- [ ]  CI passes.
- [ ]  Local execution
- [ ]  CODE execution

### DIFF
--- a/packages/dependency-graph-integrator/src/repo-functions.ts
+++ b/packages/dependency-graph-integrator/src/repo-functions.ts
@@ -1,64 +1,18 @@
 import { randomBytes } from 'crypto';
+import { createPullRequest } from 'common/src/pull-requests';
 import type { Octokit } from 'octokit';
-import { composeCreatePullRequest } from 'octokit-plugin-create-pull-request';
 import { addPrToProject } from '../../common/src/projects-graphql';
 import type { PullRequest, PullRequestParameters, StatusCode } from './types';
-
-interface Change {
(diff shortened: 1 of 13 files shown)


packages/dependency-graph-integrator/src/repo-functions.ts
```diff
@@ -1,65 +1,19 @@
 import { randomBytes } from 'crypto';
+import { createPullRequest } from 'common/src/pull-requests';
 import type { Octokit } from 'octokit';
-import { composeCreatePullRequest } from 'octokit-plugin-create-pull-request';
 import { addPrToProject } from '../../common/src/projects-graphql';
 import type { PullRequest, PullRequestParameters, StatusCode } from './types';
 
-interface Change {
```
NovemberTang, line 7: This code is _identical_ to [the code](https://github.com/guardian/service-catalogue/blob/55cda083315c6b29ba6a85ae4196735536959156/packages/common/src/pull-requests.ts#L8) in `common`. I'm not sure why it was duplicated

tjsilver (approved):
👍

PR #1274, opened 2024-09-24
Write cloudbuster findings to cloudquery 

NovemberTang:
## What does this change?

- Move `SecurityHubFinding` to common. Add fields so it's applicable to cloudbuster and obligatron
- Create function to transform raw SecurityHubFinding to a slightly expanded version of [this table](https://metrics.gutools.co.uk/d/ddi3x35x70jy8d/fsbp-compliance?orgId=1&viewPanel=7) on the FSBP dashboard
- Create prisma migration so that we can write data to the new table

## Why?

- Hopefully the dashboard will load a little faster as it's just retreieving a table, and performing minimal transformations on it.
- Create a common data model for FSBP vulnerabilities that can be used by both Cloudbuster and the dashboard, reducing the chances that these sources will diverge over time.

## How has it been verified?

It's quite fiddly to back out of a migration on CODE, so I'll run the test there **after** this PR has been approved, but **before** it's merged.

- [x] Unit tested the new function
- [x] Tested the migration and execution locally
- [x] Tested on CODE


packages/cloudbuster/src/findings.ts
```diff
@@ -1,28 +1,56 @@
...
+			return guFinding;
+		});
+	return transformedFindings;
+}
+
 /**
  * Transforms a SQL row into a finding
  */
-export function transformFinding(finding: aws_securityhub_findings): Finding {
+export function transformFinding(finding: SecurityHubFinding): Finding {
```
NovemberTang, line 38: The `Finding` type, and this function. will be retired and refactored out in a future PR. I'm trying to keep the size of this one manageable.

akash1810:
Did we consider using the obligatron resources here? These were designed to be generic; if this isn't true, it would be good to understand why and refactor.

NovemberTang:
> Did we consider using the obligatron resources here? These were designed to be generic; if this isn't true, it would be good to understand why and refactor.


A couple of reasons.
1) Obligatron only collects information about things that are outside of SLA, aka older than a certain age. Cloudbuster's primary aim is to allow teams to address issues before they get to that age. Obligatron doesn't store that information
2) Obligatron's data aggregates control failures by resource, as each individual resource under a particular obligation is supposed to be unique. It maps from an arn to a list of control failures, `myEc2Instance -> EC2.9, EC2.2`, etc. Here, we want to keep the data as flat as possible, ie `myEc2Instance -> EC2.9, myEc2Instance -> EC2.2`, so that it's easily filterable by control id. The end result is that obligatron should read its data from cloudbuster, not the other way around.

packages/common/src/types.ts
```diff
@@ -106,3 +107,20 @@ export const SLAs: Record<Severity, number | undefined> = {
...
+	Tags: Record<string, string> | null;
+	Region: string;
+	Type: string;
+};
+
+export type SecurityHubFinding = Pick<
+	aws_securityhub_findings,
+	'first_observed_at' | 'aws_account_id' | 'aws_account_name' | 'title'
+> & {
+	remediation: { Recommendation: { Url: string } };
```
tjsilver, line 122: Why are these capitalised and camel cased?

NovemberTang: > Why are these capitalised and camel cased?

They need to exactly match the field names in cloudquery, which uses snake_case for top level fields and PascalCase after that. Otherwise, TS won't be able to match them.

akash1810:
> [ ] Tested on CODE

I think this will fail. The `cloudbuster` lambda is currently using the (incorrect) database user [`repocop`](https://github.com/guardian/service-catalogue/blob/55cda083315c6b29ba6a85ae4196735536959156/packages/cdk/lib/cloudbuster.ts#L72), which will have [read-only access](https://github.com/guardian/service-catalogue/blob/main/packages/common/prisma/migrations/20240709104714_user_permissions/migration.sql#L10) to this new table.

Should we address https://trello.com/c/eJ5L5xj0/2922-create-db-user-for-cloudbuster first?

tjsilver (approved):
👍 subject to @akash1810's comment

PR #1275, opened 2024-09-24
refactor: enable PR creation by dep graph integrator

tjsilver:
## What does this change?

- Re-enables the dependency graph integrator to raise PRs on repos and add the dependency submission workflow for the repo.

## Why?

We switched this off to investigate an issue reported for the sbt submission workflow, and we now have more understanding of that issue. The team has decided this is not a blocker to rolling out the the PR creation.

## How has it been verified?
- [x] tested locally

### DIFF
--- a/packages/repocop/src/config.ts
+++ b/packages/repocop/src/config.ts
@@ -107,7 +107,8 @@ export async function getConfig(): Promise<Config> {
 		snykIntegrationPREnabled:
 			process.env.SNYK_INTEGRATION_PR_ENABLED === 'true',
 		snykIntegratorTopic: getEnvOrThrow('SNYK_INTEGRATOR_INPUT_TOPIC_ARN'),
-		depGraphIntegrationPREnabled: process.env.DEP_GRAPH_INTEGRATION_PR_ENABLED === 'true',
+		depGraphIntegrationPREnabled:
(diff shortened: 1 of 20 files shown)


packages/repocop/src/config.ts
```diff
@@ -107,7 +107,8 @@ export async function getConfig(): Promise<Config> {
 		snykIntegrationPREnabled:
 			process.env.SNYK_INTEGRATION_PR_ENABLED === 'true',
 		snykIntegratorTopic: getEnvOrThrow('SNYK_INTEGRATOR_INPUT_TOPIC_ARN'),
-		depGraphIntegrationPREnabled: false,
+		depGraphIntegrationPREnabled:
+			process.env.DEP_GRAPH_INTEGRATION_PR_ENABLED === 'true',
```
NovemberTang, line 111: I don't think this will work on AWS as we don't set this env var at the lambda level. Could we delete the flag altogether?

tjsilver: Ah! So the lambda doesn't see the .env file?

NovemberTang: no, that's just for local execution. The lambda's env vars are defined in CDK

NovemberTang (approved):
looks good!

PR #1276, opened 2024-09-25
Delete custom Finding type

NovemberTang:
## What does this change?

Cloudbuster was using a second custom type for creating digest messages. To simplify the code, I've migrated it to use cloudbuster_fsbp_vulnerabilities type used by the CQ database, which already contains all the required fields

## Why?

Reduction in complexity

## How has it been verified?

The ultimate result of this is a no-op, as the messages come out exactly the same. CI passing is sufficient.


tjsilver (approved):
👍

PR #1278, opened 2024-09-30
Create a cloudbuster user

NovemberTang:
## What does this change?

As cloudbuster will now be writing to a table, it requires different permissions to repocop. In this PR, we create a user that can only read the table containing security hub findings. It will also need to read from `cloudbuster_fsbp_findings`, but that permission will be granted in #1274 , the same PR the creates the table

## How has it been verified?

- [ ] Deployed and run cloudbuster on CODE (will do after approval)
- [x] updated SQL CI test

### DIFF
--- a/packages/cloudbuster/src/config.ts
+++ b/packages/cloudbuster/src/config.ts
@@ -29,8 +29,7 @@ export async function getConfig(): Promise<Config> {
 
 	const databaseConfig = isDev
 		? await getDevDatabaseConfig()
-		: await getDatabaseConfig(stage, 'repocop'); //TODO create a new db user for cloudbuster before deploying.
-
+		: await getDatabaseConfig(stage, 'cloudbuster');
 	return {
 		stage,
 		databaseConnectionString: getDatabaseConnectionString(databaseConfig),
--- a/packages/common/prisma/migrations/20240930091139_cloudbuster_user/migration.sql
+++ b/packages/common/prisma/migrations/20240930091139_cloudbuster_user/migration.sql
@@ -8,7 +8,7 @@ $do$
         END IF;
 
         -- Allow cloudbuster to read from all tables and views in the public schema
-        GRANT USAGE ON SCHEMA public to cloudbuster;
+        GRANT SELECT ON ALL TABLES IN SCHEMA public TO cloudbuster;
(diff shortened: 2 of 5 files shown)


packages/common/prisma/migrations/20240930091139_cloudbuster_user/migration.sql
```diff
@@ -0,0 +1,22 @@
...
+$do$
+    BEGIN
+        -- Create the `cloudbuster` user if it doesn't exist
+        IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE  rolname = 'cloudbuster') THEN
+            CREATE USER cloudbuster WITH LOGIN;
+        END IF;
+
+        -- Allow cloudbuster to read AWS SecurityHub Findings
+        GRANT USAGE ON SCHEMA public to cloudbuster;
+        GRANT SELECT ON public.aws_securityhub_findings TO cloudbuster;
```
akash1810, line 13: This contrasts with https://github.com/guardian/service-catalogue/pull/1176 which moved to allowing all DB users read access to all DB tables. Should we consolidate the access patterns for consistency?

packages/common/prisma/migrations/20240930091139_cloudbuster_user/migration.sql
```diff
@@ -0,0 +1,22 @@
+-- This is an empty migration.
```
akash1810, line 1: I spy a lie...

packages/cloudbuster/src/config.ts
```diff
@@ -29,8 +29,7 @@ export async function getConfig(): Promise<Config> {
 
 	const databaseConfig = isDev
 		? await getDevDatabaseConfig()
-		: await getDatabaseConfig(stage, 'repocop'); //TODO create a new db user for cloudbuster before deploying.
-
+		: await getDatabaseConfig(stage, 'cloudbuster');
```
akash1810, line 32: I think the infra needs to change [here](https://github.com/guardian/service-catalogue/blob/d354f5a0f002d539293c27749e80b0a152779a49/packages/cdk/lib/cloudbuster.ts#L72) too.

akash1810 (approved):

PR #1279, opened 2024-10-01
Small cloudbuster fixes

NovemberTang:
## What does this change?

- Correctly encodes Grafana URL in Anghammarad CTA, skips CTA creation if account name is null
- Only shows FSBP violations from last 60 days
- Sends notifications to teams on PROD
- Fix local execution by loading local .env file


packages/cloudbuster/src/digests.ts
```diff
@@ -20,35 +20,56 @@ export function createDigestsFromFindings(
...
-
-	return {
-		accountId: aws_account_id,
-		accountName: aws_account_name as string,
-		actions,
-		subject: `Security Hub findings for AWS account ${aws_account_name}`,
-		message: createEmailBody(accountFindings),
-	};
+			accountName: aws_account_name,
+			actions: createCta(aws_account_name),
```
adamnfish, line 59: It looks like we're already sure the aws_account_name exists, based on the check on line 54. Do we really need `createCta` to take `string | null`?

packages/cloudbuster/src/digests.ts
```diff
@@ -24,35 +24,56 @@ export function createDigestsFromFindings(
...
 
-function createDigestForAccount(
+function createCta(aws_account_name: string): Action[] {
+	if (!aws_account_name) {
+		return [];
+	} else {
+		return [
+			{
+				cta: `View all findings on Grafana`,
+				url: `https://metrics.gutools.co.uk/d/ddi3x35x70jy8d?var-account_name=${encodeURIComponent(aws_account_name)}`,
```
tjsilver, line 34: It might be nice to pass in the dashboard name as a param to make it easier to parse as a human

NovemberTang: I'm not sure I understand what you mean by dashboard name?

tjsilver (approved):
👍

PR #1290, opened 2024-10-02
Eliminate race condition

NovemberTang:
## What does this change?

Now that cloudbuster writes to a DB, to avoid race conditions, we should not be running two instances simultaneously, as this could create race conditions. Instead, we can follow repocop's model of running every day, with high severities being evaluated once a week

## How has it been verified?

- Added unit tests to verify behaviour
- CI passes


tjsilver (approved):
Looks good!

PR #1295, opened 2024-10-08
add a weekend offset for vulnerabilities raised on Fri/Sat/Sun

NovemberTang:
## What does this change?

Give developers extra time to resolve critical vulnerbilities raised on or close to the weekend

## Why?

Provide a more realistic timeline for remediation. Currently, vulnerabilities raised on a friday afternoon (ie after repocop has run that day) are already out of date by the time monday rolls around

## How has it been verified?

Unit tests have been added to verify the status of a vuln raised over the weekend on the following Monday

## Next steps

~~As a follow up PR, we should include an `is_within_sla` field in the repocop table, so that the dashboard has the same information as repocop's vulnerability digest~~ resolved in #1296

### DIFF
--- a/packages/common/src/functions.test.ts
+++ b/packages/common/src/functions.test.ts
@@ -229,6 +229,35 @@ describe('daysLeftToFix', () => {
 	test('should return 2 if a critical vuln was raised today', () => {
 		expect(daysLeftToFix(new Date(), 'critical')).toBe(2);
 	});
+
+	const monday = new Date('2024-10-07'); // Oct 7th 2024 is a Monday
+	beforeEach(() => {
+		jest.useFakeTimers().setSystemTime(monday);
+	});
+
+	afterEach(() => {
+		jest.useRealTimers();
+	});
+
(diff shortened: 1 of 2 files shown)


akash1810 (commented):
Does this mean on Thursday the message will say "you have 0 days to remediate", then the next day (Friday) it will say "you have 2 days to remediate"? I think this confuses expectations, more than it simplifies it.

NovemberTang:
> Does this mean on Thursday the message will say "you have 0 days to remediate", then the next day (Friday) it will say "you have 2 days to remediate"? I think this confuses expectations, more than it simplifies it.

No, we're just calculating an offset to add to the number of days teams already had to fix the vulnerability, so the calculation is 2 days + the weekend offset. If a vulnerability was raised on a Thursday, they would have 2 days (2+0) to fix it. If it was raised on a Friday or Saturday, they would have 4 days (2+2), and if it was raised on Sunday, they would have 3 (2+1).

tjsilver (approved):
LGTM

PR #1297, opened 2024-10-09
Allow cloudbuster to run, without sending messages

NovemberTang:
## What does this change?

- Start collecting all cloudbuster data, not just the first five rows
- Convert cloudbuster to a scheduled lambda cloudbuster, allow it to start running on AWS
- Switch off all messaging temporarily
- Cloudbuster's schedule and monitoring configuration echoes repocop's

## Why?

Cloudbuster is ready to be rolled out after the comms announcement. In advance of this, we need to start migrating the grafana tables. To migrate the grafana tables, we need a complete set of data in the Cloudquery DB. So let's allow cloudbuster to collect this

I've switched off messaging as we don't want to start alerting teams before the comms are sent out. We can reconfigure this next week

## How has it been verified?

CI passes, snapshots look normal.
Works on CODE

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -49,6 +49,35 @@ import { addPrismaMigrateTask } from './prisma-migrate-task';
 import { addRefreshMaterializedViewLambda } from './refresh-materialized-view';
 import { Repocop } from './repocop';
 
+function createProdMonitoringConfiguration(
+	app: string,
+): GuLambdaErrorPercentageMonitoringProps {
+	return {
+		toleratedErrorPercentage: 50,
+		lengthOfEvaluationPeriod: Duration.minutes(1),
+		numberOfEvaluationPeriodsAboveThresholdBeforeAlarm: 1,
+		snsTopicName: 'devx-alerts',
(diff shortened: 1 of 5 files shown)


packages/cdk/lib/service-catalogue.ts
```diff
@@ -49,6 +49,35 @@ import { addPrismaMigrateTask } from './prisma-migrate-task';
...
 
+function createProdMonitoringConfiguration(
+	app: string,
+): GuLambdaErrorPercentageMonitoringProps {
+	return {
+		toleratedErrorPercentage: 50,
+		lengthOfEvaluationPeriod: Duration.minutes(1),
+		numberOfEvaluationPeriodsAboveThresholdBeforeAlarm: 1,
+		snsTopicName: 'devx-alerts',
+		alarmDescription: `RepoCop error percentage is too high. Find the logs here ${getCentralElkLink(
```
NovemberTang, line 60:
```suggestion
		alarmDescription: `${app} error percentage is too high. Find the logs here ${getCentralElkLink(
```

tjsilver (approved):
This looks good, we went through it together and it makes sense!

PR #1298, opened 2024-10-09
stop logging the full cloudbuster findings table

NovemberTang:
## What does this change?

Previously, we were logging the entire table out, in a tabular format. In ELK this is very ugly, and the message also cuts off early because it's so large. Now, we just capture the total number of vulnerabilities so we can sense check the count while debugging


tjsilver (approved):
👍

PR #1300, opened 2024-10-14
optimise cloudbuster config

NovemberTang:
## What does this change?

Cloudbuster only uses 200-300MB when running, so we can reduce the amount of memory it has. It also only takes about 3-4 seconds to run, so I think we can also reduce its timeout

## Why?

Waste not want not.

## How has it been verified?

CI passes


tjsilver (approved):
Great!

PR #1301, opened 2024-10-15
fix: update sbt dependency submission workflow to install sbt

tjsilver:
## What does this change?

Adds a step to the dependency graph integrator PR generator to install sbt.

## Why?

The workflow currently fails because sbt is no longer bundled with latest Ubuntu.

## How has it been verified?

Tested by copying the output of running locally in `janus-app` and observed that the workflow [succeeded](https://github.com/guardian/janus-app/actions/runs/11349830699).


NovemberTang (approved):
Nice one!

PR #1302, opened 2024-10-16
Obligatron filters on within_sla

NovemberTang:
## What does this change?

Repocop now contains a `within_sla` column, which used `daysLeftToFix()` under the hood. Now, instead of performing the calculation twice, we just work out whether or not something is within SLA by using the column repocop provides

## Why?

Efficiency


adamnfish (commented):
Is it not better to calculate this in real time when the dashbaord is displayed? This is data that meaningfully changes after this data is generated.

adamnfish (approved):

PR #1303, opened 2024-10-17
feat: explicitly install java in sbt dep submission workflow

tjsilver:
## What does this change?

- Explicitly installs java in the sbt dependency submission workflow created by the dependency graph integrator
- Updates `actions/checkout` to latest
- Updates tests

## Why?

We can't guarantee that future versions of Ubuntu will ship with Java, and it's required by sbt in this workflow.

## How has it been verified?

- [x] Run tests locally
- [x] Test PR created on another repo to ensure it runs -[ tested on `security-hq`](https://github.com/guardian/security-hq/actions/runs/11388986269)

### DIFF
--- a/packages/dependency-graph-integrator/src/file-generator.test.ts
+++ b/packages/dependency-graph-integrator/src/file-generator.test.ts
@@ -17,9 +17,15 @@ jobs:
     steps:
       - name: Checkout branch
         id: checkout
-        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+      - name: Install Java
+        id: java
+        uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4.2.0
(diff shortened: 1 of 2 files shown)


packages/dependency-graph-integrator/src/file-generator.test.ts
```diff
@@ -18,6 +18,15 @@ jobs:
       - name: Checkout branch
         id: checkout
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Install Java
+        id: java
+        uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4.2.0
+        with:
+          distribution: corretto
+          java-version: 17
```
NovemberTang, line 26: Is the idea that teams can adjust the PR to suit their needs? Or that java 17 will always work with previous versions ie 11? If the former, do we need to add a troubleshooting note to the PR body?

NovemberTang: We've tested this and confirmed java17 is compatible with scala versions as early as 2.12.x, and sbt 1.5.x

PR #1304, opened 2024-10-17
enable cloudbuster messaging on CODE/PROD

NovemberTang:
## What does this change?

Allow cloudbuster to contact teams about new FSBP vulnerabilities

## Why?

So they can act on them in a timely manner

### DIFF
--- a/packages/cloudbuster/src/config.ts
+++ b/packages/cloudbuster/src/config.ts
@@ -34,7 +34,7 @@ export async function getConfig(): Promise<Config> {
 		stage,
 		databaseConnectionString: getDatabaseConnectionString(databaseConfig),
 		withQueryLogging: isDev,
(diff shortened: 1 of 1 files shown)


packages/cloudbuster/src/config.ts
```diff
@@ -34,7 +34,7 @@ export async function getConfig(): Promise<Config> {
 		stage,
 		databaseConnectionString: getDatabaseConnectionString(databaseConfig),
 		withQueryLogging: isDev,
-		enableMessaging: false, //!isDev,
+		enableMessaging: !isDev,
```
NovemberTang, line 37: on PROD, it sends messages to the teams, on CODE, it sends messages to the anghammarad test channel

tjsilver (approved):
👍

PR #1305, opened 2024-10-22
Request admin reviews on PRs raised by the service catalogue

NovemberTang:
### NB: there are three quite distinct code changes in this PR. As such, it's probably less complicated to review this commit-by-commit

- Adds a function that requests repository admins as reviewers on any PR raised using `createPrAndAddToProject`
- Simplifies findContactableOwners, as since #831 , `view_repo_ownership` has included the team slug, meaning we don't need to cross-reference with `github_teams` to retrieve it
- Deletes one instance of the previously duplicated `createPrAndAddToProject`

> [!WARNING]  
> Requires a GitHub permission change. In addition to `pull-request:write`, we need `members:read` to validate the team name

## Why?

Engineers are sometimes slow to respond to PRs. By requesting reviews, we hope to speed up the process of having DevX-raised PRs merged. If no significant change is observed, we can try another method

## How has it been verified?

- [x] Added relevant unit tests to make sure correct admins are being passed through
- [x] Updated relevant github app permission, and tested on PROD using [devx-docs](https://github.com/guardian/devx-docs/pull/29)

### DIFF
--- a/packages/common/src/pull-requests.ts
+++ b/packages/common/src/pull-requests.ts
@@ -135,18 +135,19 @@ export async function createPrAndAddToProject(
 	commitMessage: string,
 	boardNumber: number,
 	admins: string[],
+	octokit?: Octokit,
 ) {
 	if (stage === 'PROD') {
-		const octokit = await stageAwareOctokit(stage);
+		const ghClient = octokit ?? (await stageAwareOctokit(stage));
 		const existingPullRequest = await getExistingPullRequest(
-			octokit,
+			ghClient,
 			repoName,
 			owner,
 			`${author}[bot]`,
 		);
 
 		if (!existingPullRequest) {
-			const pullRequestResponse = await createPullRequest(octokit, {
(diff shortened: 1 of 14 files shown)


packages/common/src/pull-requests.ts
```diff
@@ -101,18 +134,20 @@ export async function createPrAndAddToProject(
 	fileContents: string,
 	commitMessage: string,
 	boardNumber: number,
+	admins: string[],
+	octokit?: Octokit,
 ) {
 	if (stage === 'PROD') {
-		const octokit = await stageAwareOctokit(stage);
```
NovemberTang, line 106: In the other version of this function, we were passing in octokit. As a compromise, I've made it an optional parameter, where it will create a new instance if one isn't passed in. This can be cleaned up when we fully deprecate the snyk integrator

tjsilver (approved):
Great work. Is it worth making a note that the `createPrAndAddToProject` function should be moved into dep graph integrator when snyk integrator is stripped out?

PR #1306, opened 2024-10-23
remove s3 checks from the interactive monitor

NovemberTang:
## What does this change?

Previously, the interactive monitor was checking buckets in S3 to make sure that interactives had been published to S3 before labelling them. This is overly restrictive, so we're now just checking for the presence of certain config files, so we can catch interactives while they're still in development.

## Why?

To reduce the amount of time it takes to catch an interactive, and increase the number of repos with a correct topic

## How has it been verified?

- [x] CI passes
- [x] Run on CODE

### DIFF
--- a/packages/interactive-monitor/src/index.ts
+++ b/packages/interactive-monitor/src/index.ts
@@ -1,7 +1,4 @@
-import type { _Object, ListObjectsCommandInput } from '@aws-sdk/client-s3';
-import { ListObjectsCommand, S3Client } from '@aws-sdk/client-s3';
 import type { SNSHandler } from 'aws-lambda';
-import { awsClientConfig } from 'common/aws';
 import { applyTopics, parseEvent, stageAwareOctokit } from 'common/functions';
 import type { Octokit } from 'octokit';
 import type { Config } from './config';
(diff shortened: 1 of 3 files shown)


packages/interactive-monitor/src/index.ts
```diff
@@ -48,74 +45,42 @@ async function getConfigJsonFromGithub(
...
-		Bucket: 'gdn-cdn',
-		Prefix: `${prefix}`,
-		MaxKeys: 1,
-	};
-	const command = new ListObjectsCommand(input);
-	const response = await s3.send(command);
-	return response.Contents;
-}
-
 async function s3PathIsInConfig(
```
kelvin-chappell, line 48: Do you think this function name is now misleading?

NovemberTang: Not necessarily, it's not checking to make sure that a json config file has a particular shape, namely, that it contains a `path` key, that points to a location in S3 where the interactive will be deployed. I've just removed the bit that makes sure that that path actually exists in S3. Potentially it could be better documented though!

kelvin-chappell (approved):
👍

PR #1307, opened 2024-10-23
send a configurable number of repos to the dependency graph integrator

NovemberTang:
## What does this change?

Sends a configurable number of repos to the dependency graph integrator

## Why?

So we can control the pace of integration.

## How has it been verified?

- [x] CI passes
- [x] Tested on CODE

### DIFF
--- a/packages/repocop/src/index.ts
+++ b/packages/repocop/src/index.ts
@@ -29,7 +29,7 @@ import {
 	getTeams,
 } from './query';
 import { protectBranches } from './remediation/branch-protector/branch-protection';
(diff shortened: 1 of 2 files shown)


packages/repocop/src/index.ts
```diff
@@ -166,6 +165,15 @@ export async function main() {
 		await sendUnprotectedRepo(repocopRules, config, repoLanguages);
 	}
 
+	await sendReposToDependencyGraphIntegrator(
+		config,
+		repoLanguages,
+		productionRepos,
+		productionWorkflowUsages,
+		repoOwners,
+		2,
```
tjsilver, line 174: Shall we make this number a variable like `numReposToSend`?

tjsilver (approved):
Great!

PR #1309, opened 2024-10-23
make CI depend on db-migration task

NovemberTang:
## What does this change?

Make CI depend on db-migration

## Why?

We shouldn't be able to deploy to riff-raff unless the migrations are tidy

## How has it been verified?

CI passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,68 +56,6 @@ jobs:
           script: typecheck
           deployment_key: ${{ secrets.PRIVATE_INFRASTRUCTURE_CONFIG_DEPLOY_KEY }}
 
-  CI:
-    timeout-minutes: 15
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -181,3 +119,65 @@ jobs:
             git ls-files --others --exclude-standard packages/common/prisma
             exit 1
           fi
+
+  CI:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    needs: [lint, test, typecheck, db-migration]
```
NovemberTang, line 126: This is the only line that's actually changed. Otherwise, I've just swapped the ordering of the CI job and the db-migration job so they make more logical sense

kelvin-chappell (approved):
👍

PR #1310, opened 2024-10-24
pass in the github org name via config

NovemberTang:
## What does this change?

Allows the github org name in dependency graph integrator to be configurable, similarly to #1271 

## Why?

Magic values are bad! And this will make it easier for people outside the guardian to use and modify the service catalogue if they wish.

## How has it been verified?

- [x]  CI passes.
- [x]  CODE execution


tjsilver (approved):
Tested this on CODE and it worked as expected.

PR #1312, opened 2024-10-29
Group Cloudbuster findings in digest messages

NovemberTang:
## What does this change?

Group FSBP findings by Control ID, and app. Users can click through on the findings link to see a prefiltered FSBP dashboard.

## Why?

Any given account could have an arbitrary number of FSBP failures. This means that digests could be (and have been) hundreds of lines long, making them unreadable. Let's start tackling this by grouping control failures by control ID and app, to make them a bit more readable at a glance, which is the point of a digest.

### Why group only by app?

I'm still figuring out the balance between granularity and conciseness. If you group by too may variables, you _can_ end up with almost as many entries as if you'd just listed them all individually. Too few, and you hide information that's useful. I've chosen to group by control ID and app for now.

**Why not Stack?** Most of the time there's only one riff-raff stack in any given account, and its vanishingly rare to have more than three. I think grouping by stack would only really be useful if there were two services running in the same account with the same app tag, 
deployed via different stacks, which (a) is an unlikely edge case, and (b) can be resolved by clicking through to the dashboard and looking at the other tags

**Why not Stage?** Generally, infrastructure configuration doesn't vary massively across stage. This means that in nearly every case, over 90% of the time, disambiguating resources by stage is not useful, so I've not implemented this grouping as part of this PR.

## How has it been verified?

Added and modified unit tests to verify new behaviour

Running on CODE produces messages like these. In this example, we have aggregated nineteen separate issues into three groups:

Screenshot 2024-10-29 at 13 28 13

## Future work

This work is intended as a starting point to be iterated on, rather than a finished product, due to time constraints, and a desire to keep the complexity of this PR manageable. As a piece of future work, we should add an `app` variable to the list of filters on the dashboard (maybe stack/stage/repo too), to further help teams zoom in on their outstanding issues, and then wire that back into the URL we generate in the digest. It might be worth reviewing in a few weeks to see if more granular grouping would be useful to teams, or make the digest long enough that people stop reading it again.

It would also be useful to introduce another layer of grouping, to reduce repetition. The current data structure for grouping FSBP results looks like this:

```mermaid
flowchart LR

ControlId-App1
ControlId-App2
Vulnerabilities1
Vulnerabilities2

ControlId-App1 --> Vulnerabilities1
ControlId-App2 --> Vulnerabilities2

```

But it might make more sense for our data, and therefore digests, could be structured more like this, to avoid repeating fields like the full control title in every line

```mermaid
flowchart LR

ControlId
App1
App2
Vulnerabilities1
Vulnerabilities2

ControlId --> App1 --> Vulnerabilities1
ControlId --> App2 --> Vulnerabilities2

```

### DIFF
--- a/packages/cloudbuster/src/digests.ts
+++ b/packages/cloudbuster/src/digests.ts
@@ -96,14 +96,14 @@ function formatFindings(
 	account_name: string,
 	findings: cloudbuster_fsbp_vulnerabilities[],
 ) {
-	console.log(`formatting findings for ${key}`);
 	const findingsCount = findings.length;
 	const control_id = findings[0]?.control_id;
-	const app = findings[0]?.app ?? 'unknown-app';
+	const app = findings[0]?.app ?? 'unknown';
 	const remediation = findings[0]?.remediation;
 	const title = findings[0]?.title;
+	const findingsString = findingsCount === 1 ? 'finding' : 'findings';
 	const url = `https://metrics.gutools.co.uk/d/ddi3x35x70jy8d?var-account_name=${encodeURIComponent(account_name)}&var-control_id=${control_id}`;
-	return `[${findingsCount} findings](${url}) for control [${control_id}](${remediation}), (${title}) in app ${app}`;
+	return `[${findingsCount} ${findingsString}](${url}) in app: **${app}**, for control [${control_id}](${remediation}), (${title})`;
 }
 
 function createEmailBody(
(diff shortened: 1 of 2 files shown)


packages/cloudbuster/src/digests.ts
```diff
@@ -62,26 +63,69 @@ export function createDigestForAccount(
...
+	key: string,
+	account_name: string,
+	findings: cloudbuster_fsbp_vulnerabilities[],
+) {
+	const findingsCount = findings.length;
+	const control_id = findings[0]?.control_id;
+	const app = findings[0]?.app ?? 'unknown';
+	const remediation = findings[0]?.remediation;
+	const title = findings[0]?.title;
+	const findingsString = findingsCount === 1 ? 'finding' : 'findings';
```
kelvin-chappell, line 104: Nice touch!

kelvin-chappell (approved):
Seems like a sensible grouping! 👍

PR #1313, opened 2024-10-29
docs: fix dependency graph integrator running instructions

NovemberTang:
## What does this change?

Adds `admins` field to running instructions

## Why?

This is a required input field


kelvin-chappell (approved):
👍

PR #1314, opened 2024-10-30
update github languages plugin to 0.0.5

NovemberTang:
## What does this change?

updates github languages plugin to 0.0.5

## Why?

take advantage of some security fixes

## How has it been verified?

- [x] Run the github languages task on CODE


tjsilver (approved):
👍

PR #1315, opened 2024-10-30
Fix workflow check in repocop dep graph integration

tjsilver:
## What does this change?

- Fixes the function (originally `createSnsEventsForDependencyGraphIntegration` ) that detected whether a repo has a dependency submission workflow. The function was erroneously reporting that no workflows were found so all repos were candidates for being selected and sent to dependency graph integrator for PRs to be raised.
- Fixes a bug in the filtering of workflows in the dependency tracking functionality (`hasDependencyTracking` which was still reporting correctly) which uses the same function as above (namely `doesRepoHaveDepSubmissionWorkflowForLanguage`).
- Refactors the sns messaging function `createSnsEventsForDependencyGraphIntegration`  to separate out the workflow check. The function was doing a lot of different things in addition to creating the events. I have pulled out the remaining functionality into a new function called `getReposWithoutWorkflows`, and created a new `RepositoryWithDepGraphLanguage` type to allow its output to be fed directly into `createSnsEventsForDependencyGraphIntegration`.
 - Refactors the `sendReposToDependencyGraphIntegrator` function to select suitable repos using the new `getReposWithoutWorkflows` function and shuffle/slice the required number of repos before sending them for message creation.
- Fixes some linting errors.

## Why?

To make the dependency graph integrator work correctly, and hopefully make the code clearer, with better separation of concerns.

## How has it been verified?

Tested by running repocop locally and observing the logs.

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -97,6 +97,10 @@ export interface Repository extends RepositoryFields {
 	id: NonNullable<RepositoryFields['id']>;
 }
 
+export interface RepositoryWithDepGraphLanguage extends Repository {
+	language: DepGraphLanguage;
+}
+
 // The number of days teams have to fix vulnerabilities of a given severity
 export const SLAs: Record<Severity, number | undefined> = {
 	critical: 2,
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -98,7 +98,7 @@ export interface Repository extends RepositoryFields {
 }
 
 export interface RepositoryWithDepGraphLanguage extends Repository {
-	language: DepGraphLanguage;
+	dependency_graph_language: DepGraphLanguage;
 }
 
 // The number of days teams have to fix vulnerabilities of a given severity
--- a/packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
+++ b/packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
@@ -145,9 +145,14 @@ export async function sendReposToDependencyGraphIntegrator(
 		);
 
 	if (suitableRepos.length !== 0) {
-		const eventsToSend: DependencyGraphIntegratorEvent[] = shuffle(
-			createSnsEventsForDependencyGraphIntegration(suitableRepos, repoOwners),
-		).slice(0, repoCount);
+		console.log(
+			`Found ${suitableRepos.length} suitable repos without dependency submission workflows`,
+		);
+
+		const selectedRepos = shuffle(suitableRepos).slice(0, repoCount);
+
+		const eventsToSend: DependencyGraphIntegratorEvent[] =
+			createSnsEventsForDependencyGraphIntegration(selectedRepos, repoOwners);
 
 		for (const event of eventsToSend) {
 			await sendOneRepoToDepGraphIntegrator(config, event);
--- a/packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
+++ b/packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
@@ -10,6 +10,7 @@ import type {
 	DependencyGraphIntegratorEvent,
 	DepGraphLanguage,
 	Repository,
+	RepositoryWithDepGraphLanguage,
 } from 'common/src/types';
 import type { Config } from '../../config';
 import { findContactableOwners, removeRepoOwner } from '../shared-utilities';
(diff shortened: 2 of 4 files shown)


packages/common/src/types.ts
```diff
@@ -97,6 +97,10 @@ export interface Repository extends RepositoryFields {
 	id: NonNullable<RepositoryFields['id']>;
 }
 
+export interface RepositoryWithDepGraphLanguage extends Repository {
+	language: DepGraphLanguage;
```
NovemberTang, line 101: the `github_repositories` type that `Repository` is based on already has a `language` field, used to hold the primary language of the repository. Introducing another field with the same name but a different purpose could cause confusion, or potentially errors down the line. Could we rename this to be more specific, like `dependency_graph_language`?

non-blocking edge case: what if we need two dependency graph integrator languages?

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -118,16 +137,22 @@ export async function sendReposToDependencyGraphIntegrator(
 	repoOwners: view_repo_ownership[],
 	repoCount: number,
 ): Promise<void> {
-	const eventsToSend: DependencyGraphIntegratorEvent[] = shuffle(
-		createSnsEventsForDependencyGraphIntegration(
+	const suitableRepos: RepositoryWithDepGraphLanguage[] =
```
NovemberTang, line 140: minor: Could we rename this to indicate what makes it suitable? i.e. `reposNeedingDepGraphIntegration`

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -118,16 +137,22 @@ export async function sendReposToDependencyGraphIntegrator(
...
 			productionWorkflowUsages,
-			repoOwners,
-		),
-	).slice(0, repoCount);
+		);
+
+	if (suitableRepos.length !== 0) {
+		const eventsToSend: DependencyGraphIntegratorEvent[] = shuffle(
+			createSnsEventsForDependencyGraphIntegration(suitableRepos, repoOwners),
+		).slice(0, repoCount);
```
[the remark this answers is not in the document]
tjsilver: I think this comment might have been made on an earlier commit, as I have addressed this.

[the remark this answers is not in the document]
NovemberTang: ah, you're right! i was walking through commit by commit! Apologies!

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -110,6 +85,50 @@ async function sendOneRepoToDepGraphIntegrator(
 	}
 }
 
+export function getReposWithoutWorkflows(
```
akash1810, line 88: I find the implementation of this function a little hard to follow. I wonder if we could refactor to be more functional? (I'll attempt and reply here with a suggestion).

akash1810: Here's an attempted refactor:

```ts
export function getReposWithoutWorkflows(
	languages: github_languages[],
	productionRepos: Repository[],
	productionWorkflowUsages: guardian_github_actions_usage[],
): RepositoryWithDepGraphLanguage[] {
	const depGraphLanguages: DepGraphLanguage[] = ['Scala', 'Kotlin'];

	const allReposWithoutWorkflows: RepositoryWithDepGraphLanguage[] =
		depGraphLanguages.flatMap((language) => {
			const reposWithDepGraphLanguages: Repository[] = productionRepos.filter(
				(repo) => checkRepoForLanguage(repo, languages, language),
			);

			console.log(
				`Found ${reposWithDepGraphLanguages.length} ${language} repos in production`,
			);

			return reposWithDepGraphLanguages
				.filter((repo) => {
					const workflowUsagesForRepo = productionWorkflowUsages.filter(
						(workflow) => workflow.full_name === repo.full_name,
					);
					return !doesRepoHaveDepSubmissionWorkflowForLanguage(
						repo,
						workflowUsagesForRepo,
						language,
					);
				})
				.map((repo) => ({ ...repo, dependency_graph_language: language }));
		});

	console.log(
		`Found ${allReposWithoutWorkflows.length} production repos without dependency submission workflows`,
	);
	return allReposWithoutWorkflows;
}
```

The main changes are a removal of intermediate variables (declared with `let`), in place of `.flatMap`/`.map`.

However, I wonder if we could simplify this even further. We're currently working with 3 data sets:
- `github_languages[]`
- `Repository[]`
- `guardian_github_actions_usage[]`

We're joining these datasets in this function, and I'd suspect in other locations. Given we know how the three datasets relate, could we join them together once, very early? That is, could we go from passing three datasets around, to one that looks something like this:

```ts
interface HydratedRepository {
  full_name: string;
  languages: string[];
  workflows: string[]
}
```

Note: the interface `HydratedRepository` above is likely missing some properties!

The main benefit here is a reduction in the number of `.filter(foo => foo.full_name === bar.full_name)` calls across the repository - I think it'll be reduced to 1!

This'll be a bigger refactor, so not proposing to make it in this PR, but I think it's worth investigating.

tjsilver: > I find the implementation of this function a little hard to follow. Could we add a unit test for it?
> 
> I wonder if we could refactor to be more functional too? (I'll attempt and reply here with a suggestion).

There are four tests for this function  in `send-to-sns.test.ts`, are there others you think would be useful?

tjsilver: Thank you, I have refactored the function according to your suggestion. I agree, there is definitely room for a larger simplification/refactor here to reduce duplication, which I will ticket up for a future PR.

NovemberTang (approved):

PR #1316, opened 2024-10-31
Update schema to reflect current CQ versions

NovemberTang:
## What does this change?

Cloudquery's GitHub plugin v11 has a custom properties table. We would like to make use of it. Before upgrading to v11, let's make sure our current schema is correct. The schema of some AWS tables was updated in a previous version bump, but the schema had never been updated, and as these fields are never null in reality, we haven't seen an error yet. Regardless, we should be using the correct types.

## Why?

I'm doing this work in advance to keep the size of any single migraition manageable

## How has it been verified?

CI passes


akash1810:
> The schema of some AWS tables was updated in a previous version bump, but the schema had never been updated

We updated the AWS plugin version in #1153. The file changes in that PR suggest the schema was changed. Did we miss some? In which case are our docs accurate? Or, are you doing something different in your local env?

NovemberTang:
> > The schema of some AWS tables was updated in a previous version bump, but the schema had never been updated
> 
> We updated the AWS plugin version in #1153. The file changes in that PR suggest the schema was changed. Did we miss some? In which case are our docs accurate? Or, are you doing something different in your local env?

Yeah, I think some of the tables were missed in the initial migration. This PR is just catching up with that

kelvin-chappell (approved):
👍

PR #1329, opened 2024-11-13
refactor: increase dependency submission PR creation to 5 per day

tjsilver:
## What does this change?

Increases the number of PRs raised by the dependency graph integrator from 2 to 5 per (working) day. 

## Why?

To ensure that we can get all the repos covered before Snyk is deprecated. This will also allow us time to remind teams to merge the PR if they haven't already.

PR #1330, opened 2024-11-14
feat: remove snyk integrator

tjsilver:
## What does this change?

Strips out the Snyk integrator which raised PRs to add a workflow to repos. 

## Why?

We are deprecating Snyk. This will allow us to simplify the logic in repocop also.

## How has it been verified?
- [x] Tested locally
- [ ] Tested on CODE (n.b. requires a dangerous deploy as we are deleting infrastructure)


README.md
```diff
@@ -82,10 +82,10 @@ flowchart TB
     dev[P&E Developer]
     repocop[Repocop Lambdas]
     aws[AWS APIs]
-    
+
     snyk --> |Data from snyk populates Cloudquery tables|cq
-    github --> |Data from snyk populates Cloudquery tables|cq
-    aws --> |Data from snyk populates Cloudquery tables|cq
+    github --> |Data from dependabot populates Cloudquery tables|cq
```
adamnfish, line 87: 👍

packages/repocop/src/index.ts
```diff
@@ -163,7 +163,7 @@ export async function main() {
 
 	await sendRandomRepoToSnykIntegrator(repocopRules, config, repoLanguages);
 
-	const dependencyGraphIntegratorRepoCount = 2;
+	const dependencyGraphIntegratorRepoCount = 5;
```
adamnfish, line 166: Does this mean we're also ramping up the rollout pace, here?

adamnfish (approved):
LGTM, thank you!

packages/repocop/src/index.ts
tjsilver: Ooh that's for another PR - will remove from this one!

adamnfish: Maybe this was meant to be just in the other PR?

https://github.com/guardian/service-catalogue/pull/1329#pullrequestreview-2435554323

tjsilver: Yep, I've changed the base now and that's fixed it.

NovemberTang (commented):
🫶

PR #1333, opened 2024-11-22
Make packages baked into images by Amigo visible to the service catalogue

kelvin-chappell:
## Purpose

This PR makes a new table available to the service catalogue, `amigo_bake_packages`.  This table will tell us at a glance what packages are installed on which Amigo bakes, and in conjunction with other tables in the catalogue will tell us what packages are installed on which EC2 instances and in which AWS accounts.

## Implementation

We add a new cloudformed task definition to ECS to run a CloudQuery job.  The CQ job fetches data from the Amigo Dynamo DB tables and Amigo packagelist S3 objects and stores them in the new service catalogue table on a daily schedule.

The CQ job makes use of our new custom [image package source plugin](https://github.com/guardian/cq-source-image-packages).

## Testing
Tested on Code environment where the new task has run successfully and created the new table populated correctly with Code Amigo data.


kelvin-chappell:
> Should we also add the `amigo_bake_packages` to the Prisma schema (with `@@ignored`) for now? See #1338.

👍 Done in bfa14f743131f27b0bc5fa2c626f47eee632509f

packages/cdk/lib/cloudquery/index.ts
```diff
@@ -588,6 +594,47 @@ export function addCloudqueryEcsCluster(
...
 	};
 
+	const packagesBucketName = new GuStringParameter(
+		scope,
+		'packagesBucketNameParam',
+		{
+			fromSSM: true,
+			default: `/${stage}/deploy/amigo/amigo.data.bucket`,
+		},
+	).valueAsString;
```
akash1810, line 604: Who creates/owns this SSM Parameter? Could it be created by AMIgo, as [it creates the bucket](https://github.com/guardian/amigo/blob/925f9da0203d30d4c6bab2fdaa8688bb5a1b464f/cdk/lib/amigo.ts#L120-L123)?

kelvin-chappell: I think it's created by Amigo as it pre-exists this integration.

tjsilver (approved):
Great work!

kelvin-chappell:
Ran the import process manually and it appears to have worked.  Prod Table has data and no errors in log.

PR #1334, opened 2024-11-25
refactor: check for existing PR before sending event to dep graph integrator

tjsilver:
## What does this change?

Moves the function to check for an existing dependency graph integrator PR on a repo from the dependency graph integrator lambda to the repocop lambda.

## Why?

Currently, the function checks for a PR after the required number of repos have been sent for integration. If existing PRs are found for any of the repos, then this reduces the number of new PRs raised per day from the max (currently 5) to between 0 and the max. We need to ensure we reduce the backlog of repos that are currently only protected by Snyk is reduced to zero for production repos by the time Snyk is retired in February 2025.

## How has it been verified?

- [x] tested locally
- [x] tested on CODE


packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -49,6 +51,36 @@ export function doesRepoHaveDepSubmissionWorkflowForLanguage(
...
+	owner: string,
+	author: string,
+) {
+	const pulls = await octokit.paginate(octokit.rest.pulls.list, {
+		owner,
+		repo: repoName,
+		state: 'open',
+	} satisfies PullRequestParameters);
+
+	const found = pulls.filter((pull) => isGithubAuthor(pull, author));
```
akash1810, line 75: not blocking: IIUC at the moment we're getting _all_ PRs, then filtering for ones authored by our user. Depending on how many PRs exist, this operation could be long. For example, if there are 100+ and we get rate limited. Could we use the [`head` parameter in the query](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests) to filter for our PRs in the initial request?

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -49,6 +51,36 @@ export function doesRepoHaveDepSubmissionWorkflowForLanguage(
...
+
+	if (found.length > 1) {
+		console.warn(`More than one PR found on ${repoName} - choosing the first.`);
```
akash1810, line 78: Should we log `found.length` here?

```ts
console.warn(`Expected 1 PR on ${repoName} but found ${found.length}. Choosing the first.`);
```

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -141,10 +174,30 @@ export async function sendReposToDependencyGraphIntegrator(
...
+				console.log(
+					existingPr
+						? `Existing PR found for ${repo.name}`
+						: `PR not found for ${repo.name}`,
+				);
+				if (!existingPr) {
+					selectedRepos.push(repo);
+				}
+			}
+		}
```
akash1810, line 200: minor: Would this be more readable as a `filter`?

akash1810 (approved):
Paired on a review. Couple of non-blocking comments.

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
tjsilver: I can't find a way to do this easily, as we don't have a set branch name on which to identify the PR we are interested in.

PR #1338, opened 2024-11-28
feat: add Github custom properties table to cloudquery

tjsilver:
## What does this change?
- upgrades github cloudquery plugin to v11 which provides a custom properties table
- creates a `github_repository_custom_properties` table in prisma schema
- creates the sql  migration
- adds the table to the docker-compose so that it can be queried in DEV

## Why?

This follows from #1316. Using custom properties will allow us to move away from topics. Custom properties can be
 made compulsory, which means we can enforce repos to choose a property.

The `production` custom property also applies branch protection.

## How has it been verified?

- [x] Migration applied to CODE - the Github tables are copied from CODE to DEV for local testing
- [x] Tested locally - 14 rows returned

### DIFF
--- a/packages/common/prisma/schema.prisma
+++ b/packages/common/prisma/schema.prisma
@@ -380,6 +380,19 @@ model github_repository_branches {
   @@id([org, repository_id, name], map: "github_repository_branches_cqpk")
 }
 
+model github_repository_custom_properties {
+  cq_sync_time   DateTime? @map("_cq_sync_time") @db.Timestamp(6)
+  cq_source_name String?   @map("_cq_source_name")
+  cq_id          String    @unique @map("_cq_id") @db.Uuid
+  cq_parent_id   String?   @map("_cq_parent_id") @db.Uuid
+  org            String
+  property_name  String
+  repository_id  BigInt
+  value          String?
+
+  @@ignore
+}
+
 model github_team_repositories {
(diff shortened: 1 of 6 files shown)


packages/common/prisma/schema.prisma
```diff
@@ -380,6 +380,19 @@ model github_repository_branches {
...
+  cq_sync_time   DateTime? @map("_cq_sync_time") @db.Timestamp(6)
+  cq_source_name String?   @map("_cq_source_name")
+  cq_id          String    @unique @map("_cq_id") @db.Uuid
+  cq_parent_id   String?   @map("_cq_parent_id") @db.Uuid
+  org            String
+  property_name  String
+  repository_id  BigInt
+  value          String?
+
+  @@ignore
```
kelvin-chappell, line 393: What does this do?

tjsilver: Good question! This means that it's ignored by the Prisma client ([see docs](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#ignore-1)). I added `@@ignore` to this model, as we currently don't need it because we're not doing anything with the table in service catalogue. The model being the schema means that the table will be kept up to date nonentheless, IIUC. When we start to interact with the data via the Prisma client in future work, then we will stop ignoring it.

PR #1339, opened 2024-11-29
chore: Suppress AWS Inspector FSBP findings.

AshCorr:
## What does this change?

Temporarily hide AWS Inspector findings from the FSBP dashboard.

## Why?

As far as I'm aware the guidance is to ignore this finding for now as per https://chat.google.com/room/AAAAag0I08g/5caKQThOs8Q/FHanFOG9rnw as it has a cost associated and is pending feedback from InfoSec on what to do about it.

I also want to see my accounts FSBP Dashboard drop to 0 vulnerabilities too 😅 

## How has it been verified?

Don't have credentials to test this locally unfortunately, have included tests however.

### DIFF
--- a/packages/cloudbuster/src/findings.ts
+++ b/packages/cloudbuster/src/findings.ts
@@ -28,6 +28,18 @@ export function findingsToGuardianFormat(
 		};
 	});
 }
+
+export function isSuppressedFinding(
+	finding: cloudbuster_fsbp_vulnerabilities,
+): boolean {
+	// Suppress AWS Inspector findings
+	// These currently have a cost impact associated to them and has limited language support.
+	// With InfoSec at the moment to decide what we do with this control:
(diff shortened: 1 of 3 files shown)


packages/cloudbuster/src/findings.ts
```diff
@@ -28,6 +28,18 @@ export function findingsToGuardianFormat(
 		};
 	});
 }
+
+export function isSuppressedFinding(
+	finding: cloudbuster_fsbp_vulnerabilities,
+): boolean {
+	// Suppress AWS Inspector findings
+	// These currently have a cost impact associated to them and has limited language support.
+	// With InfoSec at the moment to decide what we do with this control:
```
NovemberTang, line 37: It feels a little risky to enumerate security controls we're ignoring in a public repo. Is there a way to filter these findings in a more private way? Could @guardian/devx-security suppress them centrally?

PR #1352, opened 2024-12-05
Configure schedule and memory for AmigoBakePackages task

kelvin-chappell:
## What does this change?
It gives an explicit schedule for running the task and increases memory from 512 MiB.

## Why?
We're seeing the task bump along the current memory limit of 512 MiB.

## How has it been verified?
We should see the task run successfully.


tjsilver (approved):
👍

PR #1355, opened 2024-12-11
fix: log account cloudbuster messages are sent to in PROD

tjsilver:
## What does this change?

Adds the account ID of the target account to the console.log for `cloudbuster`.

## Why?

In `PROD`, the log was still indicating that the messages were being sent to `testing-alerts`, when in fact they were going to the real accounts. 

## How has it been verified?
When this is merged, we should see real account IDs being logged out by `cloudbuster PROD`. I tested the `CODE` variant on CODE and the logging was as expected.

### DIFF
--- a/packages/cloudbuster/src/digests.ts
+++ b/packages/cloudbuster/src/digests.ts
@@ -133,8 +133,6 @@ export async function sendDigest(
 	config: Config,
 	digest: Digest,
 ): Promise<void> {
-	const target = { Stack: 'testing-alerts' };
-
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
LGTM.

Separately, if we used [logger](https://github.com/guardian/service-catalogue/blob/2b134c7239d9577b4e9ea5c05a7d853fe275ebcf/packages/common/src/logs.ts#L59) this log line would appear with markers in Central ELK. Markers make graphing easier, for example.

PR #1356, opened 2024-12-13
refactor: use logger in cloudbuster

tjsilver:
## What does this change?

- Uses logger in place of console.log
- Refactors the logic a little to make it clearer (is it?)

## Why?

- logger allows for markers to be added - see comment in #1355

## How has it been verified?

- [ ] Tested on CODE to verify messages are logged out as expected

### DIFF
--- a/packages/cloudbuster/src/digests.ts
+++ b/packages/cloudbuster/src/digests.ts
@@ -156,10 +156,9 @@ export async function sendDigest(
 						target: { Stack: 'testing-alerts' },
 					};
 		logger.log({
-			message: `Sending ${digest.accountId} (${digest.accountName}) digest to ${JSON.stringify(notificationParameters.target, null, 4)}...`,
+			message: `Sending ${digest.accountId} (${digest.accountName}) digest...`,
 			accountName: digest.accountName,
 			target: notificationParameters.target,
-			stage,
 			enableMessaging,
 		});
 
@@ -169,7 +168,6 @@ export async function sendDigest(
 			message: `Messaging disabled. Anghammarad would have sent: ${JSON.stringify(notifyParams, null, 4)}`,
 			accountName: digest.accountName,
 			target: notifyParams.target,
-			stage,
 			enableMessaging,
 		});
 	}
--- a/packages/cloudbuster/src/digests.ts
+++ b/packages/cloudbuster/src/digests.ts
@@ -2,6 +2,7 @@ import { RequestedChannel } from '@guardian/anghammarad';
 import type { Action, Anghammarad, NotifyParams } from '@guardian/anghammarad';
 import type { cloudbuster_fsbp_vulnerabilities } from '@prisma/client';
 import { stringToSeverity } from 'common/src/functions';
+import { logger } from 'common/src/logs';
 import type { SecurityHubSeverity, Severity } from 'common/src/types';
 import { type Config } from './config';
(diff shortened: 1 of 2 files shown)


packages/cloudbuster/src/digests.ts
```diff
@@ -146,25 +147,40 @@ export async function sendDigest(
...
-
-		console.log(
-			`Sending ${digest.accountId} digest to ${JSON.stringify(testNotifyParams.target, null, 4)}...`,
-		);
-
-		await anghammaradClient.notify(testNotifyParams);
+	if (enableMessaging) {
+		if (stage === 'PROD') {
+			logger.log({
+				message: `Sending ${digest.accountId} (${digest.accountName}) digest to ${JSON.stringify(notifyParams.target, null, 4)}...`,
```
akash1810, line 153: With the addition of log line markers, do we still need to have `${JSON.stringify(notifyParams.target, null, 4)}` in the message?

packages/cloudbuster/src/digests.ts
```diff
@@ -146,25 +147,40 @@ export async function sendDigest(
...
-			target: { Stack: 'testing-alerts' },
-		};
-
-		console.log(
-			`Sending ${digest.accountId} digest to ${JSON.stringify(testNotifyParams.target, null, 4)}...`,
-		);
-
-		await anghammaradClient.notify(testNotifyParams);
+	if (enableMessaging) {
+		if (stage === 'PROD') {
```
akash1810, line 151: Do we need the stage check here in addition to `enableMessaging`? The inclusion of the stage check makes the meaning of `enableMessaging` overly confusing.

tjsilver: IIUC this check is required because we want see the messages appear in, say, the `P&E/Anghammarad Testing` channel when running on CODE with messaging on.

packages/cloudbuster/src/digests.ts
tjsilver: I guess we do, because we get an eslint error otherwise: `Invalid type "{ Stack: string; }" of template literal expression.eslint[@typescript-eslint/restrict-template-expressions](https://typescript-eslint.io/rules/restrict-template-expressions)`

packages/cloudbuster/src/digests.ts
akash1810: Ah. I think we can make this more explicit. For example:

```ts
if(enableMessaging) {
  const notificationParameters = stage === "PROD" 
    ? notifyParams 
    : { ..notifyParams, target: { Stack: 'testing-alerts' } }; // non-PROD messaging always gos to the `testing` channels

  await anghammaradClient.notify(notificationParameters);
}
```

packages/cloudbuster/src/digests.ts
akash1810: > I guess we do, because we get an eslint error otherwise: `Invalid type "{ Stack: string; }" of template literal expression.eslint[@typescript-eslint/restrict-template-expressions](https://typescript-eslint.io/rules/restrict-template-expressions)`

A little confused by this. the `message` property of the `logger.log` call is [typed as `string`](https://github.com/guardian/service-catalogue/blob/c4850b7cc47a4889827b1cb1ebc16394866570d2/packages/common/src/logs.ts#L45) 🤔.

I'm likely missing something. Want to pair?

PR #1364, opened 2024-12-23
feat: allow repos to be ignored by dep graph integrator

tjsilver:
## What does this change?

- Adds a property to config for repos to be ignored by dependency graph integrator, by language (Scala or Kotlin).
- Adds a function to test a repo to see if it's in the ignored list.
- Adds a test for this function.

## Why?

There is at least one edge case where a repo contains a dependency graph integrator language, but that doesn't use sbt (it contains just one Scala file which is used in a script). We would like to allow teams to exclude these repos from the integrator so that it doesn't repeatedly raise PRs on those repos.

## How has it been verified?

- [x] Tested locally

### DIFF
--- a/packages/repocop/src/config.ts
+++ b/packages/repocop/src/config.ts
@@ -6,6 +6,7 @@ import {
 	getDevDatabaseConfig,
 } from 'common/src/database-setup';
 import type { DatabaseConfig, PrismaConfig } from 'common/src/database-setup';
+import type { DepGraphLanguage } from 'common/types';
 
 export interface Config extends PrismaConfig {
 	/**
@@ -59,6 +60,11 @@ export interface Config extends PrismaConfig {
 	dependencyGraphIntegratorTopic: string;
 
 	/**
+	 * A list of repos to be excluded by the dependency graph integrator, by language.
+	 */
+	dependencyGraphIgnoredRepos: Record<DepGraphLanguage, string[]>;
+
+	/**
 	 * The name of the GitHub organisation that owns the repositories.
 	 */
 	gitHubOrg: string;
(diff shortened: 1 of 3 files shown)


packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.test.ts
```diff
@@ -178,6 +187,16 @@ describe('When getting suitable events to send to SNS', () => {
 
 		expect(result).toEqual(expected);
 	});
+	test('return empty array when an ignored Scala repo is found with without an existing workflow', () => {
```
adamnfish, line 190: It's a minor point, but we aren't checking that filtering works on an array of repos that partially matches the filter - it's all or nothing at the moment. Is that worth adding? And if so, as dicsussed in the other comment, is it worth adding in a describe block that surrounds that test and this `'return empty array ... '` one so that the shared test setup of ignored repos is right here alongside its users?

```typescript
describe('when repositories are ignored', () => {
  const ignoredRepo = 'guardian/ignore-me';
  const ignoredRepos = {
	Scala: ['ignore-me'],
	Kotlin: [],
  };

  test('return empty array when an ignored Scala repo is found with without an existing workflow', () => {
    // ...
  });

  test('returns an array without any ignored repos', () => {
    // ... check the partial filter result here
  });
});
```

adamnfish: Also I noticed while making this example that this test name might want a look?:

> return empty array when an ignored Scala repo is found with without an existing workflow

packages/repocop/src/config.ts
```diff
@@ -58,6 +59,11 @@ export interface Config extends PrismaConfig {
 	 */
 	dependencyGraphIntegratorTopic: string;
 
+	/**
+	 * A list of repos to be excluded by the dependency graph integrator, by language.
```
akash1810, line 63: Worth explicitly saying this is a repository's short name?

PR #1385, opened 2025-01-09
Feat: allow for exemption of repos from dependency graph integrator

tjsilver:
## What does this change?

- Repocop collects the `github_repositories_custom_properties` table.
- Adds a check for the custom property `gu_dependency_graph_integrator_ignore` and excludes any repos that have this property from dependency graph integration.
- Updates the tests and adds a couple more for the exemptions.

## Why?

There is at least one repo that we know of that contains a script written in Scala file but that doesn't use sbt. As there are no dependencies for Scala in the repo, it doesn't need to have the sbt dependency submission workflow. Currently, the dependency graph integrator will identify this repo as requiring a workflow, and continue to raise a PR against it if the existing PR is closed or if the workflow is not merged in.

## How has it been verified?

- [x] Added the custom property to the repo in question and observed the custom property appear in the Cloudquery table `github_repositories_custom_properties`.
- [x] Tested locally
- [x] Tested on CODE

### DIFF
--- a/packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.test.ts
+++ b/packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.test.ts
@@ -21,7 +21,8 @@ import {
 
 const fullName = 'guardian/repo-name';
 const fullName2 = 'guardian/repo2';
-const scalaLang = 'Scala';
+const scala = 'Scala';
+const kotlin = 'Kotlin';
 
 function createActionsUsage(
 	fullName: string,
@@ -75,8 +76,11 @@ function repositoryWithDepGraphLanguage(
 	};
 }
 
-function repoWithTargetLanguage(fullName: string): github_languages {
-	return repoWithLanguages(fullName, ['Scala', 'TypeScript']);
+function repoWithTargetLanguage(
+	fullName: string,
+	language: DepGraphLanguage,
+): github_languages {
+	return repoWithLanguages(fullName, [language, 'TypeScript']);
 }
 
 function repoWithoutTargetLanguage(fullName: string): github_languages {
@@ -104,8 +108,8 @@ describe('When trying to find repos using Scala', () => {
 	test('return true if Scala is found in the repo', () => {
 		const result = checkRepoForLanguage(
 			repository(fullName),
-			[repoWithTargetLanguage(fullName)],
-			scalaLang,
+			[repoWithTargetLanguage(fullName, scala)],
+			scala,
 		);
 
 		expect(result).toBe(true);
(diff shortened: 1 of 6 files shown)


packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -119,10 +121,27 @@ async function sendOneRepoToDepGraphIntegrator(
 	}
 }
 
-export function getReposWithoutWorkflows(
+export function repoIsExempted(
```
kelvin-chappell, line 124: Should the language be passed in here?  ie. 'Scala' or 'Kotlin'

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -119,10 +121,27 @@ async function sendOneRepoToDepGraphIntegrator(
...
+	repo: Repository,
+	exceptedCustomProperties: github_repository_custom_properties[],
+): boolean {
+	const repoIsExcepted = exceptedCustomProperties.find(
+		(property) => repo.id === property.repository_id,
```
kelvin-chappell, line 129: Should the value of the property be compared to the language here? eg. is it Scala

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.test.ts
```diff
@@ -119,6 +120,32 @@ describe('When trying to find repos using Scala', () => {
...
+		cq_source_name: null,
+		cq_id: 'id1',
+		cq_parent_id: null,
+		org: 'guardian',
+		property_name: 'gu_dependency_graph_integrator_ignore',
+		repository_id: BigInt(12345),
+		value: 'Scala',
+	};
+}
+
```
kelvin-chappell, line 148: Perhaps we need another test for Kotlin and one for an entirely irrelevant language to prove that we distinguish

tjsilver: The custom property only accepts either 'Scala' or 'Kotlin', but I've added tests for Kotlin.

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -123,18 +123,19 @@ async function sendOneRepoToDepGraphIntegrator(
...
 export function repoIsExempted(
 	repo: Repository,
-	exceptedCustomProperties: github_repository_custom_properties[],
+	exemptedCustomProperties: github_repository_custom_properties[],
+	language: DepGraphLanguage,
 ): boolean {
-	const repoIsExcepted = exceptedCustomProperties.find(
+	const exemptedRepo = exemptedCustomProperties.find(
 		(property) => repo.id === property.repository_id,
 	);
```
kelvin-chappell: To ensure that function only returns true if exemption is for the relevant language:
```suggestion
	const exemptedRepo = exemptedCustomProperties.find(
		(property) => repo.id === property.repository_id && language === property.value
	);
```

packages/repocop/src/remediation/dependency_graph-integrator/send-to-sns.ts
```diff
@@ -123,18 +123,19 @@ async function sendOneRepoToDepGraphIntegrator(
...
-	if (repoIsExcepted) {
+	if (exemptedRepo && exemptedRepo.value === language) {
```
kelvin-chappell, line 132: Then:
```suggestion
	if (exemptedRepo) {
```

kelvin-chappell (approved):
👍

PR #1387, opened 2025-01-09
refactor: explicitly add custom props table to cloudquery source

tjsilver:
## What does this change?

Explicitly adds the `github_repository_custom_properties` table to the list of those collected by the GithubRepositories Cloudquery source.

## Why?

All child tables are automatically collected, unless we specify them in the `skipTables` property. However, for clarity it is good to explicitly say what tables we want to collect.

## How has it been verified?

- [x] CI passes


kelvin-chappell (approved):
👍

PR #1392, opened 2025-01-20
Use Java 21 in dependency graph submission workflow

kelvin-chappell:
## What does this change?
Using Java 21 to run the dependency graph submission workflow rather than Java 17.

## Why?
Java 21 is the latest LTS version.

## How has it been verified?
Several workflows have successfully run using Java 21, 
eg. https://github.com/guardian/myguardian-support-services/actions/runs/12869554808


tjsilver (approved):
💯

PR #1403, opened 2025-01-31
Delete .github/workflows/snyk.yml

akash1810:
## What does this change?
From 2025, we're moving away from Snyk in favour of Dependabot. This change removes the Snyk workflow.


kelvin-chappell (approved):
👍

PR #1404, opened 2025-01-31
feat: Stop collecting Snyk data

akash1810:
## What does this change?
We're no longer using Snyk so do not need to collect data from it.

## How has it been verified?
The snapshot diff shows the Snyk task being removed.

---

Co-authored-by: @kelvin-chappell.


kelvin-chappell (approved):
👍

packages/cdk/lib/service-catalogue.ts
```diff
@@ -193,10 +193,6 @@ export class ServiceCatalogue extends GuStack {
 			dataType: ParameterDataType.TEXT,
 		});
 
-		const snykReadOnlyKey = new Secret(this, 'snyk-credentials', {
-			secretName: `/${stage}/${stack}/${app}/snyk-credentials`,
-		});
```
akash1810, line 198: From the snapshot, we can see this resource has `"DeletionPolicy": "Delete"`, therefore we expect the Secrets Manager secret to be removed on deploy.

PR #1405, opened 2025-01-31
feat: Update the CloudQuery CLI to v6.15.0

akash1810:
## What does this change?
This is the latest in the v6 major. [v6.0.0](https://github.com/cloudquery/cloudquery/blob/main/cli/CHANGELOG.md#600-2024-07-18) defaulted `skip_dependent_tables` to `true`, to avoid accidentally skipping a crucial table we're setting `skip_dependent_tables` to `false`.

We can remove this once we've switched to an "allow list" of tables.

## Why?
Keeping up to date.

## How has it been verified?
N/A

---

Co-authored by: @kelvin-chappell.

### DIFF
--- a/.env
+++ b/.env
@@ -3,7 +3,7 @@
 # Environment variables shared between ci and DEV.
 
 # See https://github.com/cloudquery/cloudquery/releases?q=cli
-CQ_CLI=5.2.0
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

PR #1407, opened 2025-01-31
feat: Remove the `github_issue_pullrequest_reviews` table from the database

akash1810:
## What does this change?
Since #1350 we have stopped collecting this table.

## Why?
The table is redundant.

---
Co-authored by: @kelvin-chappell.


kelvin-chappell (approved):
👍

PR #1428, opened 2025-02-04
feat: Change `frequency` column in `cloudquery_table_frequency` to milliseconds

akash1810:
> [!NOTE]
> These database changes have been [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/d96682c5-edcc-438c-8014-5dbedb24175e) and therefore the database schema is only compatible with this branch. Before testing another branch on CODE, we'll need to apply another migration to revert the schema change. Or, rebase the feature branch on this one.

## What does this change and why?
Updates the `cloudquery_table_frequency` table to store frequency as milliseconds. Using milliseconds allows for a higher degree of accuracy. For example, we can identify tables that are updated multiple times a day.

To understand the daily rate, we can do:

```sql
SELECT table_name
       , 86400000 / frequency AS daily_rate -- 86400000 = milliseconds in a day
FROM   cloudquery_table_frequency;
```

## How has it been verified?
I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/d96682c5-edcc-438c-8014-5dbedb24175e) and seen the migration apply:

image

Using the above SQL, we can see [the effective daily rate of each table](https://metrics.code.dev-gutools.co.uk/goto/ElIX4cKHR?orgId=1):

image

### DIFF
--- a/packages/common/prisma/schema.prisma
+++ b/packages/common/prisma/schema.prisma
@@ -636,8 +636,8 @@ model audit_results {
 }
 
 model cloudquery_table_frequency {
-  table_name String  @id
-  frequency  String?
+  table_name String @id
+  frequency  BigInt
 }
 
 model guardian_non_p_and_e_github_teams {
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -554,7 +554,7 @@ spec:
             "Command": [
               "/bin/sh",
               "-c",
-              "psql -c "INSERT INTO cloudquery_table_frequency VALUES ('amigo_bake_packages', 'DAILY') ON CONFLICT (table_name) DO UPDATE SET frequency = 'DAILY'"",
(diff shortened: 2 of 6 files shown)


packages/common/prisma/schema.prisma
```diff
@@ -636,8 +636,8 @@ model audit_results {
 }
 
 model cloudquery_table_frequency {
-  table_name String  @id
-  frequency  String?
+  table_name String @id
+  frequency  BigInt
```
adamnfish, line 640: I like that this is no longer an optional field in the sceham 👍

adamnfish (approved):
Looks great to me, I agree this is the simpler approach overall!

I hadn't had a chance to go through the rest of the catalogue (and grafana dashboards) and check if there are other jobs that would be affected by this change, but on the assumption that we've at least checked there's nothing critical that will break with these type changes, I think we could merge this and see.

I'll update the cloudquery paid rows dashboard to match the new form once this is merged and has run to populate the table.

akash1810:
I'll orchestrate this on PROD by:
1. Disabling CD
2. Stopping any running tasks
3. Deploying the database migration
4. Deploying the rest of the changes
5. Restart the stopped tasks
6. Re-enable CD

PR #1437, opened 2025-02-12
feat: Create a new SQL view for VPC details

akash1810:
## What does this change?
Querying a VPC and its subnets to understand which subnet is private or public and their IP space is fairly complicated.
Encapsulating this in a view means this complexity is done once.

Rather than detail the complexity in the PR description, I've attempted to explain it with inline comments.

The data in this view enables us to create a dashboard to more easily visualise the shape of a VPC. I'm not yet sure how often we'll `SELECT` from the view. If we find we're exhausting the database, we could consider making it a materialised view that's [automatically refreshed after data collection](https://github.com/guardian/service-catalogue/pull/1048).

## How has it been verified?
I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/dd10ec76-fb50-4677-870b-c18078a54cf3) and can now see the view in [Grafana CODE](https://metrics.code.dev-gutools.co.uk/goto/nwFrr6KNR?orgId=1).

### DIFF
--- a/packages/common/prisma/migrations/20250212092000_aws_vpc_tables/migration.sql
+++ b/packages/common/prisma/migrations/20250212092000_aws_vpc_tables/migration.sql
@@ -1,0 +1,75 @@
+BEGIN TRANSACTION;
+    CREATE TABLE IF NOT EXISTS "aws_ec2_route_tables" (
+        "_cq_sync_time" TIMESTAMP(6),
+        "_cq_source_name" TEXT,
+        "_cq_id" UUID NOT NULL,
+        "_cq_parent_id" UUID,
+        "account_id" TEXT,
+        "region" TEXT,
+        "arn" TEXT,
+        "tags" JSONB,
+        "associations" JSONB,
+        "owner_id" TEXT,
+        "propagating_vgws" JSONB,
+        "route_table_id" TEXT,
+        "routes" JSONB,
+        "vpc_id" TEXT,
+
+        CONSTRAINT "aws_ec2_route_tables_cqpk" PRIMARY KEY ("_cq_id")
+    );
+    
+    CREATE TABLE IF NOT EXISTS "aws_ec2_subnets" (
+        "_cq_sync_time" TIMESTAMP(6),
+        "_cq_source_name" TEXT,
+        "_cq_id" UUID NOT NULL,
+        "_cq_parent_id" UUID,
+        "request_account_id" TEXT,
+        "request_region" TEXT,
+        "arn" TEXT,
+        "tags" JSONB,
+        "assign_ipv6_address_on_creation" BOOLEAN,
+        "availability_zone" TEXT,
+        "availability_zone_id" TEXT,
+        "available_ip_address_count" BIGINT,
+        "cidr_block" TEXT,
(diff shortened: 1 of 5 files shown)


packages/common/prisma/migrations/20250212092000_aws_vpc_tables/migration.sql
akash1810, line 1: This migration is to make Prisma aware of the tables this new view depends on. In reality, this migration is a no-op on CODE and PROD.

packages/common/prisma/views/public/view_aws_vpcs.sql
akash1810, line 1: This file is auto-generated by the Prisma migration.

packages/dev-environment/docker-compose.yaml
```diff
@@ -28,6 +28,9 @@ services:
         - github_workflows
         - github_languages
         - aws_ec2_instances
+        - aws_ec2_route_tables
+        - aws_ec2_subnets
+        - aws_ec2_vpcs
```
akash1810, line 33: This changes mean we'll copy these tables from CODE to DEV when running locally.

packages/common/prisma/schema.prisma
akash1810, line 1: The changes in this file keep the migrations and Prisma schema in-sync. We're not going to programmatically use them right now, so the `@@ignore` annotation is added.

packages/common/prisma/migrations/20250212092300_view_aws_vpcs/migration.sql
akash1810, line 1: This is the main file of interest in this PR.

packages/common/prisma/migrations/20250212092300_view_aws_vpcs/migration.sql
```diff
@@ -0,0 +1,177 @@
...
+           - The total number of IP addresses available in a VPC is the sum of the total number of IP addresses available in its subnets
+         */
+        , data as (
+            select  vpc.account_id
+                    , acc.name as account_name
+                    , vpc.region
+                    , vpc.vpc_id
+                    , vpc.cidr_block as vpc_cidr_block
+                    , vpc.is_default
+                    , vpc.tags as vpc_tags
```
akash1810, line 153: I debated about including the VPC and subnet tags, as we could get them via a join at the point of needing them.

packages/common/prisma/migrations/20250212092000_aws_vpc_tables/migration.sql
akash1810, line 1: This migration is to make Prisma aware of the tables this new view depends on. In reality, this migration is a no-op on CODE and PROD.

packages/common/prisma/views/public/view_aws_vpcs.sql
akash1810, line 1: This file is auto-generated by the Prisma migration.

packages/common/prisma/schema.prisma
akash1810, line 1: The changes in this file keep the migrations and Prisma schema in-sync. We're not going to programmatically use them right now, so the `@@ignore` annotation is added.

jorgeazevedo (approved):

PR #1442, opened 2025-02-18
feat: remove Snyk tables from the database

tjsilver:
## What does this change?

A migration to remove the four Snyk tables found in PROD from the database. Two of these tables (`snyk_issues` and `snyk_projects`) appear in the Prisma schema, while the other two do not (`snyk_sbom` and `snyk_organizations`). I have added `DROP TABLE` commands manually to the migration file for these.

Also removes the rows related to these tables from the `cloudquery_table_frequency` table.

## Why?

We no longer need these tables as Snyk has been deprecated.

## How has it been verified?

- [x] Tested locally with two local tables which were removed from the database upon the performance of a dev migration.
- [x] Tested on CODE and all 4 tables were removed from the database after this branch was deployed. 
- [x] Tested on CODE and no snyk tables appeared in the `cloudquery_table_frequency` table after the migration was applied.


packages/common/prisma/migrations/20250218105625_remove_snyk_tables/migration.sql
```diff
@@ -0,0 +1,12 @@
+-- DropTable
+DROP TABLE "snyk_issues";
```
akash1810, line 2: Should we add a `IF EXISTS` to these statements too?

packages/common/prisma/migrations/20250218105625_remove_snyk_tables/migration.sql
```diff
@@ -0,0 +1,12 @@
...
+
+-- DropTable
+DROP TABLE IF EXISTS "snyk_projects";
+
+-- DropTable
+DROP TABLE IF EXISTS "snyk_sbom";
+
+-- DropTable
+DROP TABLE IF EXISTS "snyk_organizations";
+
```
akash1810, line 12: Should we also remove references of these tables from the `cloudquery_table_frequency` table, which drives the [usage dashboard](https://metrics.gutools.co.uk/d/debncbyh7b37ke/cloudquery-paid-usage?orgId=1)?

akash1810 (approved):

PR #1478, opened 2025-04-01
Remove unstable test to fix build

jacobwinch:
The expected result for this test seems to differ depending on when it is run (i.e. it passed yesterday and failed today). See also: https://github.com/guardian/service-catalogue/commit/a21e159107e7940ac40634a63e7633d8062a65df.

We could probably improve the function to make it more testable (or just update the expected value again), but I think it might be better to just remove this test.


jorgeazevedo (approved):
Agree! I think using a duration-based approach here is tricky because a calendar month does not have a constant duration.

PR #1486, opened 2025-04-03
feat: Adds a view to map from domains to tags

SHession:
## What does this change?

This PR attempts to implement via the Prisma migration process a view which provides convenient mapping being a domain and resource tags. This is primarily being used for the the Tools Audit Grafana dashboard which is currently identifying tools via their domains: https://metrics.gutools.co.uk/d/aegzv59r9eghsd/tool-audit?orgId=1&var-domain=workflow.gutools.co.uk&var-path=All&var-interval=6h&var-chosen_apps=All&from=now-30d&to=now

## Why?

Domains can be linked to tags in many different ways. In order to try to rationalise this and provide a consistent interface to map between the two, I have created this view.

## How has it been verified?

???


SHession:
Discussed previously with members of DevX. However, I'm unsure how to properly implement this change and have had issues running the project locally due to VPN and permission restrictions.

PR #1498, opened 2025-04-15
feat: Alarm on `EBSByteBalance` metric anomalies

akash1810:
## What does this change?
Provisions an [anomaly detection](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Anomaly_Detection.html) alarm for the `EBSByteBalance` metric of the RDS instance. This is in reaction to a recent incident where we saw the database being unusable from Grafana. In triaging this incident, we saw the `EBSByteBalance` metric appearing to correlate with performance issues. With this alarm, we hope to catch issues earlier.

An anomaly detection alarm is chosen over a standard alarm as it wasn't immediately clear what the threshold for a standard alarm should be, as the existing metrics are quite spiky.

This changes instantiates level 1 CDK constructs as there aren't any level 2 constructs available for anomaly detection. The code was largely inspired by https://aws.amazon.com/de/blogs/networking-and-content-delivery/monitoring-load-balancers-using-amazon-cloudwatch-anomaly-detection-alarms/.

See #1471.

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/7b444714-1cad-4055-947c-0f963df27dde) and inspected the resulting alarm in the AWS console, it seems OK, or at least there are few "in alarm" times:

image

### DIFF
--- a/packages/cdk/bin/cdk.ts
+++ b/packages/cdk/bin/cdk.ts
@@ -24,6 +24,7 @@ new ServiceCatalogue(app, 'ServiceCatalogue-PROD', {
 	databaseDeletionProtection: true,
 	databaseMultiAz: true,
 	databaseInstanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.LARGE),
+	databaseEbsByteBalanceAlarm: true,
 });
 
 new ServiceCatalogue(app, 'ServiceCatalogue-CODE', {
@@ -39,6 +40,7 @@ new ServiceCatalogue(app, 'ServiceCatalogue-CODE', {
 	databaseDeletionProtection: false,
 	databaseMultiAz: false,
 	databaseInstanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
+	databaseEbsByteBalanceAlarm: false,
 });
 
 // Add an additional S3 deployment type and synth riff-raff.yaml
--- a/packages/cdk/lib/service-catalogue.test.ts
+++ b/packages/cdk/lib/service-catalogue.test.ts
@@ -23,6 +23,7 @@ describe('The ServiceCatalogue stack', () => {
 				InstanceClass.T4G,
 				InstanceSize.SMALL,
 			),
+			databaseEbsByteBalanceAlarm: true,
 		});
 		const template = Template.fromStack(stack);
 		expect(template.toJSON()).toMatchSnapshot();
@@ -45,6 +46,7 @@ describe('The ServiceCatalogue stack', () => {
 				InstanceClass.T4G,
 				InstanceSize.SMALL,
 			),
+			databaseEbsByteBalanceAlarm: true,
 		});
 
 		const lambdas = stack.node
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -16,6 +16,12 @@ import {
 import { GuardianPrivateNetworks } from '@guardian/private-infrastructure-config';
 import type { App } from 'aws-cdk-lib';
 import { Duration, Tags } from 'aws-cdk-lib';
+import {
+	CfnAlarm,
+	CfnAnomalyDetector,
+	ComparisonOperator,
+	TreatMissingData,
+} from 'aws-cdk-lib/aws-cloudwatch';
 import type { InstanceType } from 'aws-cdk-lib/aws-ec2';
 import { Peer, Port } from 'aws-cdk-lib/aws-ec2';
 import type { Schedule } from 'aws-cdk-lib/aws-events';
@@ -107,13 +113,18 @@ interface ServiceCatalogueProps extends GuStackProps {
 	 * Should the database be provisioned in multiple availability zones for increased resilience?
 	 */
 	databaseMultiAz: boolean;
+
+	/**
+	 * Should the EBS byte balance metric of the database be monitored?
+	 */
+	databaseEbsByteBalanceAlarm: boolean;
 }
 
 export class ServiceCatalogue extends GuStack {
 	constructor(scope: App, id: string, props: ServiceCatalogueProps) {
 		super(scope, id, props);
 
-		const { stage, stack } = this;
+		const { account, stage, stack } = this;
 		const app = props.app ?? 'service-catalogue';
 
 		const {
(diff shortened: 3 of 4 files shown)


packages/cdk/lib/service-catalogue.ts
```diff
@@ -171,6 +186,52 @@ export class ServiceCatalogue extends GuStack {
...
+			evaluationPeriods: 1,
+			actionsEnabled: true,
+			alarmActions: [alertTopic.topicArn],
+			okActions: [alertTopic.topicArn],
+			thresholdMetricId: 'ad1',
+			metrics: [
+				{
+					id: 'ad1',
+					returnData: true,
+					expression: 'ANOMALY_DETECTION_BAND(m1, 2)',
```
jacobwinch, line 217: I worry that the current implementation will warn us about small fluctuations that aren't particularly important to the level of service that we're providing:

image

IIUC the small red segments of this line suggest that our alarm would have fired a few times in the last week(?), even though we always have around 99% balance and the service seems to be performing well from a user-perspective.

Based on [these docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/using-metric-math.html#metric-math-syntax), I think the way to make this more tolerant is to increase the second argument here to a higher number?

> The function can take two arguments. The first is the ID of the metric to create the band for. The second argument is the number of standard deviations to use for the band. If you don't specify this argument, the default of 2 is used.

packages/cdk/bin/cdk.ts
```diff
@@ -36,6 +37,7 @@ new ServiceCatalogue(app, 'ServiceCatalogue-CODE', {
 	// Do not run CloudQuery tasks in CODE, preferring instead to run them manually using the CLI.
 	enableCloudquerySchedules: false,
 	instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
+	alertTopicName: 'devx-alerts-code',
```
jacobwinch, line 40: I had a quick look at this topic in the console and it seems to be configured to email Mike M directly(!).

I wonder if we should use the main `devx-alerts` topic and/or switch off alarm actions in `CODE` to avoid noise?

packages/cdk/lib/service-catalogue.ts
```diff
@@ -171,6 +186,52 @@ export class ServiceCatalogue extends GuStack {
...
+		new CfnAlarm(this, `${metric.metricName}Alarm`, {
+			comparisonOperator:
+				ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD,
+			treatMissingData: TreatMissingData.BREACHING,
+			datapointsToAlarm: 1,
+			evaluationPeriods: 1,
+			actionsEnabled: true,
+			alarmActions: [alertTopic.topicArn],
+			okActions: [alertTopic.topicArn],
+			thresholdMetricId: 'ad1',
```
jacobwinch, line 212: Out of interest, is this the part that means that we have to use an L1 construct?

packages/cdk/lib/service-catalogue.ts
akash1810: Great points. I wonder if we should use the current implementation as a way to learn more about this AWS feature? I think we're currently using the lightest configuration; there are more options available, for example [`MetricCharacteristics.PeriodicSpikes`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudwatch-anomalydetector-metriccharacteristics.html).

WDYT?

packages/cdk/bin/cdk.ts
akash1810: Ah, thanks for checking this! Agreed - let's use `devx-alerts` and only enable in PROD 👍🏽.

As an aside, I have been wondering if `guardian/aws-account-setup` could provision an account-wide alerts topic, for reference via an SSM Parameter.

packages/cdk/lib/service-catalogue.ts
akash1810: The [L2 construct for an Alarm](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.Alarm.html) doesn't support an array of metrics, just a [singular one](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.Alarm.html#metric).

There are no L2 constructs for an Anomaly Detector. All the examples I've seen instantiate L1 constructs for the Anomaly Detector and the Alarm - they also don't create an explicit/code connection between the two, which I found confusing!

packages/cdk/lib/service-catalogue.ts
jacobwinch: > I wonder if we should use the current implementation as a way to learn more about this AWS feature?

I think it would be good to learn more about this feature but I worry that introducing a noisy alarm will train us to ignore it!

Perhaps we can ship this to `PROD` with actions disabled and review the alarm history after a few days. If it looks stable we can turn on notifications and if not we can experiment with some of these settings to stabilise things before doing so?

packages/cdk/lib/service-catalogue.test.ts
```diff
@@ -45,6 +46,7 @@ describe('The ServiceCatalogue stack', () => {
 				InstanceClass.T4G,
 				InstanceSize.SMALL,
 			),
+			databaseEbsByteBalanceAlarm: true,
```
jacobwinch, line 49: Minor: I find it a bit confusing when we have stuff in the snapshot that's not in the 'real' stack - perhaps we could set this to `false` to match: https://github.com/guardian/service-catalogue/blob/42e6e6dd47d71996c1f02c60744b3e2726521d64/packages/cdk/bin/cdk.ts#L43

jacobwinch (approved):

packages/cdk/lib/service-catalogue.test.ts
akash1810: Agreed. However I'd like to address this as a separate PR by exporting the CODE and PROD props and snapshot testing each stage.

akash1810:
The last 7 hours on PROD show there isn't too much variance and the alarm never enters the alert state. We have a few tasks that run over the weekend, so will check back next week and go from there.

image

PR #1500, opened 2025-04-22
refactor: Namespace database props

akash1810:
## What does this change?
Adding the `database` prefix to props that impact the database is more intention revealing. See also https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#flat.

This is a no-op, confirmed by the snapshots being left untouched.


adamnfish (approved):
The change LGTM, but I haven't run it to check it works

PR #1501, opened 2025-04-22
fix: Only store unique FSBP findings in Cloudbuster table

kelvin-chappell:
## What does this change?
This change ensures that the `cloudbuster_fsbp_vulnerabilities` doesn't try to store duplicate FSBP findings.

## Why?
We've noticed duplicate findings can appear in the `aws_securityhub_findings` table.  

## How has it been verified?
Tested locally using copies of the duplicated production data rows along with other data.


packages/cloudbuster/src/index.ts
```diff
@@ -28,20 +28,28 @@ export async function main() {
...
 
+	const uniqueTableContents = Array.from(new Set(tableContents));
+
 	logger.log({
 		message: `${tableContents.length} high and critical FSBP findings detected`,
 	});
 
+	if (tableContents.length !== uniqueTableContents.length) {
+		logger.warn({
+			message: `${tableContents.length - uniqueTableContents.length} duplicate FSBP findings detected`
```
adamnfish, line 39: Would it be helpful to log what they are? At the moment it would take a code change and/or running it locally to hunt this down, maybe we could save future us the hassle if something comes up?

kelvin-chappell: Yes I was thinking the same.  Just made the smallest change I could to make sure I've caught it.  As you say, they might not be completely unique but they look as though they are on a first inspection.

kelvin-chappell: Logging duplicate resource ARNs in e11547cf8790ea4539c1792b1d278cdf34ac0703

kelvin-chappell:
Cloudbuster table now updating correctly.

PR #1510, opened 2025-04-25
feat: Daily collection of `aws_costexplorer_cost_custom`

akash1810:
## What does this change?
Stops collecting data for the [`aws_costexplorer_cost_30d` table](https://hub.cloudquery.io/plugins/source/cloudquery/aws/latest/tables/aws_costexplorer_cost_30d), replacing it with the [`aws_costexplorer_cost_custom` table](https://hub.cloudquery.io/plugins/source/cloudquery/aws/latest/tables/aws_costexplorer_cost_custom). This table supports aggregation. We're aggregating by the `App`, `Stack` and `Stage` tags, which should allow us to understand the cost of an individual service.

Data is collected for every account in the AWS organisation. Previously only data for the Workflow account was collected.

Collection of `aws_costexplorer_cost_custom` runs daily at midnight, with a window for two days ago. That is, at midnight of 25 April 2025, we collect data the period of 23 April 2025 - 24 April 2025. Hopefully this allows us to ignore complications of timezones 🤞🏽.

CloudQuery uses the [`GetCostAndUsage`](https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html) for this table, which supports arbitrary time periods. This means, for example, we could get daily data for the year 2024. However I'm not sure yet how best to do this, so suggest solving this in a separate PR.

With 40 accounts in the AWS organisation, this task would now collect 1,200 rows a month (40 accounts * 30 days). This is within our contracted usage.

Additionally, as we're collecting a single day at a time, this task implements a [`write_mode`](https://cli-docs.cloudquery.io/docs/reference/destination-spec#write_mode) of `overwrite` so that historical data is kept. This differs from the default behaviour of `overwrite-delete-stale`:

> - `overwrite-delete-stale`: syncs overwrite existing rows with the same primary key, and delete rows that are no longer present in the cloud.
> - `overwrite`: Same as overwrite-delete-stale, but doesn't delete stale rows from previous syncs.
> - `append`: Rows are never overwritten or deleted, only appended.
>
> – https://cli-docs.cloudquery.io/docs/reference/destination-spec#write_mode

For explicitness, every task now specifies its `write_mode` too.

## How has it been verified?
[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/763bb307-3cd7-49ba-9f85-c9354e405512) and after running the task, there is data in the [`aws_costexplorer_cost_custom` table](https://metrics.code.dev-gutools.co.uk/goto/kCGDel1Ng?orgId=1).

To test the `write_mode` changes, the task was run twice across different days. We can see multiple [records in the table](https://metrics.code.dev-gutools.co.uk/goto/ogFAgu1NR?orgId=1) with different `_cq_sync_time` columns.

---

Closes https://github.com/guardian/service-catalogue/pull/1479.

### DIFF
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -17,6 +17,7 @@ import {
 	amigoBakePackagesConfig,
 	awsSourceConfigForAccount,
 	awsSourceConfigForOrganisation,
+	CloudqueryWriteMode,
 	fastlySourceConfig,
 	galaxiesSourceConfig,
 	githubLanguagesConfig,
@@ -177,7 +178,7 @@ export function addCloudqueryEcsCluster(
 				`export START_DATE=$(date -d "@$(($(date +%s) - ${Duration.days(2).toSeconds()}))" "+%Y-%m-%d")`,
 				`export END_DATE=$(date -d "@$(($(date +%s) - ${Duration.days(1).toSeconds()}))" "+%Y-%m-%d")`,
 			],
-
+			writeMode: CloudqueryWriteMode.Overwrite,
 			config: awsSourceConfigForAccount(
 				GuardianAwsAccounts.Workflow,
 				{
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -168,7 +168,7 @@ export function addCloudqueryEcsCluster(
 		{
 			name: 'AwsCostExplorer',
 			description:
-				'Collecting Cost Explorer information for the Worflow account. This requires the use of paid AWS APIs so we are trialling it in a single account first',
+				'Collects daily AWS costs (aggregated by App, Stack and Stage tags). We aim to keep historical data for this table (unlike other tables)',
 			schedule: Schedule.cron({ minute: '0', hour: '0' }),
 
 			// TODO replace with time variable substitution once it supports relative date only strings.
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -179,8 +179,7 @@ export function addCloudqueryEcsCluster(
 				`export END_DATE=$(date -d "@$(($(date +%s) - ${Duration.days(1).toSeconds()}))" "+%Y-%m-%d")`,
 			],
 			writeMode: CloudqueryWriteMode.Overwrite,
-			config: awsSourceConfigForAccount(
-				GuardianAwsAccounts.Workflow,
+			config: awsSourceConfigForOrganisation(
 				{
 					tables: ['aws_costexplorer_cost_custom'],
 				},
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -168,11 +168,49 @@ export function addCloudqueryEcsCluster(
 			name: 'AwsCostExplorer',
 			description:
 				'Collecting Cost Explorer information for the Worflow account. This requires the use of paid AWS APIs so we are trialling it in a single account first',
-			schedule: Schedule.rate(Duration.days(7)),
-			config: awsSourceConfigForAccount(GuardianAwsAccounts.Workflow, {
-				tables: ['aws_costexplorer_cost_30d'],
-				usePaidApis: true,
-			}),
+			schedule: Schedule.cron({ minute: '0', hour: '0' }),
+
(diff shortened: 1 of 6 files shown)


packages/cdk/lib/cloudquery/index.ts
```diff
@@ -168,11 +169,48 @@ export function addCloudqueryEcsCluster(
 			name: 'AwsCostExplorer',
 			description:
 				'Collecting Cost Explorer information for the Worflow account. This requires the use of paid AWS APIs so we are trialling it in a single account first',
```
jacobwinch, line 171: I think we'll want to update this description; perhaps something like:

```suggestion
				'Collects daily AWS costs (aggregated by App, Stack and Stage tags). We aim to keep historical data for this table (unlike other tables)',
```

packages/cdk/lib/cloudquery/index.ts
```diff
@@ -168,11 +169,48 @@ export function addCloudqueryEcsCluster(
...
+			schedule: Schedule.cron({ minute: '0', hour: '0' }),
+
+			// TODO replace with time variable substitution once it supports relative date only strings.
+			//  See https://cli-docs.cloudquery.io/docs/advanced-topics/environment-variable-substitution#time-variable-substitution-example
+			//  See https://github.com/cloudquery/cloudquery/pull/20399
+			additionalCommands: [
+				`export START_DATE=$(date -d "@$(($(date +%s) - ${Duration.days(2).toSeconds()}))" "+%Y-%m-%d")`,
+				`export END_DATE=$(date -d "@$(($(date +%s) - ${Duration.days(1).toSeconds()}))" "+%Y-%m-%d")`,
+			],
+			writeMode: CloudqueryWriteMode.Overwrite,
```
jacobwinch, line 181: Just wanted to double check my own understanding here...

Based on the [docs](https://cli-docs.cloudquery.io/docs/reference/destination-spec#write_mode) and the [primary keys for this table](https://hub.cloudquery.io/plugins/source/cloudquery/aws/latest/tables/aws_costexplorer_cost_custom#primary-keys), I think re-running this task several times on the same day[^1] would be 'safe' (i.e. it won't result in duplicate cost entries for the same day)?

[^1]: I can't imagine us doing this often but I think there is currently a perception within the team that re-running tasks in `PROD` is safe, so it'd be good to understand if this still holds with this new write-mode approach.

jacobwinch (approved):
Great work!

cc @SHession and @rhystmills, who raised a feature request related to this.

packages/cdk/lib/cloudquery/index.ts
akash1810: Yes, this is how I understand it too.

SHession:
> Great work!
> 
> cc @SHession and @rhystmills, who raised a feature request related to this.

Thanks @jacobwinch and @akash1810! 🚀

> CloudQuery uses the [GetCostAndUsage](https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html) for this table, which supports arbitrary time periods. This means, for example, we could get daily data for the year 2024. However I'm not sure yet how best to do this, so suggest solving this in a separate PR.

 It looks like the `aws_costexplorer_cost_custom` is currently empty. Is the idea to backfill it with data going back to 2024?

akash1810:
> We're aggregating by the `App`, `Stack` and `Stage` tags, which should allow us to understand the cost of an individual service.

Oops! https://github.com/guardian/service-catalogue/pull/1511.

akash1810:
> It looks like the `aws_costexplorer_cost_custom` is currently empty. Is the idea to backfill it with data going back to 2024?

We now have [yesterday's data in PROD](https://metrics.gutools.co.uk/goto/MEFbfgbNR?orgId=1). We need to have a think about how to run a one-off task to collect all the historical data.

SHession:
> We now have [yesterday's data in PROD](https://metrics.gutools.co.uk/goto/MEFbfgbNR?orgId=1). We need to have a think about how to run a one-off task to collect all the historical data.

That's really useful, thanks. Let me know if I can do anything to help this along.

akash1810:
> CloudQuery uses the [`GetCostAndUsage`](https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html) for this table, which supports arbitrary time periods. This means, for example, we could get daily data for the year 2024. However I'm not sure yet how best to do this, so suggest solving this in a separate PR.

https://github.com/guardian/service-catalogue/pull/1512.

PR #1511, opened 2025-04-28
fix: Call `GetCostAndUsage` correctly

akash1810:
## What does this change?
Patches the changes in #1510, which fails to run with the follow error:

> operation error Cost Explorer: GetCostAndUsage, https response error StatusCode: 400, RequestID: 8e41a8fb-c06f-42b8-9892-6cc3f285d3bf, api error ValidationException: Only two values for GroupBy are allowed

Confession: I only tested #1510 on CODE with a group by of `App` and `Stage`, which is why I didn't see this error originally!

### DIFF
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -197,7 +197,6 @@ export function addCloudqueryEcsCluster(
 									GroupBy: [
 										{ Type: 'TAG', Key: 'App' },
 										{ Type: 'TAG', Key: 'Stack' },
-										{ Type: 'TAG', Key: 'Stage' },
 									],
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/cloudquery/index.ts
```diff
@@ -197,7 +197,6 @@ export function addCloudqueryEcsCluster(
 									GroupBy: [
 										{ Type: 'TAG', Key: 'App' },
 										{ Type: 'TAG', Key: 'Stack' },
-										{ Type: 'TAG', Key: 'Stage' },
```
akash1810, line 200: Remove `Stage` in the basis that the total cost of a service should consider all stages.

PR #1512, opened 2025-05-01
feat: Enable on demand backfill of `aws_costexplorer_cost_custom`

akash1810:
## What does this change?
Following from https://github.com/guardian/service-catalogue/pull/1510 this change updates the Service Catalogue CLI to allow setting of environment variables within the CloudQuery container. This allows us to backfill data for the `AwsCostExplorer` task by running:

```bash
npm -w cli start -- run-task \
  --stage PROD \
  --name AwsCostExplorer \
  --env START_DATE=2025-01-01 \
  --env END_DATE=2025-02-01
```

This utilises the [container overrides](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerOverride.html).

## Why?
To backfill data on demand.

## How has it been verified?
I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/c873a3b6-807a-43db-b492-6a54d4b94610). After running the following:

```bash
npm -w cli start -- run-task \
  --stage CODE \
  --name AwsCostExplorer \
  --env START_DATE=2025-01-01 \
  --env END_DATE=2025-01-10
```

We can see some [historical data in the `aws_costexplorer_cost_custom` table](https://metrics.code.dev-gutools.co.uk/goto/IUfGpHbNg?orgId=1).

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -73,6 +73,12 @@ const parseCommandLineArguments = () => {
 						description: 'The Name tag of the task to run',
 						type: 'string',
 						demandOption: true,
+					})
+					.option('env', {
+						description:
+							'An environment variable to pass to the task. Syntax: KEY=VALUE',
+						type: 'array',
+						demandOption: false,
 					});
 			})
 			.command(Commands.runAll, 'Run all tasks', (yargs) => {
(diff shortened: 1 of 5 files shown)


packages/cli/src/index.ts
```diff
@@ -140,9 +146,32 @@ parseCommandLineArguments()
...
+
+				if (Array.isArray(env)) {
+					const envVars: Record<string, string> = (env as string[])
+						.map((str) => str.split('='))
+						.reduce((acc, [envKey, envValue]) => {
+							return {
+								...acc,
+								[envKey as string]: envValue,
+							};
+						}, {});
```
akash1810, line 162: We're using the `yargs` package to process command line arguments into a string array; `--env START_DATE=2025-01-01 --env END_DATE=2025-02-01` is represented as `[ 'START_DATE=2025-01-01', 'END_DATE=2025-02-01' ]`.

Here, we're converting the string array into an object. Split each element on `=` then use the left hand side as they key and the right hand side as the value. Therefore we have:

- Command line arguments: `--env START_DATE=2025-01-01 --env END_DATE=2025-02-01`
- `env`:  `[ 'START_DATE=2025-01-01', 'END_DATE=2025-02-01' ]`
- `envVars`: `{ START_DATE: '2025-01-01', END_DATE: '2025-02-01' }`

packages/cli/src/index.ts
```diff
@@ -140,9 +146,32 @@ parseCommandLineArguments()
 					: tasks.then((tasks) => tasks.map((task) => task['Name']));
 			}
 			case Commands.run: {
-				const { stack, stage, app, name } = argv;
+				const { stack, stage, app, name, env } = argv;
+
 				const ecsClient = getEcsClient();
 				const ssmClient = getSsmClient();
+
+				if (Array.isArray(env)) {
```
akash1810, line 154: Above, `demandOption` is `false` making the `env` flag optional. Here, we check if it was provided.

jorgeazevedo (approved):

PR #1521, opened 2025-05-12
chore(dependabot): Group eslint plugin updates

akash1810:
## What does this change?
I was surprised to see #1519 and #1520 as separate PRs. This change updates the Dependabot configuration to bump all eslint plugins [together](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#patterns-and-exclude-patterns-groups), ensuring they're compatible with each other.


kelvin-chappell (approved):
👍

PR #1538, opened 2025-06-10
Enable actions for RDS alarm

jacobwinch:
I noticed that the alarm introduced in https://github.com/guardian/service-catalogue/pull/1498 was not configured to notify the team yet.

We wanted to check that it wasn't going to be too noisy before switching the alerts on and now we have data which seems to confirm that this won't be too much of an issue:

image

I've also updated the SNS topic used so that notifications are sent to the Security and Reliability & Operations teams only (rather than the whole DevX stream). I think this smaller audience is more appropriate as the [EndToEnd team](https://galaxies.gutools.co.uk/?selectedNode=endToEndToolsAndInfrastructure) are not responsible for maintaining this service.


jorgeazevedo (approved):
Nice one!

PR #1539, opened 2025-06-16
Filter inspector findings

NovemberTang:
## What does this change?

Reduces the amount if inspector findings we ingest

It would be nice to get the severity filter in the first commit working, but CloudQuery is extremely grumpy about how to get a number into its table spec, and I haven't worked that out yet.

## Why?

Currently our daily findings our 500k. Collecting only active findings cuts that number in half, which is particularly useful when you're about to scale up your use of inspector massively.

## How has it been verified?

Deployed to CODE, verified execution and reduction in size of dataset


tjsilver (approved):
👍

PR #1549, opened 2025-06-24
show affected regions in cloudbuster digest

NovemberTang:
## What does this change?

The region that the vulnerabilities exist in are now shown in the cloudbuster digest

## Why?

Makes it easier for teams to find their vulnerabilities.

## How has it been verified?

Unit tests have been updated to reflect new message. Added a new test to verify behaviour.


tjsilver (approved):
👍

PR #1552, opened 2025-06-26
update some deps dependabot can't manage

NovemberTang:
## What does this change?

Dependabot's octokit upgrades have been failing for months. Let's update the ones we can now, and figure the rest out later.


kelvin-chappell (approved):
I keep rebasing the dep bump and failing to investigate so thanks for this! 👍

PR #1554, opened 2025-06-27
Integrate zizmor with service catalogue

NovemberTang:
## What does this change?

Integrate the service catalogue with [Zizmor](https://docs.zizmor.sh/). Zizmor is a tool that performs static analysis on GitHub actions. It seems to be named after a still-living [New York dermatologist](https://www.bbc.co.uk/news/blogs-trending-35226720), who was famous for his subway adverts?

When integrated with a repo for the first time, This action will not fail when issues are detected - it will instead generate issues in the Security tab of the repo, so teams can deal with them as and when they have time. This mimics the behaviour of dependabot, and you can see an example of existing issues [here](https://github.com/guardian/service-catalogue/security/code-scanning?query=is%3Aopen+branch%3Ant%2Fzizmor-high+).

However, if a _new_ issue is added to the repo, CI will fail. You can see an example of what that looks like in the `Checks` section of this [test PR](https://github.com/guardian/service-catalogue/pull/1555/checks?check_run_id=44909339610)

I think this might be useful to roll out more widely, as GHA is used all the way across the org, and permissions management in actions is not something that most developers are actively thinking about. Let's do some of that thinking for them.

The most important file in this PR is [zizmor.yml](https://github.com/guardian/service-catalogue/pull/1554/files#diff-8e6408444d2bf2bed6f02c5dc62a7f1f6ab7337eae098d332986e6a81411aeb7). The others are just fixing issues that this new action raised

## Why?

Supposedly, CodeQL [supports GitHub Actions](https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/). However, it's never raised any issues. I was curious to see if another tool might be more useful.

CI environment security is important for the provenance of our software. We should limit the ability of these systems to tamper with our projects wherever possible. Zizmor makes that easier by highlighting low-hanging fruit, such as permissions issues and version pinning.

## How has it been verified?

If you want to know more about how zizmor works, you can [read the docs](https://docs.zizmor.sh/) of the tool, and the [github action](https://github.com/zizmorcore/zizmor-action).

But to see how it interacts with this repo in particular, please see the following links

- Notification of existing issues [example](https://github.com/guardian/service-catalogue/security/code-scanning?query=is%3Aopen+branch%3Ant%2Fzizmor-high+)
- Rejection of new issues [example](https://github.com/guardian/service-catalogue/pull/1555/checks?check_run_id=44909339610)

## Due diligence

Zizmor is open source, and looks quite active, with a healthy contributor count. It uses an MIT licence, which is very permissive. It doesn't require any permission beyond the ability to read the repo (which is public), and create security issues (which is optional), so I'm not too concerned about its security threat.

It's also [used](https://docs.zizmor.sh/trophy-case/) by some quite big names, including Apache (Airflow), Canonical, curl, Docker Grafana, and Homebrew, all of which we use ourselves.

### DIFF
--- a/.github/actions/npm-run/action.yml
+++ b/.github/actions/npm-run/action.yml
@@ -31,6 +31,6 @@ runs:
 
     - name: Run script
       shell: bash
-      run: npm run ${{ env.SCRIPT }}
+      run: npm run "${SCRIPT}"
       env:
         SCRIPT: ${{ inputs.script }}
--- a/.github/actions/npm-run/action.yml
+++ b/.github/actions/npm-run/action.yml
@@ -21,7 +21,9 @@ runs:
 
     - uses: guardian/actions-read-private-repos@8792b5279dc2e6dfb6b9aa6ba2f26b6226be444c # v0.1.1
       with:
-        private-ssh-keys: ${{ inputs.deployment_key }}
+        private-ssh-keys: ${{ env.KEY }}
+      env:
+        KEY: ${{ inputs.deployment_key }}
 
     - name: install dependencies
       shell: bash
@@ -29,4 +31,6 @@ runs:
 
     - name: Run script
       shell: bash
-      run: npm run ${{ inputs.script }}
+      run: npm run ${{ env.SCRIPT }}
+      env:
+        SCRIPT: ${{ inputs.script }}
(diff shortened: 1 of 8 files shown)


NovemberTang:
> This looks good! I wonder if we should add the permissions tightening process for workflows to our recommendations as well?

Yes, totally! It might be good to have a more complete solution for this at some point too.

For posterity, as @tjsilver and I already spoke about this:

Zizmor will flag if it thinks dangerous permissions have been granted to every job, but it won't tell you how to fix the issue. Something like the [actions permissions](https://github.com/GitHubSecurityLab/actions-permissions) tool might help with that next step

PR #1564, opened 2025-07-09
Cloudbuster should skip the app tag if none are found

NovemberTang:
## What does this change?

Cloudbuster displays the app tag as `unknown` if it can't find one. If we can't find one, let's skip it.

## Why?

This shortens the message for users, and makes it easier for teams outside P&E (who may not use Stack/Stage/App notation as standard) to understand their messages

## How has it been verified?

Added unit tests to verify behaviour. Tested message contents locally.


kelvin-chappell (approved):
👍

PR #1566, opened 2025-07-11
feat(tests): replace jest with vitest

tjsilver:


tjsilver:
I can't get vitest to play nicely with the snapshots - it doesn't seem to respect the TEST/dummy values - so I'm closing this experiment for now.

PR #1567, opened 2025-07-14
feat(tests): replace jest with vitest

tjsilver:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A lens into your GitHub organisation",
   "scripts": {
     "test": "npm run test --workspaces --if-present",
-    "test-update": "jest -u",
+    "test-update": "vitest -u",
```
NovemberTang, line 7: has this command been tested? I thought test-update was only used with CDK

tjsilver: Oh good spot! That was a leftover from an earlier implementation attempt.

tjsilver:
> We're now using a combination of jest, node:test, and vitest across the service catalogue. 3 seems like too many to be maintainable. Why are we using vitest in addition to node:test, rather than standardising on one?

I agree, it's not ideal. I investigated using node:test instead of vitest and the conversion seemed quite complicated whereas from jest it was a very simple switchover (apart from in the `cdk` project). As part of the work to remove Jest and replace with node:test in `cdk` we could also look to replace Vitest across the other projects. I suggest we investigate this as a subsequent piece of work after upgrading Octokit. WDYT?

NovemberTang (approved):
> I agree, it's not ideal. I investigated using node:test instead of vitest and the conversion seemed quite complicated whereas from jest it was a very simple switchover (apart from in the `cdk` project). As part of the work to remove Jest and replace with node:test in `cdk` we could also look to replace Vitest across the other projects. I suggest we investigate this as a subsequent piece of work after upgrading Octokit. WDYT?

I can definitely see both sides to this. On one hand, converting additional modules represents scope creep. But my worry is that if we don't tackle it now, the subsequent investigation will languish in health backlog as its value becomes less immediate.

IMO, as there's no serious vulnerabilities in our octokit dependencies at the moment, there's no rush to get this PR merged.  Would a migration from node:test to Vitest be simpler? Would that be a change so large that it's more work than it's worth?

I'm not sure which is the best choice so I'll approve and leave it to your discretion!

packages/github-actions-usage/README.md
```diff
@@ -3,23 +3,26 @@
...
 The [`view_github_actions` SQL view](../common/prisma/migrations/20240222201635_guardian_github_actions_usage/migration.sql) also exists
 to make it easier to query the data.
 This view shows the archived status of a repository, the name of the Action being used, and the version.
 
 ## Testing
+
 The tests are written with [Node's test runner](https://nodejs.org/api/test.html).
-This is in contrast to other packages in this repository which use Jest.
-This is because Jest had issues with the `@actions/workflow-parser` package; 
+This is in contrast to other packages in this repository which used Jest.
```
NovemberTang, line 26: Do we still need this explanation if Jest usage is mostly in the past?

packages/obligatron/package.json
```diff
@@ -6,6 +6,6 @@
 		"start": "APP=obligatron tsx src/run-locally.ts",
 		"build": "esbuild src/index.ts --bundle --platform=node --target=node20 --outdir=dist --external:@prisma/client --external:prisma",
 		"typecheck": "tsc --noEmit",
-		"test": "jest --detectOpenHandles --config ../../jest.config.js --selectProjects obligatron"
+		"test": "node --import tsx --test \"**/*.test.ts\""
```
NovemberTang, line 9: This is interesting! so node:test only runs in the current directory rather than the entire project?

packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
```diff
@@ -245,14 +242,12 @@ describe('createDigestForSeverity', () => {
...
-		expect(
-			createDigestForSeverity(
-				team,
-				'high',
-				[ownershipRecord],
-				[resultWithVuln],
-				60,
-			)?.message,
-		).not.toContain(`[${noUrlVuln.package}](`);
+		assert.ok(!createDigestForSeverity(
```
NovemberTang, line 245: This assertion is tricky to read, could we add a comment explaining it, or simplify it a little

packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
```diff
@@ -164,8 +160,9 @@ describe('createDigest', () => {
 			[resultWithVuln, anotherResultWithVuln],
 			60,
 		);
-		expect(digest?.teamSlug).toBe(team.slug);
-		expect(digest?.message).toContain('leftpad');
+		assert.strictEqual(digest?.teamSlug, team.slug);
+		assert.ok(digest.message.includes('leftpad'));
+		// expect(digest.message).toContain('leftpad');
```
NovemberTang, line 165: leftover comment?

tjsilver: yes!

packages/repocop/src/remediation/vuln-digest/vuln-digest.test.ts
tjsilver: Is this better? 7d3c24b7

NovemberTang: yes, much easier to read!

PR #1568, opened 2025-07-17
add suppressed column to cloudbuster

NovemberTang:
## What does this change?

Adds a column to cloudbuster to indicate whether or not a finding has been suppressed. Updates primary key to include suppression, in case an account-level issue has been suppressed in some regions and not others, such as SSM.7. required boolean columns need a value. We are currently excluding suppressed issues, so `FALSE` is correct in all cases

## Why?

We would like to provide teams with a view of their suppressed findings in the FSBP dashboard, so they can see what has been tracked on the risk register. If there is an expiry date on the register, the issue will still need to be visible and trackable.

## How has it been verified?

Works on DEV. On approval, I will deploy to CODE. on merge, I will rollout to PROD


packages/cloudbuster/src/index.ts
```diff
@@ -31,7 +31,7 @@ export async function main() {
 	const controlIdArns = new Map<string, cloudbuster_fsbp_vulnerabilities>();
 	const duplicateControlIdArns: string[] = [];
 	tableContents.forEach((row) => {
-		const compositeKey = `${row.control_id}:${row.arn}`;
+		const compositeKey = `${row.control_id}:${row.arn}:${row.suppressed}`;
```
kelvin-chappell, line 34: Should the suppressed flag be part of the PK?  It looks more natural for the region to be part of the key if that is what varies.  As I understand it, the suppression status could change during the lifetime of a finding

tjsilver (approved):
LGTM 👍

PR #1569, opened 2025-07-21
write suppressed FSBP findings to cloudbuster table

NovemberTang:
## What does this change?

Allows teams to view their suppressed findings. The dashboard will be updated so they are not displayed by default.

Cloudbuster will continue to notify teams only about issues that have not been suppressed.

## Why?

So teams can see all vulnerabilities that they own, not just ones where the risk hasn't been accepted

## How has it been verified?

Tested on CODE, confirmed that suppressed vulnerabilities are showing up.


kelvin-chappell (approved):
👍

PR #1572, opened 2025-07-22
fix: Update S3 event pattern for Prisma migrate task to use 'Object Created' detail type

tjsilver:
Co-authored-by: Natasha <67543397+NovemberTang@users.noreply.github.com>

## What does this change?

Changes the event type for the prisma migrate trigger from Cloudtrail to S3 bucket.


akash1810 (approved):
Nice work!

PR #1573, opened 2025-07-23
feat: Display all CIDR blocks in `view_aws_vpcs`

akash1810:
## What does this change?
Updates the view created in https://github.com/guardian/service-catalogue/pull/1437 to show all CIDR blocks of a VPC. This view primarily drives the [aws-account-vpcs dashboard](https://metrics.gutools.co.uk/d/aecrxbaf4e800e/aws-account-vpcs); the column name and type remain the same (string) so the dashboard shouldn't need updating.

## Why?
Accuracy! Previously, there was an assumption that every VPC had one CIDR block.

## How has it been verified?
I've [ran the query in Grafana](https://metrics.gutools.co.uk/goto/Q7E2QjUHg?orgId=1), filtering for a VPC known to have multiple CIDR blocks. The results are correct.

### DIFF
--- a/packages/common/prisma/migrations/20250723075100_view_aws_vpcs/migration.sql
+++ b/packages/common/prisma/migrations/20250723075100_view_aws_vpcs/migration.sql
@@ -1,0 +1,182 @@
+-- @formatter:off -- this stops IntelliJ from reformatting the SQL
+begin transaction;
+    drop view if exists view_aws_vpcs;
+    drop function if exists fn_view_aws_vpcs;
+
+    create function fn_view_aws_vpcs() returns table (
+        account_id text
+        , region text
+        , vpc_id text
+        , vpc_cidr_block text
+        , is_default boolean
+        , subnet_id text
+        , subnet_type text
+        , subnet_cidr_block text
+        , subnet_total_addresses bigint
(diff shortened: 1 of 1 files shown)


packages/common/prisma/migrations/20250723075100_view_aws_vpcs/migration.sql
```diff
@@ -0,0 +1,182 @@
...
+            select  vpc_id
+                    , jsonb_array_elements(cidr_block_association_set) ->> 'CidrBlock' AS vpc_cidr_block
+            from    aws_ec2_vpcs
+        )
+        , cidr_block_string as (
+            select      vpc_id
+                        , string_agg(vpc_cidr_block, ', ') as vpc_cidr_block
+            from        cidr_blocks
+            group by    vpc_id
+        )
```
akash1810, line 146: The `data` CTE could `JOIN` with `cidr_blocks` directly, however it would then have a long `GROUP BY` clause, grouping by each non-aggregated column.

akash1810, line 146: This is new, compared to https://github.com/guardian/service-catalogue/pull/1437.

PR #1574, opened 2025-07-23
upgrade cq_aws to v32

NovemberTang:
## What does this change?

Upgrades the CloudQuery AWS plugin version from v27 to v32. I've reviewed the changelog for each major version, and there are no breaking changes to tables that we are using.

## Why?

Keeping dependencies up to date.

## How has it been verified?

We should deploy this change to CODE, then introspect the changes to the DB to create a schema migration, as we don't currently have a way to perform a sync locally. If there is a simpler way to do this, please let me know


tjsilver (approved):
👍

PR #1578, opened 2025-07-25
Notify the security team when an interactive is found

NovemberTang:
## What does this change?

Sends a notification to the security team when a new interactive has been labelled, so we can do a quick accuracy assessment.

## Why?

I think it would be useful to get an idea of how well the interactive monitor is working. For example, how often it catches something, how accurate it is, etc.

## How has it been verified?

The lambda has been run locally, and produces the expected output.

### DIFF
--- a/packages/cloudbuster/src/run-locally.ts
+++ b/packages/cloudbuster/src/run-locally.ts
@@ -5,6 +5,6 @@ import { main } from './index.js';
 config({ path: `../../.env` }); // Load `.env` file at the root of the repository
 config({ path: `${homedir()}/.gu/service_catalogue/.env.local` });
 
-if (require.main === module) {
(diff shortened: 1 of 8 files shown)


packages/cloudbuster/src/run-locally.ts
```diff
@@ -5,6 +5,6 @@ import { main } from './index.js';
 config({ path: `../../.env` }); // Load `.env` file at the root of the repository
 config({ path: `${homedir()}/.gu/service_catalogue/.env.local` });
 
-if (require.main === module) {
+if (import.meta.url === `file://${process.argv[1]}`) {
```
tjsilver, line 8: This has been updated in https://github.com/guardian/service-catalogue/pull/1579/files

PR #1579, opened 2025-07-25
fix: local running of packages

tjsilver:
## What does this change?

- Updates the run-locally scripts in various packages to use ESM syntax rather than CommonJS. ESM modules don’t expose a concept of 'main module' like CommonJS does, so Node.js can’t natively tell if a module is the entry point. Instead, we use a comparison of the URL of the current module file with the path of the script. See comments in the `run-locally.ts` files for more information.

- Adds `"type":"module"` to `package.json`s that were missing it.

## Why?

Running locally for several packages was broken after the recent bump to ESM.

## How has it been verified?

Running locally works now for all packages that were broken.

### DIFF
--- a/packages/cloudbuster/src/run-locally.ts
+++ b/packages/cloudbuster/src/run-locally.ts
@@ -1,10 +1,19 @@
 import { homedir } from 'os';
+import { fileURLToPath } from 'url';
 import { config } from 'dotenv';
 import { main } from './index.js';
 
 config({ path: `../../.env` }); // Load `.env` file at the root of the repository
 config({ path: `${homedir()}/.gu/service_catalogue/.env.local` });
 
-if (import.meta.url === `file://${process.argv[1]}`) {
+/**
+ * checks if the current module is the entry point script - replaces the CommonJs `require.main === module` check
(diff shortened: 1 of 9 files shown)


packages/cloudbuster/src/run-locally.ts
```diff
@@ -5,6 +5,6 @@ import { main } from './index.js';
 config({ path: `../../.env` }); // Load `.env` file at the root of the repository
 config({ path: `${homedir()}/.gu/service_catalogue/.env.local` });
 
-if (require.main === module) {
-	void main();
+if (import.meta.url === `file://${process.argv[1]}`) {
```
NovemberTang, line 8: Glad to see the fix is so simple! But do we know what `import.meta.url`, and `process.argv[1]` represent? I'm a bit wary of providing the program with a piece of data I don't really understand

NovemberTang (approved):
LGTM!

packages/cloudbuster/src/run-locally.ts
tjsilver: I've refactored slightly to improve the check and added a detailed comment above each use - see d1dcd83b. WDYT?

PR #1580, opened 2025-07-25
set non-prod obligatron frequency to every 30 days

NovemberTang:
## What does this change?

Typically, we only run service catalogue jobs once a month on CODE to verify that they are still working. There is no need to run obligatron lambdas every day, so let's reduce their frequency

## Why?

Running them every day is not necessary

## How has it been verified?

Snapshots show the correct schedule


tjsilver (approved):
This looks good. One (non-blocking) observation -  I've found with these schedules where only the number of days are mentioned that it's harder to work out on which actual day of the month they are supposed to run, rather than with a cron where it is a bit more obvious.

PR #1581, opened 2025-07-28
feat: Add indexes for `riffraff_deploys`

akash1810:
## What does this change?
I'm starting to query the `riffraff_deploys` table to understand which branches are deployed to which stage. This change adds an index to key fields in the JSON `content` column for improved query performance.

See also https://docs.google.com/document/d/1KkKNLCTZ_5Y1sCy0DetpDH-8aV_SwVKqEFaN3P6A3eU/edit?tab=t.0.


packages/common/prisma/migrations/20250728104000_riffraff_deploys_index/migration.sql
```diff
@@ -0,0 +1,17 @@
+BEGIN TRANSACTION;
+    CREATE TABLE IF NOT EXISTS riffraff_deploys (
+        _cq_sync_time   TIMESTAMP,
+        _cq_source_name TEXT,
+        id              UUID,
+        content         JSONB,
+
+        CONSTRAINT riffraff_deploys_id PRIMARY KEY (id)
+    );
```
akash1810, line 9: This table exists on CODE and PROD since https://github.com/guardian/service-catalogue/pull/424. We've not had a migration for it thus far as we hadn't needed to interact with it from prisma.

PR #1592, opened 2025-08-08
Track the number of repos that do not have a valid production status

NovemberTang:
## What does this change?

- Moves `getRepositories` to common
- Creates a new `REPOSITORY_STATUS` obligation that checks for the existence of particular topics

## What does this not change

At the moment, this code does not attempt to find the repo owner. It only performs the evaluation. This is because it is an MVP, and adding the logic to find the owner would increase the complexity of this PR. This is addressed in #1594 

It also does not find interactives, as `guardian_production_status` doesn't currently include  `interactive` as a value. This will be fixed in #1593 

## Why?

All unarchived repositories should have a valid production status, so that its security obligations can be evaluated accurately. Tracking the number of repos that do not have one of these statuses means we have a ceiling of how many repos are not being tracked by repocop, that should be.

The obligation has been named `REPOSITORY_STATUS` rather than `REPOSITORY_TOPIC` so we can be flexible with the implementation. We may not always rely on GitHub topics to tell us what a repo is doing, and the name reflects that.

## How has it been verified?

Unit tests have been added to verify functionality.
This PR has been tested on CODE, and seems to be running correctly.

### DIFF
--- a/packages/common/src/database-queries.ts
+++ b/packages/common/src/database-queries.ts
@@ -1,5 +1,6 @@
 import type { aws_securityhub_findings, PrismaClient } from '@prisma/client';
-import type { SecurityHubFinding, SecurityHubSeverity } from './types.js';
+import { toNonEmptyArray } from './functions.js';
+import type { Repository, SecurityHubFinding, SecurityHubSeverity } from './types.js';
 /**
  * Queries the database for FSBP findings
  */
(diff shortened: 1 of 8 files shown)


packages/common/src/database-queries.ts
```diff
@@ -24,3 +25,24 @@ export async function getFsbpFindings(
 
 	return findings as unknown as SecurityHubFinding[];
 }
+
+export async function getRepositories(
```
NovemberTang, line 29: This is not a new function, it's just moved from `repocop` to `common`

tjsilver (commented):
Wonder if it would be DRYer to pull the archived/unarchived logic into common as well so that we don't have to perform the filter twice - see here: https://github.com/guardian/service-catalogue/blob/39849778b5638503c52334b7b81f4b15b5d2c9aa/packages/repocop/src/index.ts#L74C1-L78C4

NovemberTang:
> Wonder if it would be DRYer to pull the archived/unarchived logic into common as well so that we don't have to perform the filter twice - see here: https://github.com/guardian/service-catalogue/blob/39849778b5638503c52334b7b81f4b15b5d2c9aa/packages/repocop/src/index.ts#L74C1-L78C4

Although they have a predicate in common, they're doing different things.
In repocop, we are partitioning one array into two, based on whether the predicate returns true or false. In obligatron, we are filtering away any values that fail, and returning one array. To make it DRY, we'd have to partition first, and then obligatron would have an unused array floating around, which would throw a linting error, and potentially be confusing to a future developer.

The second reason to keep things as they are is performance. Currently, each of these snippets iterates through the list once. If we made the repo archiving partition a shared function, obligatron would have to iterate through the list a second time to do the topic filtering.

PR #1597, opened 2025-08-15
Make maximum vulnerability age for digest common

NovemberTang:
## What does this change?

Allows the cutoff date for vulnerability digest notification to be common between repocop and cloudbuster

## Why?

They use the same vulnerability reporting schedule. Typically when we want to change one, we also want to change the other.

## How has it been verified?

Unit tests have been updated
Tested on CODE


NovemberTang:
> Do you think it would be possible to set the value in one place centrally and then reuse that value? Otherwise I foresee it being changed in one place and not another and then going out of sync. Unless you think we need the flexibility.

Yes! That was the initial purpose of the PR. I forgot to wire it all together at the end, and raised the change a little early, apologies. This should be ready now!

PR #1598, opened 2025-08-15
feat: Create a view for the availability dashboard

akash1810:
## What does this change and why?
The availability dashboard has quite a few variables referencing things like stack, stage, app values and asg names. There is a lot of repetition across these variables, which makes them difficult to maintain. This change introduces a view that can be used to source all the necessary information for the availability dashboard.

For example, we currently defines stages as:

```sql
SELECT DISTINCT((tags->>'Stage'))  as Stages
FROM aws_elbv1_load_balancers lb1
  LEFT JOIN grafana_cloudwatch_to_internal_infra infra
  ON lb1.account_id = infra.aws_account_id
WHERE infra.grafana_cloudwatch_datasource = '${AWSAccount:text}'

UNION DISTINCT

SELECT DISTINCT((tags->>'Stage'))  as Stages
FROM aws_elbv2_load_balancers lb2
  LEFT JOIN grafana_cloudwatch_to_internal_infra infra
  ON lb2.account_id = infra.aws_account_id
WHERE infra.grafana_cloudwatch_datasource = '${AWSAccount:text}'
ORDER BY Stages DESC
```

Using this view, it becomes:

```sql
select distinct(stage) 
from view_availability_dashboard 
where grafana_cloudwatch_datasource = '${AWSAccount:text}';
```

## How has it been verified?
There are two types of test:
1. Does the migration work?
    After [deploying to CODE](https://riffraff.gutools.co.uk/deployment/view/0856365d-826e-41ff-824b-6a6d1a4cfeff) the [logs show the migration was applied](https://logs.gutools.co.uk/s/devx/app/r/s/UIl0q) as does the [`_prisma_migrations` table](https://metrics.code.dev-gutools.co.uk/goto/q8mHvmXHg?orgId=1). Consequently the view now [exists on CODE](https://metrics.code.dev-gutools.co.uk/goto/vTCM6gXHR?orgId=1).

2. Is the data accurate?
We'd want to use PROD data for verification. The easiest way to verify this is to [run the body of view on PROD](https://metrics.gutools.co.uk/goto/mp5uDWuNR?orgId=1) and observe the output. Expected results include:
    - The service named "mobile-router-mobile-PROD" has 4 rows as it has a single ALB with multiple targets
    - The service named "prism-deploy-PROD" has a single row as it is a single ALB with a single target


packages/common/prisma/migrations/20250815101000_view_availability_dashboard/migration.sql
```diff
@@ -0,0 +1,125 @@
...
+                    , 'alb' as load_balancer_type
+                    , lb.arn as load_balancer_arn
+                    , lb.load_balancer_name
+
+                    -- Use ASG tags to support the single load balancer to multiple target groups architecture (i.e. MAPI)
+                    , asg.tags ->> 'Stack' as stack
+                    , asg.tags ->> 'Stage' as stage
+                    , asg.tags ->> 'App' as app
+                    , asg.tags ->> 'gu:riff-raff:project' as riffraff_project
+                    , asg.tags ->> 'gu:repo' as repo
```
akash1810, line 80: We need to support 2 architecture patterns:
1. One load balancer, one ASG
2. One load balancer using path based routing to multiple ASGs (used by [mobile-router](https://github.com/guardian/mobile-apps-api/blob/main/cdk/lib/mobile-router.ts))

We use tags to filter application logs within Central ELK. In the first pattern, the load balancer and ASG tags will match. In the second pattern, the load balancer will be tagged `App=mobile-router`. There are no logs returned for this filter. We need to filter logs with the tags of the ASG instead.

Another current usage of `App` is to filter access logs - we use `App` to lookup a target group to then filter the access logs. This view has all this information already, so we shouldn't need to use `App` in this scenario.

jorgeazevedo:
@akash1810 when we spoke about this last week, it was blocked on the fact that it depended on views that had been manually created. Is this still the case?

akash1810:
> @akash1810 when we spoke about this last week, it was blocked on the fact that it depended on views that had been manually created. Is this still the case?

Sort of. Within this PR we [create the view if it doesn't exist](https://github.com/guardian/service-catalogue/pull/1598/files#diff-5b2a4870949d37d4ee883eb88bca0898d15e508ba710d639248173e6a3454583R127-R153). On PROD, things will continue to work as they do now. Locally, as the view won't exist, we'll create the view with nonsensical values. I [suggest](https://github.com/guardian/service-catalogue/pull/1598/files#r2281754772) we fix this separately.

jorgeazevedo (approved):

PR #1599, opened 2025-08-19
Collect SNS topic data every night

jacobwinch:
## What does this change?

This PR adds a dedicated task for SNS topic data, so that we can refresh the data every night. Prior to this PR the `aws_sns_topics` table was updated as part of the [catch-all AWS task](https://github.com/guardian/service-catalogue/blob/22e9930660da89e2be4686462a31562e6060d56d/packages/cdk/lib/cloudquery/index.ts#L339-L371), which [runs weekly](https://github.com/guardian/service-catalogue/blob/22e9930660da89e2be4686462a31562e6060d56d/packages/cdk/lib/cloudquery/index.ts#L349).

## Why?

We want to [monitor/alert on SNS topic misconfiguration](https://trello.com/c/SJE8dmow/368-check-if-sns-topic-subscriptions-for-slo-alerts-have-all-been-confirmed), as this can lead to failures delivering alarm notifications.

These alerts will be more accurate and useful if we are refreshing the data regularly.

## How has it been verified?

I've [deployed](https://riffraff.gutools.co.uk/deployment/view/1afd844a-4665-47fa-9ef0-399a1c259a30) this branch to `CODE`, [run](https://logs.gutools.co.uk/s/devx/app/r/s/gg6Qs) the task[^1] and confirmed that the relevant table is updated by using the _Explore_ feature in Grafana `CODE`.

[^1]: `npm -w cli start run-task -- --stage CODE --name AwsOrgWideSns`

### DIFF
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -250,6 +250,16 @@ export function addCloudqueryEcsCluster(
 			policies: [listOrgsPolicy, cloudqueryAccess('*')],
 		},
 		{
+			name: 'AwsOrgWideSns',
+			description:
+				'Collecting SNS data across the organisation. Uses include monitoring alarm configuration.',
+			schedule: Schedule.cron({ minute: '0', hour: '3' }),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_sns_topics'],
+			}),
+			policies: [listOrgsPolicy, cloudqueryAccess('*')],
+		},
+		{
 			name: 'AwsOrgWideS3',
 			description:
 				'Collecting S3 data across the organisation. Uses include identifying which account a bucket resides.',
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/cloudquery/index.ts
```diff
@@ -249,6 +249,16 @@ export function addCloudqueryEcsCluster(
 			}),
 			policies: [listOrgsPolicy, cloudqueryAccess('*')],
 		},
+		{
+			name: 'AwsOrgWideSns',
+			description:
+				'Collecting SNS data across the organisation. Uses include monitoring alarm configuration.',
+			schedule: Schedule.cron({ minute: '0', hour: '3' }),
+			config: awsSourceConfigForOrganisation({
+				tables: ['aws_sns_topics'],
```
jacobwinch, line 258: There is also a [table for `aws_sns_subscriptions`](https://hub.cloudquery.io/plugins/source/cloudquery/aws/latest/tables/aws_sns_subscriptions), but I don't think we need that data for the use-case that I'm working on.

PR #1601, opened 2025-09-01
Move to allow lists for collected tables

JuliaBrigitte:
## What does this change?
This change switches our cloudquery config to an allow list model, where we only collect data with CloudQuery for the tables that we explicitly list in source control.

## Why?
CloudQuery often add new tables in new releases. This combination means our usage can dramatically increase just by changing a version number. We want to protect ourselves from this scenario.

## How to test?
Cross check the tables with the ones currently in the PROD db

These are the tables not covered in this PR:
_prisma_migrations - prisma table
amigo_bake_packages - left in config
audit_results - can we delete this?
cloudbuster_fsbp_vulnerabilities - security table
cloudquery_plugin_usage - filled by us
cloudquery_table_frequency - filled by us
endoflife_products - left in config
guardian_github_actions_usage - filled by us
guardian_non_p_and_e_github_teams - manually created
guardian_production_status - manually created
obligatron_results - filled by us
repocop_github_repository_rules - security table
repocop_vulnerabilities - security table


JuliaBrigitte:
Spilt out into smaller PRs for ease of testing

PR #1602, opened 2025-09-01
reduce maximum age for digest inclusion

NovemberTang:
## What does this change?

Reduces the window for teams to be alerted about new issues from 60 to 45 days.

## Why?

Reducing the number of times we notify teams about an issue will help teams deal with alert fatigue. 60 days was chosen as an arbitrary number longer than the SLA period - 45 also meets that brief.

## How has it been verified?

CI passes


kelvin-chappell (approved):
👍

PR #1628, opened 2025-09-11
Update versions numbers to latest versions

JuliaBrigitte:
## What does this change?
Update the versions to the latest version provided by cloudquery starting with CQ_CLI, CQ_POSTGRES_DESTINATION and CQ_AWS, as suggested on the [trello card](https://trello.com/c/4CgUsoJm/469-update-cloudquery-plugins)

## What is still unclear
For CQ_CLI the comment states: If you change this value remember to produce a new image by running the "CQ image" workflow. I tried to run it on the branch but it failed.

## Why?
We have recently seen some failures syncing AWS data. Let’s update the relevant plugins in case we are missing out on some fixes that will help with this.

## When to deploy?
We should first switch to an allow list to avoid collecting more tables than we bargained for. [Wait for this PR](https://github.com/guardian/service-catalogue/pull/1639) to be done before the update.

### DIFF
--- a/.env
+++ b/.env
@@ -5,16 +5,16 @@
 # NOTE: If you change this value remember to produce a new image by running the "CQ image" workflow.
 # See https://github.com/guardian/service-catalogue/actions/workflows/cq-image.yml
 # See https://github.com/cloudquery/cloudquery/releases?q=cli
-CQ_CLI=6.29.0
+CQ_CLI=6.29.7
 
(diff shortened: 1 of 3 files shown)


akash1810:
> For CQ_CLI the comment states: If you change this value remember to produce a new image by running the "CQ image" workflow. I tried to run it on the branch but it failed.

Opting to split this PR up following this. https://github.com/guardian/service-catalogue/pull/1685 updates the AWS source plugin.

PR #1639, opened 2025-09-17
Use allow list for remaining tables that need filtering

JuliaBrigitte:
We decided to split the [move to allow list PR](https://github.com/guardian/service-catalogue/pull/1601) into smaller steps to be better able to check the impact of the changes

## What does this change?
Collect all remaining tables lists that use wildcards that need filtering out of a larger allow list with new allow lists

## Why?
Cloudquery bills by row collected and frequently add new tables in updates so we want to move to an allow list based approach to manage the risk of collecting more tables than our allowance allows us to.


packages/cdk/lib/cloudquery/allow-list-tables/README.md
```diff
@@ -0,0 +1,14 @@
...
+## Collected Tables
+This directory contains the collected tables for CloudQuery.
+
+### History
+We started with cloudquery by collecting most data that CloudQuery offers. CloudQuery often add new tables in new
+releases. This meant our usage could dramatically increase just by changing a version number. So we decided to switch to
+an allow list approach, where we explicitly list the tables we want to collect.
+
+### Tables we have chosen not to collect early on
+The skipTables array includes tables we are skipping because they are slow and or uninteresting to us.
```
tjsilver, line 14: Nit: ```suggestion
The `skipTables array includes tables we are skipping because they are slow and/or of little value to us.
```

tjsilver (approved):
LGTM, assuming the filtering mechanism has already been tested locally or on CODE.

PR #1641, opened 2025-09-22
record all vulnerabilities, only notify for runtime ones

NovemberTang:
## What does this change?

All critical and high-level vulnerabilities are recorded in the `repocop_vulnerabilities` table, but only runtime ones trigger notification.


kelvin-chappell (approved):
📈 👍

PR #1643, opened 2025-09-24
Increase frequency for cloudwatch table

jorgeazevedo:
## What does this change?

This makes the scheduling of `aws_cloudwatch_alarms` match `aws_elbv1_load_balancers` and `aws_elbv2_load_balancers`.

## Why?

We get alerts for when a cloudwatch alarm references a non-existing LB. The frequency of the load balancer tables means we can detect when a LB disappears quite quickly, but detecting the fixed cloudwatch alarm only happens over night.

## How has it been verified?
I think this change is low risk from a code perspective, so I haven't validated it.

The question that worries me is if the increased row consumption is going to be a problem? The table currently has `3520` rows, so with this change I think it means 168,960 rows a day or ~5M/month?

### DIFF
--- a/packages/cdk/lib/cloudquery/index.ts
+++ b/packages/cdk/lib/cloudquery/index.ts
@@ -259,7 +259,7 @@ export function addCloudqueryEcsCluster(
 			name: 'AwsOrgWideCloudwatchAlarms',
 			description:
 				'Collecting CloudWatch Alarm data across the organisation. Uses include building SLO dashboards.',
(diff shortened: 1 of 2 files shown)


JuliaBrigitte (approved):

packages/cdk/lib/cloudquery/index.ts
akash1810, line 1: We've a few tasks with a description of `Collecting X data across the organisation. Uses include building SLO dashboards.`. Should we combine them into a single task; if we're collecting them for the same purpose, I think it makes sense to collect them at the same rate?

PR #1647, opened 2025-10-01
feat: add relevant languages to ignored list 

tjsilver:
## What does this change?

Adds various irrelevant languages (no dependencies to track, or misattributed languages that don't exist in the repo)  to the ignored and supported languages lists for Dependabot. This will ensure that we don't misclassify repos as not having vulnerability tracking when they actually do for any relevant languages. 

## Why?

This will help us identify repos that are genuinely not supported by Dependabot/dependency submission.

## How has it been verified?

When this PR is merged and repocop runs, we should see the list of repos without vulnerability tracking go down.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -42,3 +42,7 @@ updates:
     directory: '/'
     schedule:
       interval: 'monthly'
+    groups:
+      all:
+        patterns:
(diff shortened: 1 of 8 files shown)


kelvin-chappell (approved):
👍

PR #1664, opened 2025-10-03
Fix: update ignored languages

tjsilver:
## What does this change?

Fixes a typo in RenderScript which is case-sensitive.
Adds ObjectScript to the ignore list.

## Why?
Repos were still being reported as not having vulnerability tracking based on these languages which should be ignored.


kelvin-chappell (approved):
👍

PR #1665, opened 2025-10-03
Move remaining tables to allow list and test task size

JuliaBrigitte:
## What does this change?
Collect all remaining tables lists that use wildcards that need filtering out of a larger allow list.

Previously, collecting all remaining AWS tables in a single catch-all task caused the ECS task definition to exceed the maximum allowed size.

We tried to write a test that checks the task against the task definition, but this proved difficult, since the [task definition](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definitions.html) contains much more than only the list of tables to collect.  

## Test

A test was added to check the number of tables in each Cloudquery task, helping to prevent future task definition size issues.

Through multiple deployments, we determined an upper limit for the number of tables per task.
The current implementation splits AwsRemainingData into two tasks, which deploy successfully.

## Why?
Cloudquery bills by row collected and frequently add new tables in updates so we want to move to an allow list based approach to manage the risk of collecting more tables than our allowance allows us to.

## Background
This work is part of a broader migration to allow lists, split into smaller PRs for easier review and impact assessment. We decided to split the [move to allow list PR](https://github.com/guardian/service-catalogue/pull/1601) into smaller steps to be better able to check the impact of the changes.


packages/cdk/lib/cloudquery/cluster.ts
```diff
@@ -142,11 +147,15 @@ interface CloudqueryClusterProps extends AppIdentity {
...
+	tasks: ScheduledCloudqueryTask[];
+	sources: CloudquerySource[];
+	//aws-cdk-lib.aws_ecs.ClusterProps#containerInsights is deprecated.
+	//moved to containerInsightsV2
 	constructor(scope: GuStack, id: string, props: CloudqueryClusterProps) {
 		super(scope, id, {
 			vpc: props.vpc,
 			enableFargateCapacityProviders: true,
-			containerInsights: true,
+			containerInsightsV2: ContainerInsights.ENABLED,
```
jacobwinch, line 158: Minor point: do we need to make this change as part of this PR? It seems unrelated to me.

packages/cdk/lib/cloudquery/cluster.ts
```diff
@@ -142,11 +147,15 @@ interface CloudqueryClusterProps extends AppIdentity {
  * created in the private subnets of the VPC provided.
  */
 export class CloudqueryCluster extends Cluster {
+	tasks: ScheduledCloudqueryTask[];
+	sources: CloudquerySource[];
+	//aws-cdk-lib.aws_ecs.ClusterProps#containerInsights is deprecated.
```
jacobwinch, line 152: I think we could drop this comment, but it might be worth adding a brief note on this change in the PR description.

packages/cdk/lib/cloudquery/index.ts
```diff
@@ -251,17 +253,17 @@ export function addCloudqueryEcsCluster(
...
+				tables: filterAllowedTables(awsTables, [/^aws_ssm_parameters$/]),
 			}),
 			policies: [listOrgsPolicy, cloudqueryAccess('*')],
 		},
 		{
 			name: 'AwsOrgWideCloudwatchAlarms',
 			description:
 				'Collecting CloudWatch Alarm data across the organisation. Uses include building SLO dashboards.',
-			schedule: Schedule.rate(Duration.minutes(30)),
+			schedule: Schedule.cron({ minute: '0', hour: '2' }),
```
jacobwinch, line 264: I think we recently changed this to run every 30-minutes - https://github.com/guardian/service-catalogue/pull/1643 - is changing the schedule here intentional?

packages/cdk/lib/cloudquery/cluster.ts
JuliaBrigitte: You are right, it was just an annoying error message, when I wrote the tests. Will put it into another PR.

JuliaBrigitte:
> (Discussed over chat, adding here for prosperity) I think the testing added here is too domain specific. I think we could simplify the tests quite significantly by testing the size of a TaskDefinition resource. Maybe something like:
> 
> ```ts
> const template = Template.fromStack(stack);
> const taskDefinitions = template.findResources('AWS::ECS::TaskDefinition');
> // assertions against taskDefinitions here
> ```
> 
> Happy to pair.

I tried to do this and failed, so would definitely need to pair on this.

packages/cdk/lib/cloudquery/cluster.ts
JuliaBrigitte: https://github.com/guardian/service-catalogue/pull/1667

packages/cdk/lib/cloudquery/index.ts
JuliaBrigitte: Thank you for catching this!

packages/cdk/lib/cloudquery/index.ts
```diff
@@ -365,16 +372,55 @@ export function addCloudqueryEcsCluster(
...
+		remainingAwsTables.length / 2,
+	);
+
+	const remainingAwsSourcesPart1: CloudquerySource = {
+		name: 'AwsRemainingDataPart1',
+		description: 'Data fetched across all accounts in the organisation part 1.',
+		schedule: Schedule.cron({ minute: '0', hour: '16', weekDay: 'SAT' }), // Every Saturday, at 4PM UTC
+		config: awsSourceConfigForOrganisation({
+			tables: remainingAwsTables.slice(0, halfOfRemainingAwsTablesNumber),
+			//ServiceCatalogue collects child tables automatically.
```
akash1810, line 395:
```suggestion
			// CloudQuery collects child tables automatically.
```

I'm not sure this comment is too useful here - might it be better as a doc string for the [`skipTables` property](https://github.com/guardian/service-catalogue/blob/0ad898e32d9160ba31f4d4689e2b7a1342cd09f2/packages/cdk/lib/cloudquery/config.ts#L17)?

packages/cdk/lib/cloudquery/index.ts
akash1810, line 431: Should this list also include the tables collected in `remainingAwsSourcesPart1`?

packages/cdk/lib/cloudquery/index.ts
```diff
@@ -44,7 +45,7 @@ import {
 	securityHubTableOptions,
 } from './table-options';
 
-interface CloudqueryEcsClusterProps {
+export interface CloudqueryEcsClusterProps {
```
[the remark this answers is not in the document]
JuliaBrigitte: That was for the now rewritten test, removed it.

PR #1808, opened 2026-01-27
feat!: Drop `guardian_non_p_and_e_github_teams` table 

akash1810:
## What does this change?
The contents of `guardian_non_p_and_e_github_teams` is manually maintained. This change removes it in favour of querying the `github_teams_tree` view introduced in https://github.com/guardian/service-catalogue/pull/1807.

Whilst this is a breaking change, there are no [dashboards using `guardian_non_p_and_e_github_teams`](https://metrics.gutools.co.uk/goto/vN9t4WNvR?orgId=1). That is, no comms are needed.

> [!NOTE]
> This change increases the number of teams [from 18 to 107](https://metrics.code.dev-gutools.co.uk/goto/8Qzs47NDg?orgId=1).

## How has it been verified?
TBD.


NovemberTang (commented):
Discussed with @akash1810 . This doesn't take into account historic teams that still are the sole owners of repos, so I think we'd need to take another look at how the org has been structured before we could swap these out!

PR #1835, opened 2026-02-24
chore(postgres): bump postgres to v17

tjsilver:
## What does this change?

This change bumps postgres from v16 (see #1796) to v17.9. Follows #1868 which prepares the database for the upgrade and should be merged after that.

To ensure we're up-to-date and making use of security patches and performance enhancements.

See release notes for more info: 
17.0: https://www.postgresql.org/docs/release/17.0/
17.9 (latest v17): https://www.postgresql.org/docs/release/17.9/

The next step will be to migrate to postgres v18 which is the latest.

## How has it been verified?

- [ ] Tested on CODE (after approval only)

### DIFF
--- a/packages/cdk/lib/cloudquery/images.ts
+++ b/packages/cdk/lib/cloudquery/images.ts
@@ -14,7 +14,7 @@ export const Images = {
 		`ghcr.io/guardian/cq-source-ns1:${CloudQueryPluginVersions.CloudqueryNs1}`,
 	),
 	postgres: ContainerImage.fromRegistry(
-		'public.ecr.aws/docker/library/postgres:16-alpine',
+		'public.ecr.aws/docker/library/postgres:17-alpine',
 	),
 	/**
 	 * This image is built in CI by the service-catalogue, and tagged with the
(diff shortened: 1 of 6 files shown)


tjsilver:
CI is failing with the same error as for the attempt to bump to v18. Have put this into draft until the issue is resolved.

packages/cdk/lib/cloudquery/images.ts
```diff
@@ -14,7 +14,7 @@ export const Images = {
 		`ghcr.io/guardian/cq-source-ns1:${CloudQueryPluginVersions.CloudqueryNs1}`,
 	),
 	postgres: ContainerImage.fromRegistry(
-		'public.ecr.aws/docker/library/postgres:16-alpine',
+		'public.ecr.aws/docker/library/postgres:17-alpine',
```
akash1810, line 17: Similar to how we do in `ci.yml`, should we pin this image to a specific digest too?

tjsilver: We discussed this and agreed that this could be done in a follow-up PR

akash1810 (approved):

PR #1836, opened 2026-02-24
fix(data-audit): Explicitly handle `AccessDeniedException` errors

akash1810:
## What does this change?

Some accounts have an SCP denying account access in some regions. This manifests in the lambda failing and the log line:

```log
"AccessDeniedException: User: arn:aws:sts::<REDACTED>:assumed-role/cloudquery-access/aws-sdk-js-1771949621302 is not authorized to perform: lambda:ListFunctions on resource: * with an explicit deny in a service control policy"
```

If we encounter an `AccessDeniedException`, log it and skip that account/region combination. This should mean the lambda doesn't exit early.

## How has it been verified?
After [deploying to CODE](https://riffraff.gutools.co.uk/deployment/view/c4b765bc-effa-4cf9-b6d0-e4ef635d41a5) and invoking the lambda, the following logs can be seen:

TBD.

And the lambda's metrics show successful invocation:

TBD.

### DIFF
--- a/packages/data-audit/src/audit/aws-lambda.ts
+++ b/packages/data-audit/src/audit/aws-lambda.ts
@@ -61,7 +61,27 @@ async function numberOfLambdaFunctionsFromAws(
 ): Promise<number> {
 	const promises = accounts.flatMap((account) => {
 		return regions.map((region) => {
-			return numberOfLambdaFunctionsForAwsAccount(stage, account.Id!, region);
+			try {
+				return numberOfLambdaFunctionsForAwsAccount(stage, account.Id!, region);
+			} catch (e) {
+				// Some accounts have an SCP denying account access in some regions.
+				// On AccessDeniedException (403), log and skip that account/region combination.
+				if (e instanceof Error && e.name === 'AccessDeniedException') {
+					console.error(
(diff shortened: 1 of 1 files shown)


packages/data-audit/src/audit/aws-lambda.ts
```diff
@@ -61,7 +61,27 @@ async function numberOfLambdaFunctionsFromAws(
 ): Promise<number> {
 	const promises = accounts.flatMap((account) => {
 		return regions.map((region) => {
-			return numberOfLambdaFunctionsForAwsAccount(stage, account.Id!, region);
+			try {
+				return numberOfLambdaFunctionsForAwsAccount(stage, account.Id!, region);
+			} catch (e) {
+				// Some accounts have an SCP denying account access in some regions.
+				// On AccessDeniedException (403), log and skip that account/region combination.
+				if (e instanceof Error && e.name === 'AccessDeniedException') {
```
akash1810: TIL about [`Error.isError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/isError) available in Node 24+.

PR #1868, opened 2026-03-25
fix(migrations): make compatible with v17 postgres

tjsilver:
## What does this change?

Updates existing migrations to allow for upgrade from postgres v16 to v17.

Currently, we cannot upgrade to v17 of postgres because breaking changes have been made to the migrations.

Modifying existing migrations, while not ideal, is the only way to apply the patches that works in CI.

Release notes: https://www.postgresql.org/docs/release/17.0/, in particular:
- Change functions to use a safe [search_path](https://www.postgresql.org/docs/17/runtime-config-client.html#GUC-SEARCH-PATH) during maintenance operations (Jeff Davis) [§](https://postgr.es/c/2af07e2f7) [§](https://postgr.es/c/b4da732fd64)
- Remove server variable old_snapshot_threshold (Thomas Munro) [§](https://postgr.es/c/f691f5b80) 

## How has it been verified?

- [x] Ran the cli migrate command on the CODE DB successfully. 
- [x] Deployed to CODE and observed the prisma migrate task complete successfully. 
- [x] Deployed `main` to CODE to ensure the prisma migrate task still completes successfully.

### DIFF
--- a/packages/common/prisma/migrations/20240412103300_aws_resources/migration.sql
+++ b/packages/common/prisma/migrations/20240412103300_aws_resources/migration.sql
@@ -131,6 +131,7 @@ SET search_path FROM CURRENT;
 DROP INDEX IF EXISTS idx_aws_resources_account_id;
 DROP INDEX IF EXISTS idx_aws_resources_arn;
 DROP VIEW IF EXISTS aws_resources;
+DROP MATERIALIZED VIEW IF EXISTS aws_resources;
 CREATE MATERIALIZED VIEW aws_resources AS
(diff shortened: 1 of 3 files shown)


akash1810:
> breaking changes have been made to the migrations

Are there any release notes we can link to that detail these breaking changes?

akash1810 (approved):

tjsilver:
> > breaking changes have been made to the migrations
> 
> Are there any release notes we can link to that detail these breaking changes?

Yes, I have updated the description above.

PR #1874, opened 2026-04-01
Move prisma generated code to own directory

NovemberTang:
## What does this change?

Moves generated prisma code out of src, and into its own directory

To split hand-written and generated code.

## How has it been verified?

- [x] deploy to CODE and test lambdas
- [x] verify CLI tool still works

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,16 +1,18 @@
 {
-  "extends": "@guardian/tsconfig/tsconfig.json",
-  "compilerOptions": {
-    "esModuleInterop": true,
-    "baseUrl": ".",
-    "paths": {
-      "common": ["packages/common/src"],
-      "common/*": ["packages/common/src/*"],
-      "cloudquery-tables": ["packages/cloudquery-tables/src"],
-      "cloudquery-tables/*": ["packages/cloudquery-tables/src/*"]
-    },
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -1,16 +1,18 @@
...
-  }
+	"extends": "@guardian/tsconfig/tsconfig.json",
+	"compilerOptions": {
+		"esModuleInterop": true,
+		"baseUrl": ".",
+		"paths": {
+			"common": ["packages/common/src"],
+			"common/*": ["packages/common/src/*"],
+			"common/prisma-client": ["packages/common/prisma-client"],
+			"common/prisma-client/*": ["packages/common/prisma-client/*"],
```
akash1810, line 10: This is a little confusing - above we have `common/*` why wouldn't these include the listed `prisma-client` paths? To reduce confusion, could we change the paths to:

```json
"prisma-client": ["packages/common/prisma-client"],
"prisma-client/*": ["packages/common/prisma-client/*"],
```

NovemberTang: We don't have `common/*`, we have `common/src/*`, and so the generated `common/prisma-client/*` needs to be handled separately.

tjsilver (approved):
Good idea!

PR #1875, opened 2026-04-01
fix: update prisma in migrate container to match

tjsilver:
## What does this change?

Updates prisma to v7 to match the recent upgrade.


tjsilver:
We now bundle prisma so don't need to specify the version here anymore.

PR #1883, opened 2026-04-02
feat: create prisma-migrate lambda to replace ecs task

tjsilver:
## What does this change?
Creates a new Lambda-based workflow for running Prisma database migrations, to replace the current ECS task, if we choose to.

* Adds a new `prisma-migrate` package under `lambda-runtime`.
* Updates the build script to add a `createPrismaLambdaZip` function, which builds and packages the Lambda runtime artifact with all necessary files and dependencies.
* Modifies `dependabot.yml` to add automated dependency updates for the new `prisma-migrate` package.
* Disables the existing event rule for the ECS-based Prisma migrate task. 
* TODO: remove the ECS prisma-migrate code in a subsequent PR.

Currently the Dockerfile requires a version of prisma to be specified and currently this needs to be updated manually. With the recent upgrade of prisma to v7, this task broke as the prisma version is still on v5.

The approach was chosen to allow the prisma version to be kept up-to-date with that used in `common`, by being updateable by Dependabot. It is up to the team to decide if this migration to Lambda is ultimately worth it.

## How has it been verified?

Tested the lambda on CODE and saw the prisma command run successfully (migration failed, but for other reasons).


tjsilver:
Closing as we fixed the ECS task instead.

PR #1892, opened 2026-04-15
fix: update prisma version to match client

tjsilver:
## What does this change?

Updates the version of prisma in the prisma-migrate task to the same version as used in `packages/common`.

The prisma-migrate task is currently broken.

This is a quick fix while we look at migrating to a lambda  (see #1883 ) or find a way to keep the versions synced across ecs and the prisma client.


jorgeazevedo (approved):
Let's give this a try!

PR #1893, opened 2026-04-15
Upgrade node version from 20 to 22

JuliaBrigitte:
## What does this change?

Upgrade Lambdas using nodejs20.x to 22.

We should upgrade Lambdas using nodejs20.x before AWS blocks updates to functions which use the nodejs20.x runtime (Jul 1, 2026). At this point using this runtime becomes risky because deploying these applications will no longer be possible. This means that:

We cannot fix bugs or reliability issues affecting users

We will find it much harder to meet other security obligations, such as [fixing critical dependency vulnerabilities within 2 days](https://docs.google.com/document/d/1Q_ts5mbRvASCGo3UfsPHwUVvSXc3f2fkK2Fwp6N9Pxc/edit?tab=t.0#bookmark=id.vogwkng6po4g)

We lose our ability to rollback to the old runtime if the app behaves unexpectedly after an attempted upgrade

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -22417,7 +22417,7 @@ spec:
             "Arn",
           ],
         },
-        "Runtime": "nodejs20.x",
+        "Runtime": "nodejs22.x",
         "Tags": [
           {
             "Key": "App",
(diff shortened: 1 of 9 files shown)


JuliaBrigitte:
Close in preference for a lambda by lambda approach starting with https://github.com/guardian/service-catalogue/pull/1912

PR #1894, opened 2026-04-16
Rename dependabot group from all to gh-actions

NovemberTang:
I think the name `all` is misleading, as when I see it pop up, I assume it bumps all dependencies, not just GHA related ones, so I've renamed it


jorgeazevedo (approved):
Good idea!

PR #1895, opened 2026-04-16
fix: get the ECS container running on Prisma v7

tjsilver:
## What does this change?

Updates the Prisma migration container and build process:
* removes the direct installation of Prisma in the Dockerfile
* bundles the Prisma CLI with the schema during packaging
* removes the installation of the Prisma NPM package from the `prisma-migrate` Dockerfile
* adds a new `prisma.config.ts` file that defines the Prisma schema location and datasource URL using env vars
* updates the `build.sh` script to include the Prisma CLI 

This addresses two issues. 

1. the migration task currently fails with `Error: The datasource.url property is required in your Prisma config file when using prisma migrate deploy.`
2. The Dockerfile required its own version of Prisma which would have to be kept up-to-date manually and could easily fall out of sync with the version used on the database.

This is one of two approachs to solving this issue. The other is to migrate the task to a lambda: #1883 

## How has it been verified?

Build the container locally and ran it against a test db. 

Before: 
`Error: The datasource.url property is required in your Prisma config file when using prisma migrate deploy.`

After:
`All migrations have been successfully applied.`

### DIFF
--- a/containers/prisma-migrate/prisma.config.ts
+++ b/containers/prisma-migrate/prisma.config.ts
@@ -1,0 +1,9 @@
+import path from 'node:path';
+import { defineConfig } from 'prisma/config';
+
+export default defineConfig({
+	schema: path.join('prisma', 'schema.prisma'),
+	datasource: {
+		url: process.env.DATABASE_URL as string,
+	},
+});
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -58,6 +58,11 @@ createPrismaZip() {
     cd "$ROOT_DIR/packages/common"
     zip -qr prisma.zip ./prisma
   )
+  echo "Adding Prisma CLI to prisma zip"
+  (
+    cd "$ROOT_DIR"
+    zip -qr "$ROOT_DIR/packages/common/prisma.zip" node_modules/prisma node_modules/@prisma
+  )
 }
 
 verify() {
--- a/containers/prisma-migrate/Dockerfile
+++ b/containers/prisma-migrate/Dockerfile
@@ -6,14 +6,10 @@ WORKDIR /usr/src/app
 # This is used to fetch the prisma directory from S3
 RUN apk --no-cache add aws-cli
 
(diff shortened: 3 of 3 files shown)


containers/prisma-migrate/prisma.config.ts
```diff
@@ -0,0 +1,9 @@
+import path from 'node:path';
+import { defineConfig } from 'prisma/config';
+
+export default defineConfig({
+	schema: path.join('prisma', 'schema.prisma'),
+	datasource: {
+		url: process.env.DATABASE_URL as string,
+	},
+});
```
jorgeazevedo, line 9: Is this needed to pass the datasource to prisma as a consequence of the upgrade to v7?

scripts/build.sh
```diff
@@ -58,6 +58,11 @@ createPrismaZip() {
     cd "$ROOT_DIR/packages/common"
     zip -qr prisma.zip ./prisma
   )
+  echo "Adding Prisma CLI to prisma zip"
+  (
+    cd "$ROOT_DIR"
+    zip -qr "$ROOT_DIR/packages/common/prisma.zip" node_modules/prisma node_modules/@prisma
```
jorgeazevedo, line 64: Does this zip file need to be unzipped in the Docker container to be able to call the CLI?

containers/prisma-migrate/prisma.config.ts
tjsilver: Yes, that's right.

jorgeazevedo (approved):
A few questions to for my own understanding, but up for giving this a try in prod!

scripts/build.sh
tjsilver: Yes it's unzipped in [`run-prisma-migrate.sh`](https://github.com/guardian/service-catalogue/blob/35b4d967e0a58468931f513cf3208226d2a871c8/containers/prisma-migrate/run-prisma-migrate.sh#L14). This is the same as before.

PR #1898, opened 2026-04-17
fix: Update prisma migrate command to use the correct binary path

tjsilver:
## What does this change?

* Changes the command in `run-prisma-migrate.sh` to use `prisma/build/index.js` for running `migrate deploy` instead of the CLI binary.

To fix the error in the last run: `Error: ENOENT: no such file or directory, open '/usr/src/app/prisma/node_modules/.bin/prisma_schema_build_bg.wasm'`

### DIFF
--- a/containers/prisma-migrate/run-prisma-migrate.sh
+++ b/containers/prisma-migrate/run-prisma-migrate.sh
@@ -18,4 +18,4 @@ DB_PORT=5432
 export DATABASE_URL=postgres://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/postgres
 
 echo 'Running prisma migrate deploy'
-"${ROOT_DIR}/prisma/node_modules/prisma/build/index.js" migrate deploy --config "${ROOT_DIR}/prisma.config.ts"
+node "${ROOT_DIR}/prisma/node_modules/prisma/build/index.js" migrate deploy --config "${ROOT_DIR}/prisma.config.ts"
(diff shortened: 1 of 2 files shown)


containers/prisma-migrate/run-prisma-migrate.sh
```diff
@@ -18,4 +18,4 @@ DB_PORT=5432
 export DATABASE_URL=postgres://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/postgres
 
 echo 'Running prisma migrate deploy'
-"${ROOT_DIR}/prisma/node_modules/.bin/prisma" migrate deploy --config "${ROOT_DIR}/prisma.config.ts"
+"${ROOT_DIR}/prisma/node_modules/prisma/build/index.js" migrate deploy --config "${ROOT_DIR}/prisma.config.ts"
```
jorgeazevedo: Have you been able to test this somehow? I'd be surprised if running a .js file directly works, without it being invoked by `node` for example.

It seems to me that there are missing files from `.bin`, so maybe we should be copying `.bin` directory here? https://github.com/guardian/service-catalogue/blob/5d7f102eb0023397f133b7e75a82a1a22ec7e035/scripts/build.sh#L64

tjsilver: Ah you're right. We don't need the .bin but we do need to add node to invoke the function! Have removed the .bin [here](https://github.com/guardian/service-catalogue/pull/1898/commits/b31e52fe56e1ee71c6a97660a3ff9575ae7b61e9) and added node [here](https://github.com/guardian/service-catalogue/pull/1898/commits/13ade8aad5f45c60a7c5f0c5a357c4c7780a22a9).

jorgeazevedo (approved):
Let's give this a try!

PR #1911, opened 2026-04-21
Group non-breaking npm updates not included in other dependabot groups

NovemberTang:
## What does this change?

Adds a further group to npm dependabot updates, grouping all minor and patch versions not covered by previous groups

Reducing PR overhead


kelvin-chappell (approved):
👍

PR #1923, opened 2026-04-24
Update the cloudquery usage lambda to node version 22

JuliaBrigitte:
## What does this change?

Upgrade Lambda using nodejs20.x to 22.

## How can this be tested

---
  - https://app.asana.com/0/0/1214114330377302


JuliaBrigitte:
Decided to upgrate to node 24 instead

PR #1924, opened 2026-04-24
Update data-audit lambda to note version 22

JuliaBrigitte:
## What does this change?

Upgrade Lambda using nodejs20.x to 22.

## How can this be tested
Run in CODE (failed)

---
  - https://app.asana.com/0/0/1214114330421843


JuliaBrigitte:
Testing this with Akash led us to an aws account (outside P&E) that was not completely set up and therefore had not granted permission to service catalogue. In the end we decided, that since this lambda was last run in 10/2025 it is no longer needed.

PR #1926, opened 2026-04-27
fix: AWS credentials validation in migration test script

tjsilver:
## What does this change?

Adds another condition to the AWS credentials check function for the `InvalidClientTokenId` case to the prisma migration test script.

When I ran the script locally before fetching credentials, the container failed to start but I got the message "AWS credentials are valid" in the terminal output, which was confusing. When I echoed out the status, it was clear that it was an `InvalidClientTokenId` error.

## How has it been verified?

I ran the script again and observed the error message being output.


kelvin-chappell (approved):
👍

PR #1927, opened 2026-04-28
fix: add volume for prisma engine and add writeable mount point

tjsilver:
## What does this change?

* Adds a new prisma binary volume to the `addPrismaMigrateTask` task definition.

* Mounts the new `engine-volume` at `/tmp` in the container, allowing the Prisma engine binary to be written at runtime.

Currently the migration container doesn't work - it fails with the error: `Error: EROFS: read-only file system, mkdir '/tmp/xxx'`

## How has it been verified?

I first recreated the error locally by making the test container (`/packages/dev-environment/docker-compose-prisma-migrate.yaml`) read-only, then mirrored the fix as below and observed the migrations running successfully.
```read_only: true
    tmpfs:
      - /usr/src/app/prisma:exec
      - /tmp
```

I have not committed these changes to the test container, but maybe they should be added, to keep the two more in line. WDYT?

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -27034,6 +27034,11 @@ spec:
                 "ReadOnly": false,
                 "SourceVolume": "artifact-volume",
               },
+              {
+                "ContainerPath": "/tmp",
+                "ReadOnly": false,
+                "SourceVolume": "engine-volume",
+              },
             ],
             "Name": "prisma-migrate-taskContainer",
             "ReadonlyRootFilesystem": true,
@@ -27134,6 +27139,9 @@ spec:
           },
           {
             "Name": "artifact-volume",
+          },
+          {
+            "Name": "engine-volume",
           },
         ],
       },
--- a/packages/cdk/lib/prisma-migrate-task.ts
+++ b/packages/cdk/lib/prisma-migrate-task.ts
@@ -160,13 +160,24 @@ export function addPrismaMigrateTask(
 	const prismaArtifactVolume: Volume = {
 		name: 'artifact-volume',
 	};
+	const prismaEngineVolume: Volume = {
+		name: 'engine-volume',
+	};
 
 	taskDefinition.addVolume(prismaArtifactVolume);
+	taskDefinition.addVolume(prismaEngineVolume);
 
 	prismaTask.addMountPoints({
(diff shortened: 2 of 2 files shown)


akash1810:
Has this been tested on CODE?

tjsilver:
> Has this been tested on CODE?

No, because IIUC it is hard to test on CODE because we would need to publish the container, and this is only done on merge to main currently. Hence testing locally as best I could with emulation. If there's something I've missed though on how to test on CODE please let me know!

akash1810:
> > Has this been tested on CODE?
> 
> No, because IIUC it is hard to test on CODE because we would need to publish the container, and this is only done on merge to main currently. Hence testing locally as best I could with emulation. If there's something I've missed though on how to test on CODE please let me know!

It looks like the workflow has a `workflow_dispatch` trigger allowing for manual invocation: https://github.com/guardian/service-catalogue/blob/dacc4c67ef93207026912097b18c5f85465a2006/.github/workflows/prisma-migrate-image.yml#L17-L18

To test on CODE, I think we'd:
1. Create a new image
2. Identify a tag on the new image and update https://github.com/guardian/service-catalogue/blob/dacc4c67ef93207026912097b18c5f85465a2006/packages/cdk/lib/cloudquery/images.ts#L19-L29
3. Deploy to CODE and observe
4. Revert 2 as it doesn't need to be merged to `main`

akash1810 (approved):

tjsilver:
> > > Has this been tested on CODE?
> > 
> > 
> > No, because IIUC it is hard to test on CODE because we would need to publish the container, and this is only done on merge to main currently. Hence testing locally as best I could with emulation. If there's something I've missed though on how to test on CODE please let me know!
> 
> It looks like the workflow has a `workflow_dispatch` trigger allowing for manual invocation:
> 
> https://github.com/guardian/service-catalogue/blob/dacc4c67ef93207026912097b18c5f85465a2006/.github/workflows/prisma-migrate-image.yml#L17-L18
> 
> To test on CODE, I think we'd:
> 
>     1. Create a new image
> 
>     2. Identify a tag on the new image and update https://github.com/guardian/service-catalogue/blob/dacc4c67ef93207026912097b18c5f85465a2006/packages/cdk/lib/cloudquery/images.ts#L19-L29
> 
>     3. Deploy to CODE and observe
> 
>     4. Revert 2 as it doesn't need to be merged to `main`

Tested on CODE and the migration task ran successfully.

PR #1930, opened 2026-04-28
Upversion node for obligatron

aug24:
## What does this change?

This change upversions node.

Node is currently specified at 22 in .nvmrc, which applies to the entire repo, and 22 in obligatron in this PR, and  interactive-monitor and github-actions-usage.  The remainder are all 20.

We should discuss if everything should be upversioned in one go


akash1810:
I think a change to the `build` script in `package.json` is also needed. See also https://github.com/guardian/service-catalogue/pull/1920/changes.

Has this been tested on CODE?

tjsilver (commented):
Has this been replaced by #1934?

PR #1931, opened 2026-04-28
fix: bundle package.json into dist for dependency graph integrator

tjsilver:
## What does this change?

* Adds a `postbuild` script to copy `package.json` into the `dist` directory after building, to ensure the lambda knows this is ESM.
* Adds a `--banner:js` flag to inject `createRequire` from the Node.js `module` package to resolve a compatibility error.

The lambda has been failing with the error `ERROR","message":"(node:2) Warning: To load an ES module, set \"type\": \"module\" in the package.json or use the .mjs extension.`

## How has it been verified?

- [x] Tested on CODE

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -25244,7 +25244,7 @@ spec:
             "Arn",
           ],
         },
-        "Runtime": "nodejs20.x",
+        "Runtime": "nodejs24.x",
         "Tags": [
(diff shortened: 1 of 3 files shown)


akash1810:
Do we know why the other lambdas do not require this?

tjsilver:
> Do we know why the other lambdas do not require this?

Yes, they all have the `postbuild` script that adds the `package.json` to `dist`. I have now added that to this package, so we don't need to emit the `.mjs` explicitly.

kelvin-chappell (approved):
👍

PR #1932, opened 2026-04-28
Harden the bash script for running the dev environment

JuliaBrigitte:
## What does this change?

Harden the bash script for running the dev environment after problems with some versions causing it to exit.

To allow the setup script to work with more versions of bash.

## How has it been verified?

Ran on two different developer machines.


packages/dev-environment/script/start
```diff
@@ -10,9 +10,10 @@ LOCAL_ENV_FILE=$HOME/.gu/service_catalogue/.env.local
 clear='\033[0m'
 cyan='\033[0;36m'
 yellow="\033[1;33m"
+STEP_COUNT=0
 
 step() {
-  ((STEP_COUNT++))
+  STEP_COUNT=$((STEP_COUNT+1))
```
NovemberTang, line 18:
```suggestion
  ((++STEP_COUNT))
```
This would also work and is a bit more concise

scripts/setup.sh
```diff
@@ -13,7 +13,7 @@ ROOT_DIR=$(realpath "${DIR}/..")
 STEP_COUNT=0
 
 step(){
-  ((STEP_COUNT++))
+  STEP_COUNT=$((STEP_COUNT+1))
```
NovemberTang, line 18:
```suggestion
  ((++STEP_COUNT))
```
same suggestion

packages/dev-environment/README.md
```diff
@@ -39,7 +39,7 @@ It includes:
    ./packages/dev-environment/script/start
    ```
 
-   This will start the Docker containers, and CloudQuery will start collecting data.
+   This will create an sql directory, start the Docker containers, and CloudQuery will start collecting data.
```
NovemberTang: Could we add a few words about what goes in this directory?

NovemberTang (approved):
A couple of non-blocking comments, looks good overall!

packages/dev-environment/script/start
JuliaBrigitte: @NovemberTang This is exactly what did not work on my machine. ((STEP_COUNT++)). In my bash version that had an exit code of 0 which in conjunction with set -e stopped the script in my bash version. @tjsilver has an explanation saved about why that made sense and will add to this PR. I will add a comment to that effect.

packages/dev-environment/script/start
```diff
@@ -10,9 +10,12 @@ LOCAL_ENV_FILE=$HOME/.gu/service_catalogue/.env.local
 clear='\033[0m'
 cyan='\033[0;36m'
 yellow="\033[1;33m"
+STEP_COUNT=0
 
 step() {
-  ((STEP_COUNT++))
+  # ((STEP_COUNT++)) does return 0 for some bash versions and in combination with set -e,
+  # it can cause the script to exit with an error depending on the dev environment.
```
akash1810, line 17: I think this is more of an implementation detail suited for the PR description or commit message rather than an inline comment.

akash1810: Same applies to the comment in `scripts/setup.sh`.

packages/dev-environment/script/start
tjsilver: This is the explanation for why it might fail for some setups and not others: In Bash, (( expression )) returns exit code 1 when the result is zero (falsy). So when STEP_COUNT is 0, ((STEP_COUNT++)) increments it to 1 but the expression evaluates to the pre-increment value of 0, returning exit code 1, which causes set -e to exit the script.

This only happens on the first call to step() when STEP_COUNT=0.

packages/dev-environment/script/start
JuliaBrigitte: Well, if somebody changes it to be more concise the problem will occur again.

packages/dev-environment/script/start
NovemberTang: @JuliaBrigitte @tjsilver  I ran into the exact same issue, so I moved the ++ to the beginning of the string rather than the end, which flips the evaluation to postincrement, and resolves the error

PR #1933, opened 2026-04-28
Delete data audit lambda

JuliaBrigitte:
## What does this change?
Deletes data audit lambda

No longer needed, deleted instead of updating node version since last run 10/2025

---
  - https://app.asana.com/0/0/1214114330421843


akash1810:
Should there also be an edit to the `package-lock.json`: https://github.com/guardian/service-catalogue/blob/dacc4c67ef93207026912097b18c5f85465a2006/package-lock.json#L15412-L15420

akash1810 (approved):
Shall we also create a card to drop the `audit_results` table once the migrations are working?

JuliaBrigitte:
@akash1810 Thanks for the review, a had already created a follow up card as discussed. https://app.asana.com/1/1210045093164357/project/1214113788426225/task/1214357303046637?focus=true

PR #1934, opened 2026-04-29
Move obligatron to node 24

aug24:
## What does this change?

1) use .tool-versions to set node version to 24 for development
2) remove .nvmrc
3) Change obligatron runtime to nodejs24
4) update doco
5) switch sslMode depending on host for localhost postgres


package.json
```diff
@@ -34,5 +34,8 @@
     "tsx": "4.21.0",
     "typescript": "^5.9.3"
   },
-  "prettier": "@guardian/prettier"
+  "prettier": "@guardian/prettier",
+  "dependencies": {
```
aug24, line 38: This dependency had to be added because Prettier uses it but does not declare it, and node24 is much more strict on resolution.

tjsilver (approved):
Looks good, given that testing on CODE is successful.

packages/common/src/database-setup.ts
```diff
@@ -76,7 +76,9 @@ export async function getDatabaseConfig(
 export function getDatabaseConnectionString(config: DatabaseConfig) {
 	const { user, password, hostname } = config;
 
+	const sslmode = hostname === 'localhost' ? 'disable' : 'verify-full';
+
 	return `postgres://${user}:${encodeURIComponent(
 		password,
-	)}@${hostname}:${databasePort}/postgres?schema=public&sslmode=verify-full&connection_limit=20&pool_timeout=20`;
+	)}@${hostname}:${databasePort}/postgres?schema=public&sslmode=${sslmode}&connection_limit=20&pool_timeout=20`;
```
akash1810, line 83: I'm not sure I see the connection between this change and using Node 24; could you elaborate a little please?

akash1810 (dismissed):
Should we split this PR[^1]? I think adopting dev containers is a separate concern to updating the runtime of a lambda.

[^1]: See also https://github.com/guardian/recommendations/blob/main/pull-requests.md#small.

package.json
akash1810: Hmm, this dependency is listed by `prettier`: https://github.com/prettier/prettier/blob/2c713de00fc2265aaf34047108326f42bbca1a94/package.json#L77

What happens without this change?

aug24: CI failure.  Can't remember the exact error but it was imported without being available.

aug24:
> Should we split this PR? I think adopting dev containers is a separate concern to updating the runtime of a lambda.

This isn't dev containers.  Just moving the definition of node into tool versions.  Can split it though.

akash1810:
> This isn't dev containers. Just moving the definition of node into tool versions. Can split it though.

D'oh! Sorry, you're right, I mis-read. If we can split the move to `.tool-versions` that would be good. I think we could remove the `github:guardian/devenv 20260408-150528` line too, until we make the switch to dev containers?

packages/common/src/database-setup.ts
aug24: That was necessary to test locally.  NT's suggestion.  Don't know why is wasn't like that before. Definitely not specific to Node 24.

aug24:
Split nvmrc out to https://github.com/guardian/service-catalogue/pull/1939/changes

packages/common/src/database-setup.ts
akash1810: Did this also fail in CI? If not, that might suggest an issue with your local machine?

PR #1936, opened 2026-04-29
refactor(docs): add guidance on testing the prisma migrate container

tjsilver:
## What does this change?

Adds some documentation on how to test the prisma migration container on CODE. 

It follows on from trying to test the changes in this [PR](#1927).


JuliaBrigitte:
Great documentation, but this needing the revert is a bit hard I guess? I there any way to test is in dev instead of on CODE?

aug24 (approved):

tjsilver:
> Great documentation, but this needing the revert is a bit hard I guess? I there any way to test is in dev instead of on CODE?

You can test in DEV, but it's not possible to emulate the exact ECS container, so for the type of testing I was doing I needed to use the exact same environment. This is more for those cases where testing in DEV is not sufficient.

PR #1940, opened 2026-04-30
feat: Add monitoring for Dependency Graph Integrator

tjsilver:
## What does this change?

- Adds monitoring for the dependency graph integrator.
- Renames the lambda as it is only used by the integrator.
- Renames the props as it was slightly confusing to me that 'non-prod' props were being used in the PROD lambda.
- Fixes a typo in the repocop lambda props.
- Removes the `TEST` stage from the code as this doesn't exist.

We recently discovered that the integrator had been failing for some months and only discovered this by chance. 

This follows the pattern used by Obligatron.

## How has it been verified?

- [x] Deploy broken version of lambda to CODE (pre-fix error) to ensure it is broken!
- [x] Redeploy broken version with monitoring in base props to CODE and observe alerting (no alerting observed - maybe only happens on PROD?)
- [x] Deploy fixed version and ensure it still works with monitoring in place
- [x] After merge test broken version on PROD and observe alerting

### DIFF
--- a/packages/cdk/lib/repocop.ts
+++ b/packages/cdk/lib/repocop.ts
@@ -40,7 +40,7 @@ export class Repocop {
 			},
 		);
 
-		const repocopLampdaProps: GuScheduledLambdaProps = {
+		const repocopLambdaProps: GuScheduledLambdaProps = {
 			app: 'repocop',
 			architecture: Architecture.ARM_64,
 			fileName: 'repocop.zip',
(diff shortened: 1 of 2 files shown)


NovemberTang (commented):
Your test strategy sounds right to me, maybe temporarily move the monitoring to the base props so you can deploy to CODE?

packages/cdk/lib/repocop.ts
```diff
@@ -40,7 +40,7 @@ export class Repocop {
 			},
 		);
 
-		const repocopLampdaProps: GuScheduledLambdaProps = {
```
NovemberTang, line 43: oops!

kelvin-chappell (approved):
👍

PR #1941, opened 2026-05-01
Move Repocop to node 24

aug24:
## What does this change?

Repocop is currently running under Node 20 (LTS), which is due to EOL this year.  We are pre-emptively moving to the current LTS, 24.


packages/obligatron/package.json
```diff
@@ -4,9 +4,9 @@
 	"type": "module",
 	"scripts": {
 		"start": "APP=obligatron tsx src/run-locally.ts",
-		"build": "esbuild src/index.ts --bundle --platform=node --target=node20 --outdir=dist --external:@prisma/client --external:prisma  --format=esm --banner:js=\"import { createRequire } from 'module'; const require = createRequire(import.meta.url);\"",
+		"build": "esbuild src/index.ts --bundle --platform=node --target=node24 --outdir=dist --external:@prisma/client --external:prisma  --format=esm --banner:js=\"import { createRequire } from 'module'; const require = createRequire(import.meta.url);\"",
```
aug24, line 7: Please note that this is Obligatron.  I missed the bundler.  My bad.  The same change has been made for Repocop below.

packages/repocop/src/config.ts
```diff
@@ -77,6 +77,9 @@ export async function getConfig(): Promise<Config> {
 
 	const stage = getEnvOrThrow('STAGE');
 
+	const nodeVersion = process.version;
```
aug24, line 80: Added because it's nice to see this sort of announcement in the logs.

package.json
```diff
@@ -34,8 +34,5 @@
     "tsx": "4.21.0",
     "typescript": "^5.9.3"
   },
-  "prettier": "@guardian/prettier",
```
aug24, line 37: The following lines were added to make the previous change (Obligatron to node 24) pass CI, but we couldn't work out why they were needed.  On attempting to reproduce the problem by removing them, CI passed.  So, they aren't needed.  One of life's little mysteries.

akash1810: I don't think change has been reflected in `package-lock.json`? https://github.com/guardian/service-catalogue/blob/1c14fae648c0b84b4ea4cf2e683725a5795b2bfe/package-lock.json#L13-L15

tjsilver (approved):
👍

PR #1943, opened 2026-05-01
Update the cloudquery usage lambda to node version 24

JuliaBrigitte:
## What does this change?

Upgrade Lambda using nodejs20.x to 24.

## How has this be tested
Started local environment und ran
npm -w cloudquery-usage start
and checked table cloudquery_plugin_usage was populated

---
  - https://app.asana.com/0/0/1214114330377302

### DIFF
--- a/packages/cloudquery-usage/README.md
+++ b/packages/cloudquery-usage/README.md
@@ -1,5 +1,6 @@
 # CloudQuery Usage
 Tracks the number of rows collected with [CloudQuery plugin, per day](https://api-docs.cloudquery.io/#tag/teams/operation/GetGroupedTeamUsageSummary).
+The resulting table `cloudquery_plugin_usage` powers the Grafana dashboard CloudQuery paid usage.
 
 By default, yesterday's data is collected. To customise this, set the `START_DATE` and `END_DATE` environment variables.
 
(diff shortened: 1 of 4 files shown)


packages/cloudquery-usage/README.md
```diff
@@ -1,5 +1,6 @@
 # CloudQuery Usage
 Tracks the number of rows collected with [CloudQuery plugin, per day](https://api-docs.cloudquery.io/#tag/teams/operation/GetGroupedTeamUsageSummary).
+The resulting table `cloudquery_plugin_usage` powers the Grafana dashboard CloudQuery paid usage.
```
akash1810: Could we add a link to this dashboard?

akash1810 (approved):
I think the PR title is incorrect as the changes move to Node 22. Otherwise, LTGM!

JuliaBrigitte:
> I think the PR title is incorrect as the changes move to Node 22. Otherwise, LTGM!

@akash1810 Yes I tried to upgrade the branch name to 24 when I decided to upgrade to 24 instead of 22 and it got messy, cleaning this up now.

packages/cloudquery-usage/README.md
JuliaBrigitte: Sure, will add the link.

PR #1944, opened 2026-05-01
Retest get-east-asian-widths issue

aug24:
## What does this change?

package-lock still has references to get-east-asian-widths in it.  It shouldn't have.


tjsilver (commented):
What is the context of this dependency being in the package.lock?

NovemberTang (approved):

PR #1946, opened 2026-05-01
Add read-only user

jorgeazevedo:
## What does this change?

This adds a new user with simple read-only permissions and a secret to hold the password of that user (the password will be set manually, after the prisma migration creates the user).

For developers to access the database from an IDE, without having perms to make changes. This is also useful for AI agents to run queries safely.

Identical to what's documented as the process to create the read-only user grafana user
https://github.com/guardian/service-catalogue/blob/0331ce144829447d8020ec55deb007f962918605/docs/cloudquery-implementation.md?plain=1#L60-L69

And also uses patterns similar to the creation to the repocop user creation (i.e. guard clauses)

https://github.com/guardian/service-catalogue/blob/0331ce144829447d8020ec55deb007f962918605/packages/common/prisma/migrations/20240614105000_repocop_user/migration.sql#L1-L7

## How has it been verified?

I've ran this locally on dev.

I've also included extra tests to check that the new user cannot write or delete tables.

### DIFF
--- a/packages/cdk/lib/service-catalogue.ts
+++ b/packages/cdk/lib/service-catalogue.ts
@@ -279,8 +279,8 @@ export class ServiceCatalogue extends GuStack {
 			dataType: ParameterDataType.TEXT,
 		});
 
-		new Secret(this, 'DevReadonlyPassword', {
-			secretName: `/${stage}/${stack}/${app}/devreadonly-password`,
+		new Secret(this, 'DevReadOnlyPostgresPassword', {
+			secretName: `/${stage}/${stack}/${app}/devreadonly-postgres-password`,
 			description:
 				'Password for the devreadonly Postgres user (readonly access)',
 		});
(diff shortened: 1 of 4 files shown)


packages/cdk/lib/service-catalogue.ts
```diff
@@ -279,6 +280,12 @@ export class ServiceCatalogue extends GuStack {
 			dataType: ParameterDataType.TEXT,
 		});
 
+		new Secret(this, 'DevReadonlyPassword', {
```
tjsilver: It might be good to put 'postgres' in the secret name somewhere to make it clear that it's a password for the postgres user.

tjsilver (approved):
Looks good - small non-blocking comment about naming of the user. Wondering if some documentation about this new user would be helpful - use cases etc?

packages/cdk/lib/service-catalogue.ts
jorgeazevedo: Good idea!

PR #1948, opened 2026-05-06
temp: break dep graph integrator

tjsilver:
## What does this change?

Breaks the dependency graph integrator (see #1931) to enable testing of the recently-added monitoring (see #1940).

Once deployed, I will run the lambda and see if an alert is triggered. Then I will revert this PR.


kelvin-chappell (approved):
👍

PR #1949, opened 2026-05-06
feat: add migrate resolve to cli

tjsilver:
## What does this change?

Adds the functionality to run prisma migrate resolve through the cli package.

I found this useful when needing to resolve a failed migration on CODE.

It follows the same pattern as the migrate function.

## How has it been verified?

- [x] Tested on the CODE database by running 
```
npm -w cli start migrate-resolve -- \
  --stage CODE \
  --migration 20240412103300_aws_resources \
  --state rolled-back \
  --confirm
``` 
and observed the command being executed and resolving the migration failure.
Also tested without the `--confirm` flag and observed the 'Refusing... ' message.

### DIFF
--- a/packages/cli/src/database.ts
+++ b/packages/cli/src/database.ts
@@ -78,6 +78,44 @@ export async function migrateRdsDatabase(
 	return Promise.resolve(0);
 }
 
+type ResolveState = 'rolled-back' | 'applied';
+
+export async function resolveRdsMigration(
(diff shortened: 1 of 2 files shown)


tjsilver:
Closing after discussion with @akash1810  - on balance this is not required and we are looking at removing the migrate functionality from CLI as this is now automated in the ECS container.

PR #1957, opened 2026-05-13
feat: upgrade postgres to v18

tjsilver:
## What does this change?

Upgrades the postgres database to v18.3 which is the latest. Follows #1835.

To ensure we're up-to-date and making use of security patches and performance enhancements, especially now that [AI is detecting long-standing vulnerabilities](https://www.csoonline.com/article/4167137/ai-finds-20-year-old-bugs-in-postgresql-and-mariadb.html).

Release notes:
- [v18.0](https://www.postgresql.org/docs/release/18.0/)
- [v18.3](https://www.postgresql.org/docs/release/18.3/)

This is almost a like-for-like upgrade from postgres v17, but where it is appropriate to pin the version to a SHA, it has been. I have also included the alpine version in the postgres image tag, which makes it easier for us to use the right alpine version in `docker-compose.yaml`.

## How has it been verified?

- [x] Tested migration locally by running `./packages/dev-environment/script/start-prisma-migrate-test`
- [x] Tested upgrade locally 
- [x] Tested on CODE and observed migrate container complete successfully (after approval)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
     services:
       # See https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers
       postgres:
-        image: postgres:17.9-alpine@sha256:6f30057d31f5861b66f3545d4821f987aacf1dd920765f0acadea0c58ff975b1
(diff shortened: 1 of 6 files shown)


kelvin-chappell (approved):
🚀

PR #1959, opened 2026-05-19
Add devcontainer configuration

tjsilver:
## What does this change?

- Adds a devcontainer configuration using guardian/devenv. 
- Modifies the start and setup scripts to be compatible with running dev-environment either locally or in a devcontainer. 
  - Modifies the setup script to use mise for installing node and to use `.tool-versions` as the repo no longer has `.nvmrc`.
  - Removes the VPN check from the start script as this is incompatible with running in a devcontainer.

To provide a secure and consistent environment for local testing.

Devenv was chosen as this is our new implementation of devcontainers which simplifies config and setup on a repo.

## How has it been verified?

- [x] Dev container starts from shared config
- [x] Can run dev-environment in devcontainer 
- [x] Can run migrate dev in devcontainer
- [x] Can run cdk synth in devcontainer
- [x] Can still run dev-environment locally


containers/db-copy/entrypoint.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-if [[ -z "$TABLES" ]]; then
```
tjsilver, line 5: This change was made because apparently this was in bash and not ash format

scripts/setup.sh
```diff
@@ -19,35 +19,59 @@ step(){
...
+}
+
 check_node_version() {
   step "Checking node version"
+  local runningNodeVersion requiredNodeVersion runningNodeVersionNumber
+  local runningNodeMajor requiredNodeMajor
+
   runningNodeVersion=$(node -v)
-  requiredNodeVersion=$(cat "$ROOT_DIR/.nvmrc")
+  requiredNodeVersion=$(get_required_node_version) || exit 1
```
kelvin-chappell, line 48: I think you can get this more easily by querying mise and then inspecting the result:
```suggestion
  requiredNodeVersion=$(mise list node --current --json | jq -r '.[0].version') || exit 1
```

tjsilver: This does mean that we expect mise to be installed on the user's machine for setup to work though. This script could be run either on the user's machine or in a devcontainer. I created this solution to work regardless of whether mise was installed at that point, then in the function `setup_node_env` checks for mise and/or reminds the user to install it.

kelvin-chappell (approved):
👍

PR #1960, opened 2026-05-19
remove CODE/PROD migration tasks from cli

NovemberTang:
## What does this change?

We have a prisma migrate task available on ecs. It's possible (and preferred) to run prisma migrations on AWS by running `npm -w cli start run-task -- --stage <STAGE> --name prisma-migrate-task`, so this is now the only way to do it.


tjsilver (approved):
👍

PR #1962, opened 2026-05-20
feat: Sync `aws_autoscaling_groups` with load balancer tables

akash1810:
> [!NOTE]
> Recommended reviewing [commit by commit](https://github.com/guardian/service-catalogue/pull/1962/commits).

## What does this change?
Moves the sync of the `aws_autoscaling_groups` table into the same task that syncs the load balancer tables. These tables underpin the Availability dashboard; collecting them at the same time so that joins across them are accurate.

Previously, `aws_autoscaling_groups` was synced once a day at midnight, and the load balancer tables every 30 minutes. This meant that joining the `target_group_arns` column of the `aws_autoscaling_groups` table to `aws_elbv2_target_groups` could yield no results.

Consequently, the task to sync `aws_autoscaling_groups` by itself is no longer needed.


akash1810:
Closing in favour of https://github.com/guardian/service-catalogue/pull/1964.

PR #1967, opened 2026-05-27
add maintenance label to dependency graph integrator PRs

NovemberTang:
## What does this change?

Add a maintenance label to PRs raised by the dependency graph integrator

So we can be core4 compliant

This seemed like the simplest way to do it

## How has it been verified?

Tested on PROD, see the resulting PR with correct label [here](https://github.com/guardian/service-catalogue/pull/1968)


tjsilver (approved):
💯

PR #1969, opened 2026-05-27
feat: add alert_type column to repocop_vulnerabilities table

tjsilver:
## What does this change?

- Adds a 'alert_type' column to the `repocop_vulnerabilities` table, which reflects the 'classification' of the advisory - whether it's 'general' or 'malware'.
- Adds a type and updates test cases to allow test and typecheck to pass.

To enable us to differentiate between malware alerts ('malware') and vulnerabilities ('general').

- [x] Tested on CODE

### DIFF
--- a/packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
+++ b/packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
@@ -1,0 +1,7 @@
+-- Add classification column to repocop_vulnerabilities table. This will enable us to differentiate Dependabot malware alerts from vulnerability alerts.
+BEGIN TRANSACTION;
+
+ALTER TABLE IF EXISTS repocop_vulnerabilities
+  ADD COLUMN IF NOT EXISTS classification text;
+
+COMMIT;
--- a/packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
+++ b/packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
@@ -1,5 +1,19 @@
 -- Add alert_type column to repocop_vulnerabilities table. This will enable us to differentiate Dependabot malware alerts from vulnerability alerts.
 
+BEGIN;
+
 ALTER TABLE IF EXISTS repocop_vulnerabilities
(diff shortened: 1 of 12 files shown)


packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- Add classification column to repocop_vulnerabilities table. This will enable us to differentiate Dependabot malware alerts from vulnerability alerts.
+BEGIN TRANSACTION;
+
+ALTER TABLE IF EXISTS repocop_vulnerabilities
+  ADD COLUMN IF NOT EXISTS classification text;
```
kelvin-chappell: Nitpick: Would it be better to call it `alert_type` or `alert_category`?  Just thinking that `classification` suggests data sensitivity.

NovemberTang: could this be a `NOT NULL`? enforcing this property at the DB acts as a built in error check

packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- Add classification column to repocop_vulnerabilities table. This will enable us to differentiate Dependabot malware alerts from vulnerability alerts.
+BEGIN TRANSACTION;
```
NovemberTang: nitpick: As this is a single command, it doesn't need to be a transaction.

packages/common/prisma/migrations/20260527164433_repocop_vulnerabilities_classification_column/migration.sql
tjsilver: Done in https://github.com/guardian/service-catalogue/pull/1969/commits/a6b6151795aecc1efe0a554e8377240e1dc68409

tjsilver: Done in [dd5aede](https://github.com/guardian/service-catalogue/pull/1969/commits/dd5aede6d61379db252364989a26e0142aab0b75)

kelvin-chappell (approved):
👍

PR #1970, opened 2026-06-01
chore(ci): Remove redundant configuration

akash1810:
## What does this change?
Update the configuration of `guardian/actions-riff-raff`; when `projectName` is provided, `app` is ignored.

## How has it been verified?
The action should continue to work as normal. For example, it should add a comment to this PR with quick links to deploy the project to CODE.

---

See also https://github.com/guardian/service-catalogue/pull/64.


kelvin-chappell (approved):
👍

PR #1971, opened 2026-06-01
Switch to new stackset role

jorgeazevedo:
## What does this change?

Following the change in aws-account-setup to deploy the role using stacksets instead of riff-raff, we need to update service-catalogue to use the new role name https://github.com/guardian/aws-account-setup/pull/612

As per https://github.com/guardian/aws-account-setup/pull/612#discussion_r3317045484 , the approach to test this should be to run service-catalogue in CODE to check the new role operates as expected. Then we can clean up the old role from aws-account-setup

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -877,7 +877,7 @@ spec:
   otel_endpoint_insecure: true
   spec:
     org:
-      member_role_name: cloudquery-access
+      member_role_name: service-catalogue-access
       organization_units:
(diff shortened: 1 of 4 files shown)


akash1810 (approved):

akash1810:
It looks like we might have missed a couple of [references](https://github.com/search?q=repo%3Aguardian%2Fservice-catalogue%20cloudquery-access&type=code), for example: 

https://github.com/guardian/service-catalogue/blob/025c50d01bb4122e6231f913b870fa3d90ca3920/packages/cdk/lib/cloudquery/config.ts#L185

https://github.com/guardian/service-catalogue/pull/1972.

PR #1978, opened 2026-06-17
feat: use prisma seeding for running dev-environment

tjsilver:
## What does this change?

This PR builds on the work that @NovemberTang undertook in #1925.  

We have replaced copying tables from the CODE postgres database to the dev-environment, and now use Prisma seeding (which became available in Prisma v7) to populate tables in the local database instead. This removes our dependency on the VPN.

- Updates the `start` script to launch the required containers, run Prisma migrations, generate the client, and seed the local database. 
- Removes the `db-copy` service and its associated Dockerfile and scripts, eliminating the previous approach of copying data from the CODE database via AWS and SQL dumps. 
- Updates the Repocop README to revise local running instructions to match the new setup (also took the opportunity to remove the camel-case on Repocop, as we generally don't use that form!). 
- Adds `tsx` as a development dependency to support TypeScript-based Prisma seeds. 

### Fixes
- It turns out cloudquery wasn't fetching its plugin locally due to it pointing to an outdated Github registry. I removed this and now it collects the aws tables again.

### Tweaks

- Moves the custom properties query to an earlier location. This was just a small refactor to fit the new seeding flow — custom properties are now seeded local input, so Repocop loads them earlier with the rest of its startup data.
* Minor fix to the defaulting of the `GALAXIES_BUCKET` env var in the CloudQuery service to remove an irritating warning when running locally.

It has been becoming increasingly difficult to run `dev-environment` and/or test packages locally without errors, as the database tables were copied over from CODE. The CODE database had drifted somewhat.

We also had to rely on the VPN in order to copy the tables over from CODE. See #1925 for more details.

Using Prisma seeding removes this repo's dependency on the VPN, which was required for copying over the CODE database tables. 
TODO: write an ADR for this.

## How has it been verified?

- I have successfully tested locally by running `npm -w dev-environment start` followed by the following packages, and queried the DEV database to ensure tables were populated as expected.  (I re-ran `dev-environment` each time to reseed the database first.)
  - `npm -w repocop start`
  - `npm -w cloudbuster start`
  - `npm -w cloudquery-usage start`
  - `npm -w github-actions-usage start`
  - `npm -w obligatron start`
  - `npm -w refresh-materialized-view start`
 
- I have tested this both in the dev-container and on my local machine.


packages/common/prisma.config.ts
```diff
@@ -9,16 +9,16 @@ if (process.env.STAGE !== 'PROD' && process.env.STAGE !== 'CODE') {
...
 export default defineConfig({
 	schema: path.join('prisma', 'schema.prisma'),
+	migrations: {
+		seed: 'npx tsx ./prisma/seed.ts',
+	},
 	datasource: {
 		url: getDatabaseConnectionString({
 			hostname: process.env.DATABASE_HOSTNAME as string,
 			user: process.env.DATABASE_USER as string,
 			password: process.env.DATABASE_PASSWORD as string,
```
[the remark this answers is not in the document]
tjsilver: Fixed in f63ead9561554c7210a21538c444dfa6e2d16be5

packages/common/prisma/seed.ts
```diff
@@ -0,0 +1,726 @@
...
+            )
+            LANGUAGE sql
+        AS $$
+            SELECT
+                NULL::text, NULL::text, NULL::text, NULL::text,
+                NULL::text, NULL::text, NULL::text, NULL::boolean,
+                NULL::jsonb
+            WHERE false;
+        $$;
+    `);
```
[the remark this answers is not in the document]
tjsilver: Fixed in 843356fd

packages/dev-environment/docker-compose.yaml
```diff
@@ -51,19 +7,32 @@ services:
...
+
+  prisma-seed:
+    image: node:24.16-trixie
+    # Runs as root to avoid permission issues with named volumes in local dev
+    user: "0:0"
+    working_dir: /workspace/packages/common
+    volumes:
+      - ../../:/workspace
+      - prisma_seed_root_node_modules:/workspace/node_modules
+      - prisma_seed_common_node_modules:/workspace/packages/common/node_modules
```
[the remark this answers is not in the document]
tjsilver: Fixed in 6b86bdb

packages/common/prisma/seed.ts
```diff
@@ -0,0 +1,650 @@
+import { randomUUID } from 'node:crypto';
+import { PrismaPg } from '@prisma/adapter-pg';
+import { Prisma, PrismaClient } from '../prisma-client/client.js';
+
+const databaseUrl = process.env.DATABASE_URL;
+if (!databaseUrl) {
+	throw new Error('DATABASE_URL is not set');
+}
```
NovemberTang, line 23:
```suggestion

```

I think we could simplify and get rid of this, as it's covered by the try/catch below. A `null`/`undefined`/`""` will return a `TypeError`

packages/common/prisma/seed.ts
```diff
@@ -0,0 +1,650 @@
...
+	throw new Error(`DATABASE_URL is invalid: "${databaseUrl}"`);
+}
+
+const prisma = new PrismaClient({
+	adapter: new PrismaPg({ connectionString: databaseUrl }),
+});
+
+const orgName = 'guardian';
+const cqSourceName = 'seed';
+const seededAt = new Date('2026-06-01T12:00:00.000Z');
```
NovemberTang: Should this date be hardcoded?

packages/common/prisma/seed.ts
```diff
@@ -0,0 +1,650 @@
...
+
+const defaultBranchName = 'main';
+const branchNames = [defaultBranchName, 'develop', 'feature-1'] as const;
```
NovemberTang: Why do we have `as const` on some of these values but not others?

packages/common/prisma/seed.ts
NovemberTang, line 1: This file is very long. I'm not sure we need to keep all of the detail in one place. Could it be broken up a little? Ideally `seed.ts` would handle the top level seeding logic/logging, and the complexity/implementation detail can be hidden in other files to keep it more maintainable/readable/reviewable!

packages/dev-environment/script/start
```diff
@@ -12,68 +12,91 @@ cyan='\033[0;36m'
 yellow="\033[1;33m"
 STEP_COUNT=0
 
+COMPOSE_CMD=(docker compose -f "${DIR}/../docker-compose.yaml" --env-file "$MAIN_ENV_FILE")
```
NovemberTang, line 19: minor: Could we add a commend explaining what this command does? I had assumed that providing two env files meant that the "last one wins" but it seems like what actually happens is that they get [loaded in order and are only ovewritten if there are duplicate keys in both files](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#substitute-with---env-file).

We were always doubling up on --env-file, this change just brought it to my attention, so this is more of a boy scout suggestion than a comment on anything in this PR

packages/dev-environment/script/start
```diff
@@ -12,68 +12,91 @@ cyan='\033[0;36m'
...
+  step "Checking Prisma CLI"
 
-  STATUS=$(aws sts get-caller-identity --profile deployTools 2>&1 || true)
-  if [[ ${STATUS} =~ (ExpiredToken) ]]; then
-    echo "Credentials for the deployTools profile have expired. Please fetch new credentials, and run this script again."
-    exit 1
-  elif [[ ${STATUS} =~ ("could not be found") ]]; then
-    echo "Credentials for the deployTools profile are missing. Please fetch some, and run this script again."
+  if ! npm --prefix "$ROOT_DIR/packages/common" exec prisma -- --version >/dev/null 2>&1; then
+    echo "Prisma is not installed for packages/common."
```
NovemberTang, line 46: Smart!

packages/dev-environment/docker-compose.yaml
```diff
@@ -85,10 +38,11 @@ services:
       CQ_AWS: ${CQ_AWS}
       CQ_GITHUB: ${CQ_GITHUB}
       GITHUB_ACCESS_TOKEN: ${GITHUB_ACCESS_TOKEN}
-      GALAXIES_BUCKET: ${GALAXIES_BUCKET}
+      GALAXIES_BUCKET: ${GALAXIES_BUCKET:-}
```
NovemberTang: What does this do? I think it was mentioned briefly in the PR description, but I'm still not sure what the issue was, or how these new characters fix it

packages/dev-environment/README.md
```diff
@@ -39,39 +41,73 @@ It includes:
    ./packages/dev-environment/script/start
    ```
 
-   This will create a sql directory with the definition of the tables to be populated in the dev envionment, will start the Docker containers, and CloudQuery will start collecting data.
+3. This will:
+   - start the local Postgres container
+   - run Prisma migrations
+   - generate the Prisma client
+   - seed the database with local development data using Prisma
+   - start CloudQuery to collect additional data
```
NovemberTang, line 49: Good to include this, I'd actually forgotten we were still using CQ to collect some stuff on the fly. Maybe we can reduce our reliance on this in the future

packages/dev-environment/README.md
```diff
@@ -39,39 +41,73 @@ It includes:
...
+   - start Grafana
+
+   Unlike the previous setup, this does **not** copy data from another database.
```
NovemberTang: I don't think we need references to previous implementations outside an ADR/PR/commit message, it might cause confusion to anyone looking at this in a year or so

NovemberTang (commented):
Looks broadly good to me! I've suggested a couple of small changes and one big one (splitting the seed file up).

We should also change the ADR status to `Accepted` as part of this PR. Obviously that's not possible until we've rebased on to #1980 , but thought I'd mention it now before I forget.

packages/common/prisma/seed.ts
tjsilver: Removed in 5896c0bf. Was added to be a useful 'fail early' message if the database url was somehow malformed, but in practice this is unlikely as it's formed from the .env variables anyway.

packages/common/prisma/seed.ts
tjsilver: Yes, but I noticed it's only used once so I've moved into it's containing folder. See [eb318c5](https://github.com/guardian/service-catalogue/pull/1978/commits/eb318c55ad788ace9d38183229fa2ccbebf6baa4).

packages/common/prisma/seed.ts
tjsilver: It's not strictly necessary but useful for fixed seed data such as these, as it means typescript preserves the exact literal values and marks them as readonly,  i.e.
`const branchNames = ['main', 'develop']; // type: string[]`
`const branchNames = ['main', 'develop'] as const; // type: readonly ['main', 'develop']`
It's not on everything because it's most useful for fixed arrays/objects; it adds little value for ordinary scalar constants or Date values, for example.

ADR/05-prisma-seeding.md
akash1810, line 1: I think there is a forth option, which possibly has the least amount of maintenance - use [`pgdump`](https://www.postgresql.org/docs/current/app-pgdump.html) against the database, outputting to S3. Then, locally, use that S3 object to populate the database. We could run `pgdump` on a schedule to avoid stale data.

This option:
- Removes the complexity of using Prisma seeding
- Uses real data
- Results in DEV being more reflective of PROD as all tables have values

WDYT?

packages/common/prisma/seed.ts
tjsilver: Agreed - I've broken it up into logical modules with explanatory comments - hope this is much clearer! Done in 70c81234.

packages/dev-environment/script/start
tjsilver: Done in d865839.

packages/dev-environment/docker-compose.yaml
tjsilver: I've added an explanatory comment - it basically stops us getting the warning that `GALAXIES_BUCKET` is not set when running the start script. See db4bec0.

packages/dev-environment/README.md
tjsilver: Done in fa647ae.

ADR/05-prisma-seeding.md
tjsilver: I've added it as a fourth position to the ADR so we can decide as a team, here 41512fe.

tjsilver:
> Looks broadly good to me! I've suggested a couple of small changes and one big one (splitting the seed file up).
> 
> We should also change the ADR status to `Accepted` as part of this PR. Obviously that's not possible until we've rebased on to #1980 , but thought I'd mention it now before I forget.

I haven't marked it as 'accepted' yet (it was not merged to main, but to this branch, so it's not on the repo yet), as we are now considering a fourth option, see https://github.com/guardian/service-catalogue/commit/41512fe605640c7bba9130bbf87d6576c369d2d8.

ADR/05-prisma-seeding.md
```diff
@@ -0,0 +1,160 @@
...
+### 4. Restore local data from a scheduled Postgres dump
+
+Here, a PostgreSQL dump of a representative environment is created on a schedule with pg_dump and stored in S3. Local development restores that dump into the local Postgres database during startup or setup.
+
+#### Advantages
+
+- Uses real data rather than synthetic fixtures.
+- Can make local development more reflective of CODE or PROD-derived data because more tables and relationships are populated.
+- Reduces the need to design and maintain Prisma seed fixtures by hand.
+- Decouples local setup from direct runtime access to CODE at startup time, since developers restore from a stored dump rather than copying live data.
```
NovemberTang: This approach looks like what we were already doing with the db_copy container, but with the introduction of a cache?

I assume one intended benefits of this is speed, as the current process being slow is a pain. Do we know if the bottleneck here is:

1. pgdump operation
2. network speed
3. db write speed

If it's 1, then we could reasonably say this was a performance improvement, but if it's 2 or 3, we probably won't see a meaningful difference in startup time

ADR/05-prisma-seeding.md
```diff
@@ -0,0 +1,160 @@
...
+- A scheduled dump can provide a reasonably current dataset without requiring every developer to connect to source systems directly.
+
+#### Disadvantages
```
NovemberTang, line 106: One of the intentions of this work (which could possibly be more emphasised in this ADR) is to allow automated testing of lambdas, locally and in CI. One disadvantage I'd add to this scenario is that nondeterministic table contents makes this kind of testing difficult on a test runner, and also means granting the runner access to some (though stale) potentially sensitive production data, which is a security risk. Even more so as supply chain attacks on CI runners are increasing.

I think the easier thing to do if we wanted a prod-like environment, would be to just deploy whatever you wanted to test to code, and test it there

packages/common/prisma/seed/seed-assembly.ts
```diff
@@ -0,0 +1,112 @@
...
+	repoDefinitions,
+	teamDefinitions,
+} from './seed-data.js';
+import { createSeedMetadata } from './seed-helpers.js';
+import type { SeedData, TeamSlug } from './seed-types.js';
+
+/**
+ * Builds the intentionally invalid workflow row used by local testing scenarios.
+ */
+function createInvalidSeedWorkflow(
```
NovemberTang:
```suggestion
function createInvalidWorkflowSeed(
```
minor: This function name makes it sound like it's the seed that's invalid, rather than the workflow

NovemberTang (approved):
Amazing work!

PR #1980, opened 2026-06-26
docs: add ADR for Prisma seeding

tjsilver:
## What does this change?

Adds an ADR for the Prisma seeding work in #1978.

### DIFF
--- a/ADR/05-prisma-seeding.md
+++ b/ADR/05-prisma-seeding.md
@@ -6,9 +6,7 @@ Proposed.
 
 ## Context
 
-The local development environment (`dev-environment`) previously populated its database by copying selected tables from CODE using a dedicated `db-copy` container.
-
-It now uses Prisma in the `common` package, including a seed script at `packages/common/prisma/seed.ts`.
+We run Service Catalogue projects locally by spinning up a Postgres database in a container, and testing against that.
 
 For local development, the database needs enough representative data to:
 
@@ -16,6 +14,10 @@ For local development, the database needs enough representative data to:
 - support local execution of packages and queries
 - provide a baseline for local debugging and development
 - allow further local population by CloudQuery where required.
+
+The local development environment (`dev-environment`) previously populated its database by copying selected tables from CODE using a dedicated `db-copy` container.
+
+It now uses Prisma in the `common` package, including a seed script at `packages/common/prisma/seed.ts`.
 
 This ADR records the decision to move local database initialisation away from copying CODE data and towards Prisma seeding.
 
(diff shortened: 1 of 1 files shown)


ADR/05-prisma-seeding.md
```diff
@@ -0,0 +1,130 @@
...
+
+Proposed.
+
+## Context
+
+The local development environment (`dev-environment`) previously populated its database by copying selected tables from CODE using a dedicated `db-copy` container.
+
+It now uses Prisma in the `common` package, including a seed script at `packages/common/prisma/seed.ts`.
+
+For local development, the database needs enough representative data to:
```
NovemberTang, line 11: imo it reads more clearly if we have this list up at the top.

ADR/05-prisma-seeding.md
```diff
@@ -0,0 +1,130 @@
...
+- Can expose issues that only appear with large or messy real-world datasets.
+
+#### Disadvantages
+
+- Couples local development to VPN access, CODE availability and credentials.
+- Makes the local starting dataset non-deterministic.
+- Risks local failures if copied tables drift from the current schema or application expectations.
+- Adds operational complexity through an extra container and data-copy process.
+- Makes local debugging and test reproduction less reliable.
+- Brings more production-derived data into local environments than is necessary.
```
NovemberTang, line 43:
```suggestion
- Brings more production-derived data into local environments than is necessary.
- Allowing developers (and by extension, AIs) access to this database poses a security risk.
```

I believe we don't place strong restrictions on the user that performs the copy, so once we're connected, we can do whatever we like. Keeping a local copy of production data on our machines also poses a risk

ADR/05-prisma-seeding.md
```diff
@@ -0,0 +1,130 @@
...
+- selected AWS resources
+- local-only helpers (where required by dependent views).
+
+#### Advantages
+
+- The development environment no longer depends on the VPN or copying from CODE.
+  Local startup is simpler and more self-contained.
+- Data is deterministic and reproducible.
+- Seed data can be reviewed and versioned alongside schema changes.
+- Developers can evolve the local dataset intentionally to cover specific scenarios.
```
NovemberTang, line 67:
```suggestion
- Developers can evolve the local dataset intentionally to cover specific scenarios.
- Deterministic data makes E2E and integration testing much simpler, something we only do manually at the moment.
```

ADR/05-prisma-seeding.md
```diff
@@ -0,0 +1,130 @@
...
+
+This replaces the previous `db-copy` approach for local development.
+
+The seeded dataset should remain intentionally small and focused on supporting local development, rather than mirroring CODE in full.
+
+## Consequences
+
+### Positive
+
+- Local development is more reliable and reproducible.
```
NovemberTang:
```suggestion
- Local development is more reliable and reproducible, and faster.
```

ADR/05-prisma-seeding.md
NovemberTang, line 11:
```suggestion
We run service catalogue projects locally by spinning up a postgres DB in a container, and testing against that. For local development, the database needs enough representative data to:
```

Adding a bit of context on why the local db exists

NovemberTang (approved):
I think this looks great! No major changes, just some suggestions for improving the flow, and a couple of extra advantages/disadvantages

PR #1987, opened 2026-07-02
Add tables required for breakglass alerting

NovemberTang:
## What does this change?

Makes two new tables available to prisma

- `aws_iam_users`
- `aws_iam_credential_reports`

Adds a `state` column to `aws_organizations_accounts`, which was probably added in a minor version update, but was missed as it's not currently used anywhere

syncs iam_users at at the same time as credential_reports

To allow cloudbuster to send alerts relating to incorrect setup of IAM users

Doing this via cloudbuster means we can utilise and already existing alert system and expanding, similar to what we could to with the [experimental repocop features](https://github.com/guardian/service-catalogue/blob/a5a125d03f5e5b0fde605837ab7af308954b5a3c/packages/repocop/src/evaluation/repository.ts#L366)

## How has it been verified?

Deployed to code, migration completes successfully


akash1810 (approved):
> To allow cloudbuster to send alerts relating to incorrect setup of IAM users

You're likely aware of it already - we previously reduced what Security HQ does (https://github.com/guardian/security-hq/pull/1129) and created a [dashboard](https://metrics.gutools.co.uk/d/bdn97cui5rbi8f/iam-credentials-report) in its place. Not sure if it'll be helpful as a dashboard to link to in these new alerts?

PR #1988, opened 2026-07-06
split cloudbuster into user report and fsbp report

NovemberTang:
## What does this change?

Pull FSBP logic into its own function, `createFsbpTableAndAlerts`. Create a new function,`userReport` for notification of improperly configured breakglass credentials.

In order for the cloudbuster user to read these tables, I've expanded its permissions in line with #1176 

To simplify cloudbuster maintenance going forward, split the two features into two functions.

The cadence of and permissions of cloudbuster is appropriate for this kind of alert. It makes sense to use the service catalogue and postgres to deliver these messages.

## How has it been verified?

- [x] Cloudbuster still functions on CODE
- [x] Cloudbuster is able to access the new tables
- [ ] Test migration/user report on CODE after approval, before merge

### DIFF
--- a/packages/cloudbuster/src/index.ts
+++ b/packages/cloudbuster/src/index.ts
@@ -93,7 +93,7 @@ async function createFsbpTableAndAlerts(
 	);
 }
 
-async function userReport(prisma: PrismaClient) {
+async function breakglassUserReport(prisma: PrismaClient) {
 	const user = await prisma.aws_iam_users.findFirst();
 	const report = await prisma.aws_iam_credential_reports.findFirst();
 
@@ -102,6 +102,7 @@ async function userReport(prisma: PrismaClient) {
 }
 
 export async function main() {
+	logger.log({ message: 'Cloudbuster run starting.' });
 	// *** SETUP ***
 	const config = await getConfig();
 	const prisma = getPrismaClient(config);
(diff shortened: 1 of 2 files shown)


packages/cloudbuster/src/index.ts
```diff
@@ -79,19 +83,35 @@ export async function main() {
...
+	// *** NOTIFICATION SENDING ***
 	logger.log({ message: `Sending ${digests.length} digests.` });
 	await Promise.all(
 		digests.map(
 			async (digest) => await sendDigest(anghammaradClient, config, digest),
 		),
 	);
+}
+
+async function userReport(prisma: PrismaClient) {
```
tjsilver: Maybe this could be more explictly named, like `breakglassUserReport`?

packages/cloudbuster/src/index.ts
```diff
@@ -79,19 +83,35 @@ export async function main() {
...
+	const user = await prisma.aws_iam_users.findFirst();
+	const report = await prisma.aws_iam_credential_reports.findFirst();
+
+	console.log(`User: ${user?.user_name}`);
+	console.log(`Report: ${report?.user}`);
+}
+
+export async function main() {
+	// *** SETUP ***
```
tjsilver, line 106: Should there also be a 'starting Cloudbuster' log line at the beginning of main now?

tjsilver (approved):
👍

PR #1990, opened 2026-07-06
Fix dev environment

NovemberTang:
co-author: @tjsilver 

## What does this change?

This was failing because we were trying to apply prisma migrations twice. Additionally, some chunks of #1978 were removed due to an erroneous rebase. This PR fixes both problems

## How has it been verified?

Containers run locally now

### DIFF
--- a/packages/dev-environment/docker-compose.yaml
+++ b/packages/dev-environment/docker-compose.yaml
@@ -1,7 +1,4 @@
 services:
-  # Postgres starts empty. The schema is created and the database is seeded by
-  # Prisma (migrate deploy + seed) from the start script, so no init scripts are
-  # mounted here.
   postgres:
     image: postgres:18.3
     ports:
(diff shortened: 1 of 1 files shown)


packages/dev-environment/docker-compose.yaml
```diff
@@ -1,50 +1,7 @@
...
-        - aws_elbv2_load_balancers
-        - aws_acm_certificates
-        - aws_apigateway_domain_names
-        - fastly_service_backends
-        - fastly_service_domains
-        - aws_cloudwatch_alarms
-        - aws_autoscaling_groups
-        - aws_elbv2_target_groups
+  # Postgres starts empty. The schema is created and the database is seeded by
+  # Prisma (migrate deploy + seed) from the start script, so no init scripts are
```
tjsilver: Not sure we need this comment

NovemberTang: agreed

tjsilver (approved):

PR #1993, opened 2026-07-08
find users with no name tag/mfa

NovemberTang:
## What does this change?

Implements the core logic for identifying users with at least one of

- No MFA
- No google username tag

As part of keeping Janus resilient, we need to make sure that AWS breakglass users are compliant, and findable.

We could message users directly if they don't have MFA set up, but this introduces some problems

1. Setting up email infrastructure is fiddly, and requires collaboration across departments which significantly slows things down.
2. If the user has no email, we don't know who we are meant to contact in the first place
3. Account health is a whole-team responsibility, not just the responsibility of the breakglass user
4. If the breakglass user has left, the email will just bounce

Extending via cloudbuster has the following benefits

1. We can plug into an existing messaging system
2. We can reach the entire team, which mitigates challenges 2-4 above
3. We can follow the same cadence that cloudbuster uses.

## How has it been verified?

Added unit tests. Verified on CODE, except for last commit

In the next PR, I'll add some end-to-end tests, making use of our seeded local db!

### DIFF
--- a/packages/cloudbuster/src/breakglass.ts
+++ b/packages/cloudbuster/src/breakglass.ts
@@ -1,15 +1,15 @@
 import type {
-    aws_accounts,
-    aws_iam_credential_reports,
-    aws_iam_users,
+	aws_accounts,
+	aws_iam_credential_reports,
+	aws_iam_users,
 } from 'common/prisma-client/client.js';
 
 export interface BreakglassUser {
-    accountName: string | null;
-    user: string | null;
-    userUrl: string | null;
-    mfaActive: boolean | null;
-    hasUsernameTag: boolean;
+	accountName: string | null;
+	user: string | null;
+	userUrl: string | null;
+	mfaActive: boolean | null;
+	hasUsernameTag: boolean;
 }
 
 /**
@@ -19,7 +19,7 @@ export interface BreakglassUser {
  * E.g. `arn:aws:iam::123456789012:user/alice` -> `123456789012`.
  */
 function accountIdFromArn(arn: string | null): string | null {
-    return arn?.split(':')[4] ?? null;
+	return arn?.split(':')[4] ?? null;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/cloudbuster/src/breakglass.ts
```diff
@@ -0,0 +1,57 @@
...
+
+			return {
+				accountName: account?.name ?? null,
+				user: cr.user,
+				userUrl: cr.user
+					? `https://console.aws.amazon.com/iam/home#/users/${cr.user}`
+					: null,
+				mfaActive: cr.mfa_active,
+				hasUsernameTag: tags?.['GoogleUsername'] != null,
+			};
```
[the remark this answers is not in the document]
NovemberTang: I think we can be reasonably sure about the format of tags. I'm willing to accept the risk.

packages/cloudbuster/src/index.ts
```diff
@@ -94,11 +100,28 @@ async function createFsbpTableAndAlerts(
...
+	);
+
+	logger.log({
+		message: JSON.stringify(
+			report.map((r) => ({
+				accountName: r.accountName,
+				user: r.user,
+			})),
+		),
+	});
```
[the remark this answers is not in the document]
NovemberTang: This is only for testing purposes, it will be removed when we start sending messages.

packages/common/src/database-queries.ts
```diff
@@ -70,3 +73,21 @@ export async function getExternalTeams(
 	const teams = await client.guardian_non_p_and_e_github_teams.findMany();
 	return toNonEmptyArray(teams.map((t) => t.team_name));
 }
+
+export async function getIamCredentialReports(
+	client: PrismaClient,
+): Promise<aws_iam_credential_reports[]> {
+	return await client.aws_iam_credential_reports.findMany();
```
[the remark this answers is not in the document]
NovemberTang: This is not inconsistent with other query helpers. They just have an intermediate await.

packages/cloudbuster/src/breakglass.ts
```diff
@@ -0,0 +1,47 @@
...
+): BreakglassUser[] {
+	const accountsById = new Map(awsAccounts.map((a) => [a.id, a]));
+	const usersByArn = new Map(iamUsers.map((u) => [u.arn, u]));
+
+	return credentialReports
+		.filter((cr) => cr.password_enabled === 'true')
+		.map((cr) => {
+			const accountId = cr.account_id;
+			const account = accountId ? accountsById.get(accountId) : undefined;
+			const user = cr.arn ? usersByArn.get(cr.arn) : undefined;
```
NovemberTang, line 32: We need to find users by ARN as usernames can be repeated in different accounts

packages/cloudbuster/src/breakglass.ts
```diff
@@ -0,0 +1,47 @@
+import type {
+	aws_accounts,
+	aws_iam_credential_reports,
+	aws_iam_users,
+} from 'common/prisma-client/client.js';
+
+export interface BreakglassUser {
+	accountName: string | null;
+	user: string | null;
```
kelvin-chappell, line 9: Can this ever be null?  It looks like it could be undefined

packages/cloudbuster/src/breakglass.ts
```diff
@@ -0,0 +1,47 @@
...
+	userUrl: string | null;
+	mfaActive: boolean | null;
```
kelvin-chappell, line 11: I don't think this can be null either

NovemberTang: I've taken these straight from the prisma schema, where the [DB defines these fields as nullable](https://github.com/guardian/service-catalogue/blob/ac68ee20d229b8f5051d785f9b2ae6cbf43162f0/packages/common/prisma/schema.prisma#L1281). The types are tightened up to better reflect reality in a later [PR](https://github.com/guardian/service-catalogue/pull/1997/changes#diff-e08de8c114c019170458c16003483a33e8a6e0748ac2d0cc8362d948d658f8be) (follow the link and search for `interface BreakglassUser`)

kelvin-chappell (approved):
👍

PR #1994, opened 2026-07-08
Update getting started guide with `devreadonly` instructions

jorgeazevedo:
## What does this change?

Since https://github.com/guardian/service-catalogue/pull/1946, there is a database user with read only permissions that can be used for querying service catalogue locally.

This adds instructions for how to use it to the getting started guide.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -55,7 +55,7 @@ You can get started in the Explore tab, as demonstrated below.
 
 ![A gif showing how to interact with CloudQuery using the Grafana UI](img/queryVideo.gif)
 
-### Read-only database user
+### Local SQL client
 
 An alternative way to query the database is using a local SQL client. To enable this use case, the database has a
 read-only user configured, the credentials for which are available in Secrets Manager at the path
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -43,16 +43,28 @@ contact details in CloudQuery!
...
-
-![A gif showing how to interact with CloudQuery using the Grafana UI](img/queryVideo.gif)
-
 We've stored our CloudQuery data in a Postgres database, so you can use SQL to query it. _You don't have to be technical
 or an engineer to be able to use SQL_. Its basic syntax is designed to be very intuitive. If you are unfamiliar with
 SQL, there are lots of resources online to get you started. Julia Evans has
 a [SQL playground](https://sql-playground.wizardzines.com/) to get beginners used to making simple queries.
 
+### Grafana
+
```
tjsilver, line 52: I wonder if the Grafana section should stay above the sql part, and then the read-only part could go here?

akash1810 (approved):

docs/getting-started.md
jorgeazevedo: I was a bit unsure about how to structure this. In reviewing it I actually just changed a header as I finally figured out what I was trying to do. So my intention was to present service catalogue as a SQL database, that can be accessed either by grafana or from a local client (using the readonly user). Maybe the result isn't as clear as I hoped! Do you see a clearer structure?

PR #1995, opened 2026-07-09
Set up test data for local integration testing of cloudbuster user report

NovemberTang:
## What does this change?

1. Switches cloudbuster's breakglass user report from using `aws_accounts` to `aws_organization_accounts`
2. Creates seed data for testing the user report locally, renaming the existing seeds to be explicitly GitHub related

1. `aws_accounts` is a [view](https://github.com/guardian/service-catalogue/tree/main/packages/common/prisma/migrations/20231026134216_view_aws_accounts/migration.sql). In order to populate it we would need to seed three separate tables: `aws_organizations_accounts`, `aws_organizations_account_parents`, `aws_organizations_organizational_units`. To simplify this, (and probably improve performance very slightly), The AWS account name now comes from `aws_organizations_accounts` instead, meaning we only have to seed one table.
2. One issue with repocop is that it's been difficult to test from end to end, due to the many db calls it makes. Seeding (introduced in #1978 ), helps solve this problem, as we can put deterministic test data in a local db, and then run it however we like (in docker, in an action job, etc). Here, we are creating test data to verify that the breakglass user report correctly identifies IAM users that are missing certain configurations.
## How has it been verified?

Ran `npm run start -w cloudbuster` locally, observed expected output

## What doesn't it do?

To limit scope, this deliberately doesn't introduce an e2e testing framework, or a way to run these tests in CI etc. The rest of cloudbuster, and maybe other parts of the service catalogue, can be tested like this in future PRs

### DIFF
--- a/packages/cloudbuster/src/breakglass.test.ts
+++ b/packages/cloudbuster/src/breakglass.test.ts
@@ -1,9 +1,9 @@
 import assert from 'assert';
 import { describe, it } from 'node:test';
 import type {
-	aws_accounts,
 	aws_iam_credential_reports,
 	aws_iam_users,
+	aws_organizations_accounts,
 } from 'common/prisma-client/client.js';
 import type { BreakglassUser } from './breakglass.js';
 import { createBreakglassUserReport } from './breakglass.js';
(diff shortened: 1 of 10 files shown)


tjsilver (approved):
Tested locally and worked as expected. 👍

PR #1997, opened 2026-07-09
Set up messaging for breakglass misconfiguration reports. Use more realistic types

NovemberTang:
## What does this change?

- Creates `formatMessage` to make the results nice and human-readable. Sends them to users. Tests said message. For now, messages only go to the security team. We can roll out more widely when we're ready.
- Separates FSBP logic from breakglass logic, as index.ts was getting a bit busy. The easiest way to view this change is probably commit by commit. 

## How has it been verified?

Added unit tests, run locally with AWS credentials for anghammarad access.


packages/cloudbuster/src/index.ts
```diff
@@ -135,6 +19,7 @@ export async function main() {
 	);
 
 	await createFsbpTableAndAlerts(config, prisma, anghammaradClient);
-	await breakglassUserReport(prisma);
+	await sendBreakglassUserAlerts(config, prisma, anghammaradClient);
```
aug24: Are these actually dependent, or could this be Promise.All here?  I assume they are genuinely dependent.

aug24 (approved):

aug24:
If we are removing snake case in favour of initcap, could we do it for more things?  This looks like it covers about half.  Might as well keep going :)

packages/cloudbuster/src/index.ts
NovemberTang: You're right. They're independent. Good catch

PR #2008, opened 2026-07-14
refactor: show prisma migrate task in service catalogue app logs

tjsilver:
## What does this change?

- Updates the `APP` environment variable and the `App` Docker label in `addPrismaMigrateTask` to use `service-catalogue` instead of `prisma-migrate-task`.
- Adds a note to the migration README to reflect this change.

Currently, when the task is started via the cli, a link is created to ECS to view the logs. The url sets sets `app.keyword:"service-catalogue"` which means that the logs don't show up for the migrate task, as this has the `app.keyword:"prisma-migrate-task"`. This change should fix that and the link should work.

[Asana ticket](https://app.asana.com/1/1210045093164357/project/1214110347938932/task/1214980667159868)

The issue was limited to Kibana log attribution, so only the logging metadata has changed, not the task identity itself. This avoids altering task family naming, IAM/resource naming, task discovery conventions, and other infrastructure metadata. 

## How has it been verified?

- [x] Tested locally by deploying this branch to CODE and running the cli task (`npm -w cli start run-task -- --stage CODE --name prisma-migrate-task`) and observing the link correctly displaying the [logs](https://logs.gutools.co.uk/s/devx/app/r/s/AOnD1).

### DIFF
--- a/docs/database-migrations.md
+++ b/docs/database-migrations.md
@@ -82,6 +82,12 @@ You can apply a migration to CODE or PROD using the [CLI](../packages/cli)):
 npm -w cli start run-task -- --stage [CODE|PROD] --name prisma-migrate-task
 ```
 
+#### Viewing migration logs
+
+> [!NOTE]
+> Although the ECS task is named `prisma-migrate-task`, its shipped logs are attributed to `service-catalogue` in Kibana.
+> To narrow to a specific migration run, use the ECS task ARN from the CLI log link.
+
 See also:
 
 - https://www.prisma.io/docs/orm/prisma-client/deployment/deploy-database-changes-with-prisma-migrate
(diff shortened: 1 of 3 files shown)


akash1810 (changes requested):
> Currently, when the task is started via the cli, a link is created to ECS to view the logs. The url sets sets app.keyword:"service-catalogue" which means that the logs don't show up for the migrate task, as this has the app.keyword:"prisma-migrate-task". This change should fix that and the link should work.

Could we patch the [code that generates the deep-link to Central ELK](https://github.com/guardian/service-catalogue/blob/c50fe5323dbfd66051483634dc0d7f38ef36686c/packages/common/src/logs.ts#L13-L39) instead? The service catalogue repository consists of multiple apps. I feel like this change will lead to confusion.

tjsilver:
> > Currently, when the task is started via the cli, a link is created to ECS to view the logs. The url sets sets app.keyword:"service-catalogue" which means that the logs don't show up for the migrate task, as this has the app.keyword:"prisma-migrate-task". This change should fix that and the link should work.
> 
> Could we patch the [code that generates the deep-link to Central ELK](https://github.com/guardian/service-catalogue/blob/c50fe5323dbfd66051483634dc0d7f38ef36686c/packages/common/src/logs.ts#L13-L39) instead? The service catalogue repository consists of multiple apps. I feel like this change will lead to confusion.

Agreed by the team - will close this PR and open a new one for the link change.

PR #2018, opened 2026-07-16
fix: Use ARN in links to Kibana

tjsilver:
## What does this change?

- Adds a function to build the log link using the task ARN rather than the app name. 
- Adds tests for the new function.
- Drive-by removes a reference to the VPN requirement in migration docs.

Note: The service catalogue log linking is unaffected, as this still calls the original function directly. 

Currently, when the task is started via the cli, a link is created to ECS to view the logs. The url sets sets app.keyword:"service-catalogue" which means that the logs don't show up for the migrate task, as this has the app.keyword:"prisma-migrate-task". This change should fix that and the link should work.

[Asana ticket](https://app.asana.com/1/1210045093164357/project/1214110347938932/task/1214980667159868)

In the [previous PR](#2008) to address this, it was suggested that we just change the link rather than the task. Now, we use the ARN to build the link so it always goes to the right place. 

## How has it been verified?

- [x] Tested locally by running the cli task (`npm -w cli start run-task -- --stage CODE --name prisma-migrate-task`) and observing the link correctly displaying the [logs](https://logs.gutools.co.uk/s/devx/app/r/s/xsTHn).

### DIFF
--- a/docs/database-migrations.md
+++ b/docs/database-migrations.md
@@ -74,7 +74,6 @@ Prerequisite:
 
 1. You have an approved Pull Request
 2. You have `deployTools` credentials from Janus
-3. You are connected to the VPN
 
 You can apply a migration to CODE or PROD using the [CLI](../packages/cli)):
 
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

PR #2020, opened 2026-07-20
Add a script for quickly testing the deps-graph-integrator on CODE

andrew-nowak:
## What does this change?

Adds a script which sends a test event onto the dependency-graph-integrator's input topic on CODE. 

Makes it quicker and easier to send a test event, without having to enter & navigate the AWS console each time.

We chose to send the test event via the SNS topic, as it avoids having to create the SNS event wrapper and fill it with fake data, plus it tests that the topic is correctly set up to invoke the lambda.

## How has it been verified?

The DGI has been run on CODE using this script.

### DIFF
--- a/packages/dependency-graph-integrator/package.json
+++ b/packages/dependency-graph-integrator/package.json
@@ -12,7 +12,7 @@
 		"postbuild": "cp package.json dist/package.json",
 		"prestart": "npm run copy-templates",
 		"start": "tsx src/run-locally.ts",
-		"send-test-event": "tsx src/test-code.ts",
+		"invoke-code": "tsx src/test-code.ts",
 		"pretest": "npm run copy-templates",
 		"test": "node --import tsx --test **/*.test.ts",
 		"test:update": "UPDATE_SNAPSHOTS=true node --import tsx --test **/*.test.ts",
--- a/packages/dependency-graph-integrator/package.json
+++ b/packages/dependency-graph-integrator/package.json
@@ -12,6 +12,7 @@
 		"postbuild": "cp package.json dist/package.json",
 		"prestart": "npm run copy-templates",
 		"start": "tsx src/run-locally.ts",
+		"send-test-event": "tsx src/test-code.ts",
 		"pretest": "npm run copy-templates",
 		"test": "node --import tsx --test **/*.test.ts",
 		"test:update": "UPDATE_SNAPSHOTS=true node --import tsx --test **/*.test.ts",
--- a/packages/dependency-graph-integrator/package.json
+++ b/packages/dependency-graph-integrator/package.json
@@ -3,7 +3,8 @@
 	"version": "1.0.0",
 	"type": "module",
 	"devDependencies": {
-		"@types/aws-lambda": "^8.10.162"
+		"@types/aws-lambda": "^8.10.162",
+		"terminal-link": "^5.0.0"
 	},
 	"scripts": {
 		"copy-templates": "cp ../../.github/workflows/template_dep_submission_gradle.yaml src/template_dep_submission_gradle.yaml && cp ../../.github/workflows/template_dep_submission_sbt.yaml src/template_dep_submission_sbt.yaml",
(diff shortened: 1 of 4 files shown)


packages/dependency-graph-integrator/package.json
```diff
@@ -12,6 +12,7 @@
 		"postbuild": "cp package.json dist/package.json",
 		"prestart": "npm run copy-templates",
 		"start": "tsx src/run-locally.ts",
+		"send-test-event": "tsx src/test-code.ts",
```
akash1810: minor: would `invoke-code` be more intention revealing in that it includes the stage and "test" is a little overloaded?

packages/dependency-graph-integrator/src/test-code.ts
```diff
@@ -0,0 +1,38 @@
...
+	const publish = new PublishCommand({
+		Message: JSON.stringify(message),
+		TopicArn: getEnvOrThrow('DEPENDENCY_GRAPH_INPUT_TOPIC_ARN'),
+	});
+
+	await snsClient.send(publish);
+
+	console.log(
+		"Test event sent to CODE dependency-graph-integrator's input topic.",
+	);
```
akash1810, line 39: Worth also [deep-linking](https://github.com/guardian/service-catalogue/blob/d1083788c00c10c8274e2ec9de084f3d6e2b5f6f/packages/common/src/logs.ts#L13-L39) to the the lambda's logs in Central ELK?

packages/dependency-graph-integrator/src/test-code.ts
```diff
@@ -0,0 +1,46 @@
...
+	const logLink = getCentralElkLink({
```
tjsilver, line 40: I tested this locally and when I clicked on the 'Follow link' command in the terminal, only part of the url was captured. It was truncated at the first `'`, just after `(stage`. I think the queryString mapping in getCentralElkLink needs each value to be URI-encoded. It was fine when I highlighted the entire URL though, so maybe not a change for this PR.

tjsilver (approved):
One comment, non-blocking.

packages/dependency-graph-integrator/src/test-code.ts
andrew-nowak: Yes, I noticed that too. I did have a try with `encodeURI` and `encodeURIComponent` in a few places, but I wasn't able to find a version of it which made the terminal recognise the full link. The best I've been able to find is to wrap the url with `"` quote marks, which at least works for the vscode terminal, although others still seem to not recognise the full url (iterm doesn't include the last bracket, ghostty stops at the first `(`). I don't think there's a good and complete fix for kibana URLs, their format just doesn't play nicely with terminal URL detection

akash1810: I think the issue lies in the `console.log` below:

```ts
console.log(`View the logs at "${logLink}"`);
```

It looks like it doesn't play nice with the [string `getCentralElkLink` returns](https://github.com/guardian/service-catalogue/blob/bae24b6ac17edfdb2bd3bcfe16562ee40135302e/packages/common/src/logs.test.ts#L6-L17).

I'd suggest splitting this across two lines:

```ts
console.log('View the logs at:');
console.log(`  ${logLink}`); // Indent added to provide a sense of hierarchy
```

andrew-nowak: Thanks, I've given this a try, but putting the URL on its own line doesn't seem to have changed behaviour at all to me -

image

image

image

Adding the double quotes at least helps with the vscode terminal, even if it doesn't with the others I've tried

image

akash1810: Ah, totally forgot that we use [`terminal-link`](https://www.npmjs.com/package/terminal-link) too! Example:

https://github.com/guardian/service-catalogue/blob/bae24b6ac17edfdb2bd3bcfe16562ee40135302e/packages/cli/src/aws.ts#L330-L338

Hopefully that helps?

andrew-nowak: Yep, much better (I spent some time trying to figure out why the nicer links weren't working in my non-vscode terminals, but it turns out that the devcontainer cli doesn't automatically forward the [env var that terminal-link uses](https://github.com/chalk/supports-hyperlinks/blob/main/index.js#L32-L33) to determine if the terminal knows how to render links, whereas vscode's implementation does.)

PR #2022, opened 2026-07-20
use incremental sync on AWS inspector table

NovemberTang:
## What does this change?

Runs `aws_inspector2_findings` as an incremental sync.

To reduce the duration of syncs, the number of rows we consume, and improve reliability by reducing rate limiting.

The `@` symbol used in the `backend_options` breaks the yaml construction, as it's a reserved character, so we need to do some slightly fancier escaping in order to surround that value in quotes. The snapshot tests make sure that the existing serialisation is conserved, and we add two new tests to make sure the new fields act how they should

## How has it been verified?

It won't be possible to verify if this saves us _time_, because the CODE DB is smaller than the PROD one, and so the bottleneck is DB throughput rather than cloudquery, but we should be able to get an idea of the number of rows we save from one sync to the next

- [x] Run on CODE (confirm new config works)
- [x] Run on CODE again about 24 hours later (observe multiple cursors)
- [x] Try to observe ratio of new vs old rows, and calculate a rough % saving per day

After running the jobs 16 hours apart (the normal gap is 24), I observed a drop in row consumption of about 65%. Which is promising number for one table that's about 15% of our overall CloudQuery usage. We'll have to observe the drop over a few days to get a fuller picture of the row/cost savings by consulting the Cloudquery Paid Usage dashboard

## Anything else

We should be careful about how we roll out incremental sync, as it's available on a table-by-table basis. Only a few AWS/GitHub tables have this option, but they tend to be the bigger ones

### DIFF
--- a/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/service-catalogue.test.ts.snap
@@ -2253,6 +2253,724 @@ spec:
       },
       "Type": "AWS::IAM::Policy",
     },
+    "CloudquerySourceAwsIncrementalInspector2FindingsScheduledEventRuleBBCC5C30": {
+      "Properties": {
+        "ScheduleExpression": "cron(0 21 * * ? *)",
+        "State": "ENABLED",
+        "Tags": [
+          {
+            "Key": "gu:cdk:version",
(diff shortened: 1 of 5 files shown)


andrew-nowak (approved):
all makes sense! lgtm

PR #2023, opened 2026-07-21
extract and test non-side effecting code

NovemberTang:
## What does this change?

Extracts some non-side-effecting code relating to breakglass user reports and unit tests it.

More unit tests are good.

## How has it been verified?

New unit tests pass, ran successfully on DEV


kelvin-chappell (approved):
👍

PR #2024, opened 2026-07-21
only send messages under certain conditions

NovemberTang:
## What does this change?

Only sends anghammarad notifications if both of the following are true

- ENABLE_MESSAGING is on
- It's Thursday

Sometimes we want to be able to control whether or not we send a notification, especially if we need to test an issue that shows up in PROD but not CODE. We do this with the cloudbuster FSBP report, and also should apply this to the breakglass report.

I'm restricting this message to once a week, as every day seems excessive. I've chosen a different day to the FSBP/repocop notifications to spread things out a bit

## How has it been verified?

Tested locally, observed expected behaviour

### DIFF
--- a/packages/cloudbuster/src/breakglass/digests.ts
+++ b/packages/cloudbuster/src/breakglass/digests.ts
@@ -104,6 +104,7 @@ export async function sendAnghammaradNotification(
 	const variableFields: VariableAnghammaradFields[] =
 		groupUsersAndCreateNotifications(report, awsAccounts);
 
+	// Avoid Mondays (Bank Holidays), Fridays (weekends) and Tuesdays (used for other notifications)
 	const isThursday = new Date().getDay() === 4;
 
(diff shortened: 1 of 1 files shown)


packages/cloudbuster/src/breakglass/digests.ts
```diff
@@ -104,13 +104,21 @@ export async function sendAnghammaradNotification(
...
-			const notification: AnghammaradNotification = {
-				...fixedFields,
-				...fields,
-			};
-			await anghammaradClient.notify(notification);
-		}),
-	);
+	const isThursday = new Date().getDay() === 4;
+
+	if (config.enableMessaging && isThursday) {
```
kelvin-chappell, line 110: Maybe add a comment to explain why Thursday.  I imagine future readers thinking this is significant!

kelvin-chappell (approved):
👍

PR #2025, opened 2026-07-21
send messages to real teams

NovemberTang:
## What does this change?

Sends real messages to teams on PROD

## How has it been verified?

The content of the messages has been tested extensively. The routing of the messages is largely up to anghammarad


kelvin-chappell (approved):
👍

PR #2026, opened 2026-07-21
fix: Remove base branch parameter from dependency graph integrator (DGI)

tjsilver:
## What does this change?

- Removes the `base` parameter from `composeCreatePullRequest` function in the DGI.
- Adds a unit test to ensure the plugin correctly defaults to the `default-branch` of the repo.

The base branch was set to `main` which caused the DGI to fail on a repo which does not have `main` as its default branch.

The `octokit-plugin-create-pull-request` plugin [automatically uses a repo's default branch if `base` is not specified](https://github.com/gr2m/octokit-plugin-create-pull-request/blob/main/README.md). 

## How has it been verified?

- [x] Ran unit test for the plugin successfully.
- [x] Merged this PR and ran the DGI against `test-repocop-prs` which now has the default branch `notmain`. Observed the PR creation. PR: https://github.com/guardian/test-repocop-prs/pull/47

### DIFF
--- a/packages/dependency-graph-integrator/src/pull-requests.test.ts
+++ b/packages/dependency-graph-integrator/src/pull-requests.test.ts
@@ -1,6 +1,7 @@
 import assert from 'assert';
 import { describe, it } from 'node:test';
-import { generateBranchName } from './pull-requests.js';
+import type { Octokit } from 'octokit';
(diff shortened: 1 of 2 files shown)


NovemberTang (approved):
LGTM!

PR #2033, opened 2026-07-23
ci: Refactor to use new parallel feature

akash1810:
## What does this change?
In https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/, GitHub announced native support for parallel build steps. This means we can improve on the changes in https://github.com/guardian/service-catalogue/pull/1156 which moved the lint, test and typecheck tasks to individual jobs as, whilst this approach works, it requires git checkout and dependency installation on each job.

Switching to the [native parallel capability](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsparallel) means:
- We install dependencies once, potentially improving overall build time and cache optimisation
- Removal of a custom action, reducing maintenance
- We have a simpler/shorter GitHub workflow

I've also converted `scripts/build.sh` into parallel steps for a further optimisation.

## How has it been verified?
Look at the [build](https://github.com/guardian/service-catalogue/pull/2033/checks)! For a speed comparison, the [latest build on `main`](https://github.com/guardian/service-catalogue/actions/workflows/ci.yml?query=branch%3Amain) took 4m13s. [This branch](https://github.com/guardian/service-catalogue/actions/workflows/ci.yml?query=branch%3Aaa%2Fci-parallel) took 3m56s. Not a massive speed-up TBH, so the primary benefit of this PR is the change in maintenance.

If we compare the files in Riff-Raff for [`main`](https://riffraff.gutools.co.uk/deployment/request/deployFiles?projectName=deploy%3A%3Aservice-catalogue&id=9171) and [this branch](https://riffraff.gutools.co.uk/deployment/request/deployFiles?projectName=deploy%3A%3Aservice-catalogue&id=9183) we can see the size of the artifacts are the same, demonstrating this is a no-op.

For extra credit, I've checked to see what the experience is like when one of the parallel steps fails. I added the following:

```typescript
const unused: string = 1;
```

This fails the lint and typecheck at least and is reported as:

image

We can also see the build was stopped correctly too:

image


scripts/create-prisma-migrations-artifact.sh
```diff
@@ -0,0 +1,37 @@
...
+# This script packages Prisma migrations into a zip file.
+
+set -e
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+ROOT_DIR=$(realpath "${DIR}/..")
+
+echo "Creating zip of Prisma files"
+
+TEMP_DIR=$(mktemp -d)
+
```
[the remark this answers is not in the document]
akash1810: This is OK in the context of CI.

scripts/create-lambda-artifact.sh
```diff
@@ -0,0 +1,67 @@
...
+  # Create a zip file of the dist directory
+  (
+    cd "$ROOT_DIR/packages/$PACKAGE_NAME/dist"
+    zip -qr "$PACKAGE_NAME.zip" .
+    checkLambdaArtifactSize "$PACKAGE_NAME.zip"
+  )
+}
+
+if [ "$INCLUDE_PRISMA" = "--include-prisma" ]; then
+  createLambdaWithPrisma
```
NovemberTang, line 64: nit: I think it's clearer if these both conform to `createLambda*` or `createZip*`

.github/workflows/ci.yml
```diff
@@ -150,8 +114,41 @@ jobs:
...
+
+      - parallel:
+          - name: lint
+            run: npm run lint
+          - name: test
+            run: npm run test
+          - name: typecheck
+            run: npm run typecheck
+          - name: build
+            run: npm run build
```
kelvin-chappell: Should build be in its own separate block?  There's no point in building if tests fail.

kelvin-chappell (commented):
🚀

.github/workflows/ci.yml
NovemberTang: jinx! This might also speed things up if we can run the lint/test/typecheck alongside the db migration

akash1810: From a risk POV, I don't think either way is better than the other. Steps run in parallel all have to complete successfully, so if the build is ok but the tests fail, the entire CI aborts.

Putting build as a discrete step improves readability 👍🏽 .

> run the lint/test/typecheck alongside the db migration

I'd be reluctant to do this as it starts to mix concerns.

NovemberTang: ah, I meant in a separate job, at the same time, not as part of the same container. I agree, bundling them into the same step would not be ideal

akash1810: Addressed in https://github.com/guardian/service-catalogue/pull/2033/commits/99fdafe24fdfc83db8b6b6b69c0b315124987067.

scripts/create-lambda-artifact.sh
akash1810: Addressed in https://github.com/guardian/service-catalogue/pull/2033/commits/1c00ed0c4698e402f407939d535aaf27036f61b2.

kelvin-chappell (approved):
👍

